PR #13, opened 2016-01-24
Require HAWK authorization header to call the signing api endpoint, fixes #4

### LINKED ISSUE #4: Authenticate API calls using Hawk
all inbound calls to the API should have valid Hawk Authorization tokens

jvehent:
This is ready for review & merge. It implements hawk authorization of requests to the signing endpoint. It does not (yet) implement mapping an authorization to a signing key.

@ncloudioj & @mostlygeek : Since you're both familiar with Hawk, could you please take a look at your convenience?


authorize.go
```diff
@@ -6,16 +6,109 @@
...
+func (a *autographer) authorize(r *http.Request, body []byte) (userid string, authorize bool, err error) {
+	var (
+		auth *hawk.Auth
+	)
+	if r.Header.Get("Authorization") == "" {
+		return "", false, fmt.Errorf("missing Authorization header")
+	}
+	auth, err = hawk.ParseRequestHeader(r.Header.Get("Authorization"))
+	if err != nil {
+		return
```
ncloudioj: Let's return the triplet explicitly here.

authorize.go
```diff
@@ -6,16 +6,109 @@
...
+	}
+	if time.Now().Sub(auth.Timestamp) > maxauthage {
```
ncloudioj: Not sure weather we should use UTC time or not here, but the client and the server should be using the same one.

authorize.go
```diff
@@ -6,16 +6,109 @@
...
+	return true
+}
+
+func (a *autographer) removeNonces() {
+	now := time.Now()
+	for i, n := range a.nonces {
+		if now.Sub(n.timestamp) > maxauthage {
+			// the nonce is too old, delete it
+			a.noncesLock.Lock()
+			a.nonces = append(a.nonces[:i], a.nonces[i+1:]...)
```
ncloudioj: Mutating the nonces inside the loop on itself is not safe. The server crashed in the following case,

When using the correct user id with the incorrect key, the server crashed here with backtrace as:
```bash
panic: runtime error: slice bounds out of range

goroutine 20 [running]:
main.(*autographer).removeNonces(0xc82006e780)
        /Users/njiang/Projects/github/autograph/authorize.go:104 +0x3e4
main.main.func1(0xc82006e780)
        /Users/njiang/Projects/github/autograph/main.go:61 +0x21
created by main.main
        /Users/njiang/Projects/github/autograph/main.go:64 +0x595
```

mostlygeek: The lock should really be at the beginning of the func. ie: 

````
func (a *autographer) removeNonces() {
   a.noncesLock.Lock()
   defer a.noncesLock.Unlock()
   // .... 

````

File: handlers.go [hunk 4]
```diff
@@ -11,11 +11,14 @@ import (
...
 	"net/http"
+	"sync"
 )
 
 // A autographer signs input data with a private key
 type autographer struct {
-	signers []Signer
+	signers    []Signer
+	nonces     []nonce
+	noncesLock sync.Mutex
```
mostlygeek: why not: 

````
type autographer struct {
   sync.Mutex
   signers []Signer
}
````

Then you can: 

func (a *autographer) SomeFunc() {
   a.Lock()
   defer a.Unlock()
}

It makes cleaner looking code.

mostlygeek:
It looks pretty good. I didn't run it. 

Other than nit-picking on style it seems pretty straight forward. 

Personally, I like to wrap my http handlers so I don't need so much boilerplate code in my real handlers. [Example from go-syncstorage](https://github.com/mostlygeek/go-syncstorage/blob/master/api/context.go#L110). Since you only have a single handler it doesn't make any difference. If you start adding more functionality it'll save some boilerplate.

authorize.go
jvehent: It's not mandatory because Go knows how to do the timezone translation, but I agree it is cleaner. Hawk specifies the timestamp as `the number of seconds since January 1, 1970 00:00:00 GMT`, and the `hawk-go` library uses Go's `Unix()` time to generate it, so it's always UTC.

authorize.go
jvehent: I can agree with that. Part of me doesn't like relying on the default value for an authentication result.

authorize.go
jvehent: #fail
I don't want to just lock the entire function, because that prevent adding nonces to the array, which blocks incoming signing requests.
I think I'll just record the positions of the nonces that need removal, and do a second pass on the array, under lock, just to remove them.

handlers.go
jvehent: Mostly because I think we'll need more than one lock to do various operations without locking the entire handler. I don't want to run in a situation where we lock the handler regularly for things that don't necessarily impact normal operation.

authorize.go
mostlygeek: Hmm.. if you mutate `a.nonces` in a different goroutine while `remoteNonces()` is running what does the `range` do? AFAIK `range` is not concurrency safe. You need a lock on it (or somewhere) so nothing else can change `a.nonces` while you're working on it.

jvehent: [... offline discussion happened ...]
I'm not sure that this is the right approach at all anymore. The race condition @mostlygeek mentions is one valid concern, but load balancing is another: if we run N autographers, we can reply a given request N times because each instance keeps its own nonces cache. The solution is to add a central cache (memcache or redis), but I don't really want to do that because it causes an extra roundtrip. So maybe not checking nonce, or doing partial checking, is good enough, given that the requests are already replayed for only 60 seconds, and transferred over https anyway.

ncloudioj: I agree with @mostlygeek that the locking scope should be at the function level here. Given that there is a go routine running background to remove the nonces, it needs to acquire the lock before scanning for the expired nonces one by one. Similarly, it's also necessary to acquire the lock at the function level in the `lookupNonce`. Locking only inside the `range` loop will introduce race condition.

jvehent: I agree, but it's going to block processing of new requests, and I don't want that. I'm looking into using this [lru](https://github.com/hashicorp/golang-lru) package @mostlygeek 's mentioned instead. It will be an imperfect nonce checking, because not centralized, but I think that's an acceptable trade-off.

authorize.go
jvehent: Fixed in https://github.com/mozilla-services/autograph/commit/c8669fe6177c00dd3770703c32af6cf48239d215

authorize.go
jvehent: Fixed in https://github.com/mozilla-services/autograph/commit/c8669fe6177c00dd3770703c32af6cf48239d215

authorize.go
jvehent: Fixed in https://github.com/mozilla-services/autograph/commit/7c8c065e5e79a75e09bf8467214d04256fb27f59

ncloudioj:
LGTM, r+ @jvehent 

My failure tests for HAWK get passed now. Thanks for fixing it!

jvehent:
Thank you Gentlemen for the review :+1:

Issue #21, opened 2016-02-16
Add x5u for dev chain of trust, hosted on github


jvehent:

jvehent:
This is blocked until we have more information about the format of signing certificates (EKU, ...). cc @franziskuskiefer and @mgoodwin

franziskuskiefer:
we will be using unique EKUs for each "context" (will be defined shortly). So the signing certificate will be a normal EE certificate without any KU bits and a EKU according to its "context" (things it's allowed to sign). Is there any other information you need despite the EKU values?

jvehent:
That sounds good to me. I'll be waiting for the spec (EKU values) to implement it. Aside from that, maybe an openssl config example, but I can also figure it out myself.

PR #26, opened 2016-02-19
Various fixes following kinto workweek feedback

jvehent:
work in progress, I have one more change to the signature response format to add `x5u` values (#21), but the rest is there and should be good to review.

r? @ncloudioj & @almet whenever you get a chance


README.md
```diff
@@ -116,6 +116,26 @@ The request body is a json array where each entry of the array is an object to s
...
 
+* signature_encoding: by default, signatures returned by autograph use a R||S
+  string format encoded with base64_urlsafe. The R||S format simply concatenates
+  the two integer value that compose an ECDSA signature into one big number
+  (for p384, each value is 48 bytes long, so the total is 96 bytes). This format
+  avoid relying on ASN.1 parser to read the signatures, but can make it difficult
+  to verify signatures without custom code. The base64_urlsafe encoding format
+  strips base64 padding and replaces characters  `+` and `/` with `-` and `_`
+  respectively.
+  The R||S base64_urlsafe format complies with the Content Signature protocol,
```
almet: I don't know what the "Content Signature protocol" is. Seems like a good place to add a link :)

encoders.go
```diff
@@ -0,0 +1,97 @@
...
+		rs := make([]byte, len(sig.R.Bytes())+len(sig.S.Bytes()))
+		copy(rs[:len(sig.R.Bytes())], sig.R.Bytes())
+		copy(rs[len(sig.S.Bytes()):], sig.S.Bytes())
+		str = base64.StdEncoding.EncodeToString(rs)
+		if format == "rs_base64url" {
+			str = b64Tob64url(str)
+		}
+		return str, nil
+	}
+	if format[0:10] == "der_base64" {
```
almet: What about using [`strings.hasPrefix`](https://golang.org/pkg/strings/#HasPrefix)?

almet:
Great! Looks good to me with some small nitpicks. However, I haven't really checked the go changes, just basically went over the documentation and python changes.

encoders.go
jvehent: Good point. I never think about using, because doing it by hand it so straightforward. Fwiw, the `HasPrefix` implementation does exactly the same thing:
```go
   373	func HasPrefix(s, prefix string) bool {
   374		return len(s) >= len(prefix) && s[0:len(prefix)] == prefix
   375	}
```

encoders.go
```diff
@@ -0,0 +1,97 @@
...
+import (
+	"encoding/asn1"
+	"encoding/base64"
+	"fmt"
+	"math/big"
+	"strings"
+)
+
+func encode(sig *ecdsaSignature, format string) (str string, err error) {
+	if len(format) < 4 {
```
ncloudioj: Perhaps you want to use "rs_base64url" by default? The magic number 4 is a bit confusing here.

ncloudioj:
@jvehent LGTM, except recommending using enum for the signature format.

Looks like the coveralls is not happy, probably there is some new code not being covered by tests.

encoders.go
jvehent: Agreed. I blame jetlag :p I'm replacing this with a simple `if format == "" { }`

jvehent:
Thanks for the review :) Tests are passing, I'm merging this and will tag a branch for @micheletto to deploy.

PR #28, opened 2016-03-03
Fix the base64url padding issue in the python client

ncloudioj:


ncloudioj:
@jvehent r?

jvehent:
That looks good, and verified (manually) that it works as expected. r+

PR #29, opened 2016-03-04
Fix signature padding when R or S is smaller than 48 bytes

jvehent:
@franziskus, would be so kind to give this one a quick review?


jvehent:
The two methods are virtually identical; working on the final signature size seems more natural to me, but that could just be a side effect of how our brains are wired (final results vs inner details).

jvehent:
Rebased and tests pass, merging it. Thank you for the review!

PR #30, opened 2016-03-05
Input rework: split /signature into /sign/data and /sign/hash

jvehent:
This PR is based on #29 and the commits will change once it is merged.

It implements two new endpoints to simplify integration. The description of each endpoint is included in the README. In a nutshell:

* `/sign/data` is meant to receive input of raw data, template it and hash it. It takes a number of parameters in the signature requests to perform those tasks

* `/sign/hash` is a simplified version of the above that assumes the input data is already templated and hashed, and bypasses those steps entirely.

The response format does not change at all, and I kept the former `/signature` endpoint to ease the migration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,8 +187,10 @@ Each signature response contains the following fields:
   be trusted by applications through other means (like a local truststore).
 
 * `content-signature` is the raw HTTP header of the Content-Signature protocol.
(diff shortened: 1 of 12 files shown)


jvehent:
Ha, looks like I'm `0.198%` short on tests :)

PR #31, opened 2016-03-11
Remove /signature endpoint

jvehent:
Autograph clients should either call `/sign/data` or `/sign/hash`.

r? @almet & @ncloudioj : let me know when your clients are migrated to the new endpoints and we can merge this. No rush, it can wait a couple weeks if needed.

### DIFF
--- a/authorize_test.go
+++ b/authorize_test.go
@@ -19,7 +19,7 @@ import (
 func TestMissingAuthorization(t *testing.T) {
 	body := []byte("aaaaaaaaaaaaaaaaaaaaaaaaaaaa")
 	bodyrdr := bytes.NewReader(body)
-	req, err := http.NewRequest("POST", "http://foo.bar/signature", bodyrdr)
+	req, err := http.NewRequest("POST", "http://foo.bar/sign/data", bodyrdr)
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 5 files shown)


ncloudioj:
As for the `sign/hash` endpoint, is it required to template the input before hash it? e.g. add a "Content signature" prefix.

I couldn't find the answer in the document, could you confirm it please? @jvehent

jvehent:
Yep, if you call `/sign/hash`, the `input` need to be templated and hashed already. I'll clarify in the readme.

ncloudioj:
Thanks for confirming it. @jvehent 

Splice currently only hashes the raw input without adding any template. I assume "Content-Signature:\x00" is standard header for now, correct? Will make it configurable for the future updates.

jvehent:
You can use `Content-Signature:\x00`, it should be standard and not change. (cc @franziskuskiefer)

I'd still recommend that you call the `/sign/data` endpoint with just the raw base64 input, and let autograph handle the templating and hashing. I'm not concerned about request size, since you're only signing a couple times a day max.

ncloudioj:
I totally get your point. It's also easier for us to do so.

Our concern is that even if there are only several signs per day, sending the raw input to the signing server will significantly increase the processing time for the large size content, and it might end up hitting HTTP timeout as Splice currently doesn't offload the actual signing to a separate worker queue.

Also, we will be following closely with the specification of Autograph to avoid any incompatible issues.

almet:
That works for me. Opened kinto/kinto-signer#47 about it.

ncloudioj:
@jvehent 
The endpoint cutover for Splice completed. Everything looks good here.

jvehent:
Awesome, thank you @ncloudioj .

almet:
I've updated kinto-signer to use the new APIs. See https://github.com/Kinto/kinto-signer/commit/472ff66dc5aa8d9bbffb507a59f5af46fc47de30

jvehent:
Thanks @almet, I'm going to merge this and it will ride with the next release.

almet:
Cool. We have a stack deployed that relies on the currently old behaviour. Be sure to ping us before deploying this new (incompatible) version, otherwise, things will break :)

jvehent:
@almet: I'll sync up with @phrawzty to make sure we don't break staging.

PR #39, opened 2016-05-12
Dockerfile: Use CMD instead of ENTRYPOINT.

mythmon:
This makes the service Just Work when launched via Docker, and still allows for running other commands, such as bash, for debugging purposes.

This helps with [some tooling we have for testing in Normandy](https://github.com/mozilla/normandy-compose).

@jvehent r?


jvehent:
That's fine by me. We don't use the container now, so if this solves something for you, I'm all for it :)

Issue #37, opened 2016-05-04
Make a monitoring client


jvehent:
Monitoring should could test a signature on a validation key to make sure the service is running correctly.

@micheletto: which format and where should the test output go?

Micheletto:
Would it be possible to tie this into the __heartbeat__ endpoint?  
We've been using JSON as the format for most of our tests.

Issue #36, opened 2016-05-03
Log warning when signing certs reach expiration


jvehent:
This warning should trigger 39 days before expiration with a weekly email and 15 days before expiration with a daily page (via pagerduty).

jvehent:
Covered by the monitoring client

Issue #25, opened 2016-02-19
Add additional awesomeness to __heartbeat__ endpoint.


Micheletto:
The present __heartbeat__ endpoint is functional but could use a bit more awesomeness.  Particularly, as more features are implemented, it would be good if they were rolled into the health check here.

Here's a really good example from the Hello project:
https://github.com/mozilla-services/loop-server/blob/a4068bc919ed02ec870caaa49492f7401d55c2be/loop/routes/home.js#L35-L97

jvehent:
Done.

PR #43, opened 2016-08-05
Signers monitoring

jvehent:
@micheletto : What do you think of the proposed logic described in the README, under `/__monitor__` ?

The idea is to verify the status of all signers without granting full signing rights to a monitoring client. The monitor endpoint will only sign a hardcoded string that doesn't mean anything and  will not accept any input, but it will return signatures of that hardcoded string using all the configured signers. This will allow us to check that all signers are operational, and we can check the status of X5U certs by following the links returned with the signatures.

I'm looking for feedback before implementing. This should cover #25, #36 and #37.
TODO:

[X] test coverage > 85%
[X] read lambda configuration from SOPS file
[ ] support invocation of lambda through SQS events


jvehent:
This isn't ready yet, I need to write unit tests.

Micheletto:
> The idea is to verify the status of all signers without granting full signing rights to a monitoring client. The monitor endpoint will only sign a hardcoded string that doesn't mean anything and will not accept any input, but it will return signatures of that hardcoded string using all the configured signers. This will allow us to check that all signers are operational, and we can check the status of X5U certs by following the links returned with the signatures.

This sounds reasonable to me.  However, we should note that we're putting some of the verification back on the monitoring side here.  Typically the /__heartbeat__ endpoints have been completely self contained with regard to status reporting.  I like the idea of having two levels to the monitor:

1. Simple: 200 OK and X5U cert payload exists in some structured format.
2. Fancy: Monitor verifies X5U certs.

Micheletto:
> The idea is to verify the status of all signers without granting full signing rights to a monitoring client. The monitor endpoint will only sign a hardcoded string that doesn't mean anything and will not accept any input, but it will return signatures of that hardcoded string using all the configured signers. This will allow us to check that all signers are operational, and we can check the status of X5U certs by following the links returned with the signatures.

This sounds reasonable to me.  However, we should note that we're putting some of the verification back on the monitoring side here.  Typically the /__heartbeat__ endpoints have been completely self contained with regard to status reporting.  I like the idea of having two levels to the monitor:

1. Simple: 200 OK and X5U cert payload exists in some structured format.
2. Fancy: Monitor verifies X5U certs.

Micheletto:
> The idea is to verify the status of all signers without granting full signing rights to a monitoring client. The monitor endpoint will only sign a hardcoded string that doesn't mean anything and will not accept any input, but it will return signatures of that hardcoded string using all the configured signers. This will allow us to check that all signers are operational, and we can check the status of X5U certs by following the links returned with the signatures.
> 
> This sounds reasonable to me. However, we should note that we're putting some of the verification back on the monitoring side here. Typically the /heartbeat endpoints have been completely self contained with regard to status reporting. I like the idea of having two levels to the monitor:
> 
>     Simple: 200 OK and X5U cert payload exists in some structured format.
>     Fancy: Monitor verifies X5U certs.

We discussed this yesterday.  I just logged it here because I thought I'd commented on this before, but I had neglected to hit the comment button.  We came to the conclusion that this was more of a synthetic transaction, is authenticated, and therefore merits a different endpoint.

Micheletto:
> We discussed this yesterday. I just logged it here because I thought I'd commented on this before, but I had neglected to hit the comment button. We came to the conclusion that this was more of a synthetic transaction, is authenticated, and therefore merits a different endpoint.

Sometimes weird things happen.  Looks like I really did hit comment. D'oh!

jvehent:
The python approach in Lambda fails because of dependencies to OpenSSL via libffi that can't be shipped with the lambda function (not without large amount of black magic I'm not willing to perform).

Fallback plan: use a python wrapper that invokes a Go program to perform the verification. Work in progress.

jvehent:
I rewrote the lambda job entirely in Go, and the attached Python script is only used as an execution wrapper.

This is functional. Tested in dev. The configuration file can be `sops` encrypted or cleartext. If sops encrypted, the lambda function must have the right role and permissions to access kms.

jvehent:
Sample run on localhost:
```
$ ./monitor.py 
Calling monitoring client using: PATH="/home/ulfr/opt:/usr/local/go/bin:/home/ulfr/bin:/home/ulfr/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/home/ulfr/.rvm/bin:/home/ulfr/.fzf/bin" LAMBDA_TASK_ROOT="." ./autograph-monitor
2016/09/28 11:05:21 Accessing configuration from ./monitor.autograph.yaml
2016/09/28 11:05:21 Retrieving monitoring data from http://localhost:8000/
2016/09/28 11:05:21 Response 0 passes verification
2016/09/28 11:05:21 Response 1 passes verification
2016/09/28 11:05:21 Response 2 passes verification
2016/09/28 11:05:21 Response 3 passes verification
2016/09/28 11:05:21 Response 4 passes verification
2016/09/28 11:05:21 Response 5 passes verification
2016/09/28 11:05:21 Certificate 0 has a valid signature from parent certificate 1
2016/09/28 11:05:21 Certificate 0 is valid from 2016-07-06 21:57:15 +0000 UTC to 2021-07-05 21:57:15 +0000 UTC
2016/09/28 11:05:21 Certificate 1 has a valid signature from parent certificate 2
2016/09/28 11:05:21 Certificate 1 is valid from 2016-07-06 21:49:26 +0000 UTC to 2021-07-05 21:49:26 +0000 UTC
2016/09/28 11:05:21 Certificate 2 is a valid root
2016/09/28 11:05:21 Certificate 2 is valid from 2016-07-06 18:15:22 +0000 UTC to 2026-07-04 18:15:22 +0000 UTC
2016/09/28 11:05:21 Response 6 passes verification
2016/09/28 11:05:21 Certificate 0 has a valid signature from parent certificate 1
2016/09/28 11:05:21 Certificate 0 is valid from 2016-07-06 21:57:15 +0000 UTC to 2021-07-05 21:57:15 +0000 UTC
2016/09/28 11:05:21 Certificate 1 has a valid signature from parent certificate 2
2016/09/28 11:05:21 Certificate 1 is valid from 2016-07-06 21:49:26 +0000 UTC to 2021-07-05 21:49:26 +0000 UTC
2016/09/28 11:05:21 Certificate 2 is a valid root
2016/09/28 11:05:21 Certificate 2 is valid from 2016-07-06 18:15:22 +0000 UTC to 2026-07-04 18:15:22 +0000 UTC
2016/09/28 11:05:21 Response 7 passes verification
```

jvehent:
ping @mozmark @franziskuskiefer: It would be great if one of you could review `monitor.go`, since the logic should be close to what firefox does: https://github.com/mozilla-services/autograph/blob/8252f5df3d2b262136e6fd7c82cc6f2aabddc0be/tools/autograph-monitor/monitor.go

tools/autograph-monitor/monitor.go
```diff
@@ -0,0 +1,361 @@
...
+		return pubkey, fmt.Errorf("Failed to parse public key DER: %v", err)
+	}
+	pubkey = keyInterface.(*ecdsa.PublicKey)
+	return pubkey, nil
+}
+
+func verifyCertChain(certs []*x509.Certificate) error {
+	for i, cert := range certs {
+		if (i + 1) == len(certs) {
+			// this is the last cert, it should be self signed
```
franziskuskiefer: You can't rely on the order of certs in the chain (well you can because you know what you get, but in general you can't).

tools/autograph-monitor/monitor.go
```diff
@@ -0,0 +1,361 @@
...
+
+	if sr.ContentSignature != "" {
+		data = make([]byte, len("Content-Signature:\x00")+len(inputdata))
+		copy(data[:len("Content-Signature:\x00")], []byte("Content-Signature:\x00"))
+		copy(data[len("Content-Signature:\x00"):], inputdata)
+	} else {
+		data = make([]byte, len(inputdata))
+		copy(data, inputdata)
+	}
+	datahash, err := digest(data, sr.Hash)
```
franziskuskiefer: This looks pretty complicated but I guess that's the way you have to do it in go.

PR #44, opened 2016-09-12
Cleanup python client

mythmon:
When the verifier I added to Normandy went through code review in mozilla/normandy#259, several improvements were pointed out. This "backports" those fixes to the autograph client. The main change here is removing the `un_urlsafe` function, since the python base64 module has one built in, and ecdsa is smart enough to detect urlsafe base64.


mythmon:
I'm pretty sure the problem with CI is that I'm not a collaborator on this repo, so secrets were not included in the build, and the tests are not resilient to those secrets being missing.

jvehent:
Thanks! The patch looks good, so merging it. 👍

PR #56, opened 2017-10-04
2.0.5

jvehent:
Fixing a few issues @g-k found in his review, plus healthy dose of vendoring to really feel like we're duplicate half of github in each go project.


handlers.go
```diff
@@ -49,11 +49,17 @@ func (a *autographer) handleSignature(w http.ResponseWriter, r *http.Request) {
...
 	}
+
 	if len(body) < 10 {
 		// it's impossible to have a valid request body smaller than 10 bytes
 		httpError(w, r, http.StatusBadRequest, "empty or invalid request request body")
 		return
 	}
+	if len(body) > 104857600 {
+		// the max body size is hardcoded to 100MB. Seriously, what are you trying to sign?
+		httpError(w, r, http.StatusBadRequest, "request exceeds max size of 100MB")
```
g-k: 👍 This seems generous, but it's good to set a limit.

jvehent: We might have to sign larger blobs of data in the future so I'm being extra cautious.

PR #61, opened 2017-12-29
APK Signer

jvehent:
WORK IN PROGRESS

- [x] Create an APK signer that issues PKCS7 detached signatures
- [x] Support SHA256 PKCS7 signatures in PKCS7 packages (without breaking XPIs)
- [x] Support ZIP Alignment of APK archives
- [x] Unit tests


tools/autograph-client/client.go
```diff
@@ -201,3 +232,7 @@ func verifyXPI(input []byte, resp signatureresponse) bool {
 	}
 	return true
 }
+
+func verifyAPK(input []byte, resp signatureresponse) bool {
+	return true
```
g-k: is this still WIP?

jvehent: errr nope I should fix this

PR #71, opened 2018-03-29
monitor: become a go lambda

jvehent:
currently a python lambda, but not for long!


tools/autograph-monitor/monitor.go
```diff
@@ -46,6 +47,14 @@ var conf configuration
...
 
 func main() {
+	if os.Getenv("LAMBDA_TASK_ROOT") != "" {
+		// we are inside a lambda environment so run as lambda
+		lambda.Start(Handler)
+	}
+	Handler()
+}
+
+func Handler() {
```
ameihm0912: Should we define this function to return `error` instead of no return value?

jvehent: I wonder... the function will `log.Fatal` if errors happen, which is the behavior I want because it'll get captured by lambda as a failed run. Do you think there's a better way to achieve this?

ameihm0912: Maybe that's OK too; I'm not sure if log.Fatal immediately exiting would be treated differently by the Lambda handler (e.g., abnormal termination vs. an explicit error response); e.g., would the logs for the Lambda function just display the error message as a standard log followed up with an indication it exited abnormally instead of displaying an error message?

jvehent: I certainly displays the logs (this is how it works right now) but I'm not sure about the exit status because this function is currently running under a python wrapper. I'll change the error handling to be more idiomatic.

PR #72, opened 2018-03-30
autograph-edge: initial version

jvehent:
WIP
This is a lambda edge function that accepts XPI and APK in authenticated requests (simple token) and forwards them to autograph for signature. It has fine grained permissions, such that a token can only sign a given APK or a given XPI, which makes it easier to distribute to devs for integration in CI, and avoid opening up autograph to the public.


tools/autograph-edge/main.go
```diff
@@ -0,0 +1,144 @@
...
+			fmt.Fprintf(os.Stderr, "error: %v\n", err)
+			os.Exit(1)
+		}
+		log.Printf("%+v", resp)
+	}
+}
+
+func Handler(r events.APIGatewayProxyRequest) (events.APIGatewayProxyResponse, error) {
+	var headers string
+	for k, v := range r.Headers {
```
g-k: Does golang or AWS limit the # and size of request headers? Wondering if there's a possible DOS vector here.

tools/autograph-edge/main.go
```diff
@@ -0,0 +1,144 @@
...
+		headers += k + ": " + v + "; "
+	}
+	log.WithFields(log.Fields{
+		"remoteAddressChain": "[" + r.Headers["X-Forwarded-For"] + "]",
```
g-k: I assume lamdas sit behind an ELB thing that sets X-FF, but it'd be good to confirm this can't be spoofed.

tools/autograph-edge/main.go
jvehent: this was mostly here for testing. I'll remove it for the merge.

tools/autograph-edge/main.go
jvehent: I assume it can be polluted and the elb will add values at the end, so this is similar to what we're doing in other places (don't trust the first value to be the public ip, count from the end).

jvehent:
@g-k @ameihm0912 : this is ready to go, can I get a final review?

tools/autograph-edge/main.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	decodedBody, err := base64.StdEncoding.DecodeString(r.Body)
+	if err != nil {
+		log.WithFields(log.Fields{"rid": r.RequestContext.RequestID}).Error(err)
+		return events.APIGatewayProxyResponse{StatusCode: http.StatusBadRequest, Body: err.Error()}, err
+	}
+
+	log.WithFields(log.Fields{"rid": r.RequestContext.RequestID}).Info("checking authorization token")
+	auth, err := authorize(r.Headers["Authorization"])
```
ameihm0912: Should we verify the authorization before handing the request body off to the decoder vs. after?

jvehent: good catch! fixed in https://github.com/mozilla-services/autograph/pull/72/commits/4fa775a1b769db1ba4bf3936b690f8ed82cf5f42

signer/apk/jar.go
```diff
@@ -22,8 +22,12 @@ func makeJARManifests(input []byte) (manifest, sigfile []byte, err error) {
 	mw := bytes.NewBuffer(manifest)
 
 	for _, f := range r.File {
-		// reserved signature files do not get included in the manifest
 		if isSignatureFile(f.Name) {
+			// reserved signature files do not get included in the manifest
+			continue
+		}
+		if f.FileInfo().IsDir() {
+			// directories do not get included
```
ameihm0912: Are there any other considerations required here for symbolic links or other file types that are not regular files?

jvehent: I don't think so. Java does not honor symlinks in JAR files, so I wouldn't worry about them too much.

PR #73, opened 2018-04-04
xpi: add SignFile interface

jvehent:
This is very similar code than the APK manifest generation, with minor modifications specific to XPIs. I still have to port the unit tests, but it's a good starting point, and it seems to work :crossed_fingers: 

@entequak could you take a look as well? Check that I got all the manifest generation bits right?


signer/xpi/jar.go
```diff
@@ -0,0 +1,173 @@
...
+		h1.Write(data)
+		fmt.Fprintf(mw, "SHA1-Digest: %s\n", base64.StdEncoding.EncodeToString(h1.Sum(nil)))
+		h2 := sha256.New()
+		h2.Write(data)
+		fmt.Fprintf(mw, "SHA256-Digest: %s\n\n", base64.StdEncoding.EncodeToString(h2.Sum(nil)))
+	}
+	manifestBody := mw.Bytes()
+	manifest = []byte(`Manifest-Version: 1.0
+
+`)
```
EnTeQuAk: It isn't necessarily clear if that newline is important or just accidental and should be commented or otherwise added (e.g escaped and via `\n` or so)

jvehent:
@g-k I grabbed an xpi from central for the unit tests. I agree reusing an xpi would be useful but maybe only in CI tests to avoid hotlinking to central in unit tests. Probably not a blocker for this PR though.

PR #74, opened 2018-04-05
xpi: add rsa key cache

jvehent:
Signing XPIs is currently slow due to the RSA key generation required for every signature. This patch adds a cache of 100 private keys that gets populated in the background to accelerate signing requests.


jvehent:
Consuming from a channel is thread safe. For example, those 5 goroutines are consuming from the same channel and the same value never gets read twice. https://play.golang.org/p/jxCioPPvT_r

signer/xpi/x509.go
```diff
@@ -7,12 +7,63 @@ import (
...
+// every minute, add an rsa key to the cache. This will block if
+// the cache channel is already full, which is what we want anyway
+func (s *PKCS7Signer) populateRsa2048Cache() {
+	for {
+		key, err := rsa.GenerateKey(rand.Reader, 2048)
+		if err != nil {
+			log.Fatalf("xpi.populateRsa2048Cache: %v", err)
+		}
+		s.rsa2048Cache <- key
+		time.Sleep(time.Minute)
```
autrilla: The channel will block when writing if it's full, so I don't think this is needed

signer/xpi/x509.go
```diff
@@ -7,12 +7,63 @@ import (
...
+
+// retrieve a key from the cache or generate one if it takes too long
+// or if the size is wrong
+func (s *PKCS7Signer) getRsaKey(size int) (*rsa.PrivateKey, error) {
+	switch size {
+	case 2048:
+		select {
+		case key := <-s.rsa2048Cache:
+			return key, nil
+		case <-time.After(100 * time.Millisecond):
```
autrilla: If the sleep when generating keys is removed, I don't think this is needed

signer/xpi/x509.go
```diff
@@ -7,12 +7,63 @@ import (
...
+			// generate a key if none available
+			return rsa.GenerateKey(rand.Reader, size)
+		}
+	case 4096:
```
autrilla: nit: perhaps cleaner to keep a `map[int]chan *rsa.PrivateKey` to look up the caches instead of duplicated code

probably overkill for two key sizes, but maybe more sizes will be added in the future?

signer/xpi/x509.go
jvehent: I we remove this, we'll deplete the entropy when autograph starts, which could slow down other types of signing. This is a way to play nice with others.

autrilla: Oh, nice! makes sense

signer/xpi/x509.go
jvehent: agreed, but indeed something I'd do in a refactoring if we end up needed more key sizes. I think we'll move to ecdsa before that though.

jvehent:
I think we can trust clients because we issue hawk tokens to them. The
service is also internal only, and not opened to the internet.

jvehent:
Alright, thanks everyone for the reviews! I refactored it one last time because I realized the cache knows the key size at init, since it matches the issuer private key. I also added unit test, and discovered `t.Parallel()`, which is pretty awesome.

@g-k: could I get one last pass when you get a chance please?

PR #76, opened 2018-04-10
XPI cose signing

g-k:
WIP branch to add COSE signatures to addons

~~Assume one signer per addon (i.e. we don't need multiple signatures on the COSE message)?~~

edit: @jvehent I'm thinking we change the signer config as follows:

* add a `cose_signers` key 
* its value is an array with required keys algorithm, certificate, and privatekey with values:
  * algorithm: from the "Name" column of https://www.iana.org/assignments/cose/cose.xhtml#header-algorithm-parameters Might need quotes.
  * certificate and privatekey: same PEM blocks as top level signer values

For example:

```yaml
signers:
- id: extensions-pkcs7-ecdsa
   type: xpi
   certificate: |
        -----BEGIN CERTIFICATE-----
        ... 
   privatekey: | 
        -----BEGIN EC PARAMETERS----- 
        ...
- id: extensions-signer-pkcs7-and-cose
   type: xpi
   certificate: |  # an optional PKCS7 signer
        -----BEGIN CERTIFICATE----- # probably the same as the first cose
        ... 
   privatekey: | 
        -----BEGIN EC PARAMETERS----- 
        ...
   cose_signers: # enables cose signing when present and makes PKCS7 signing optional
     - algorithm: ES256
       certificate: |
        -----BEGIN CERTIFICATE-----
        ... 
       privatekey: |
       -----BEGIN EC PARAMETERS----- 
       ...
     - algorithm: PS256
       certificate: |
        -----BEGIN CERTIFICATE-----
        ... 
       privatekey: |
       -----BEGIN PRIVATE KEY-----
       ...
```
 
This lets us:

* use multiple or different intermediates for COSE signing
* enable and disable PKCS7 signing with COSE signing using a different intermediate if necessary 
* keep existing PKCS7 signers by having a client's authorizations default to the non-COSE signer but allowing clients to select a COSE signer with an explicit signer keyid


franziskuskiefer:
> Assume one signer per addon (i.e. we don't need multiple signatures on the COSE message)?

For now one signature is enough but we might need more in future. Firefox processes all signatures it finds. So for testing it might make sense to add more than one signature per file.

g-k:
OK given that we can use the same intermediate for all signatures.

Our signer chains are a tree and not a forest i.e. we just have chains like Intermediate1 -> EE1, Intermediate1 -> EE2 and not Intermediate1 -> EE1, Intermediate2 -> EE2 and I think the conf can look like:

```yaml
signers:
- id: extensions-pkcs7-ecdsa
   type: xpi
   certificate: |
        -----BEGIN CERTIFICATE-----
        ... 
   privatekey: | 
        -----BEGIN EC PARAMETERS----- 
        ...
- id: extensions-signer-pkcs7-and-cose
   type: xpi
   certificate: |
        -----BEGIN CERTIFICATE----- # probably the same as the first cose
        ... 
   privatekey: | 
        -----BEGIN EC PARAMETERS----- 
        ...
   options:
     cose_algorithms:
        - ES256
        - PS256
```

With behavior as follows:

```
always generate a PCKS7 signature as usual
if cose_algorithms is present
   if any algorithm is invalid (not in the IANA algs)
      return ErrUnrecognizedAlg
   if any algorithm is not supported
      return ErrUnsupportedAlg
   if there is not at least one valid algorithm
      return ErrMissingCOSEAlg

  generate EEs from the provided intermediate and include signatures in the COSE message in the order provided in the config
```

Adding the XPI-specific signer config under a new `signer.Configuration.Options` field with type `interface{}` to:

* not pollute `signer.Configuration` with fields specific to different signer types 
* allow each signer to `json.Unmarshal` options into signer specific options 

signers implementing {Data,File,Hash}Signer that use Options should include them in whatever `GetDefaultOptions` returns

g-k:
per IRC, going to move everything to per request options (this avoids confusion over whether options are from the config vs. request) then we can restrict cose enabled/disabled and allowed algs later in hardcoded global or per signer configs

g-k:
From what I can tell COSE doesn't specify the RSA key bit length for PS256, thinking >= `max(2048, issuerCert bit length)` will be fine.

franziskuskiefer:
> From what I can tell COSE doesn't specify the RSA key bit length for PS256, thinking >= max(2048, issuerCert bit length) will be fine.

According to [RFC 8230](https://tools.ietf.org/html/rfc8230#section-6)
> A key size of 2048 bits or larger MUST be used with these algorithms.

and 
> Implementations SHOULD be able to encrypt and decrypt with modulus between 2048 and 16K bits in   length.

But since we deviate from the COSE RFC by using certs instead of actual `COSE_Key` I think only the `>=2048` matters.

signer/xpi/x509.go
```diff
@@ -104,3 +133,99 @@ func (s *PKCS7Signer) MakeEndEntity(cn string) (eeCert *x509.Certificate, eeKey
 	}
 	return
 }
+
+
+// KeyType is the type to use in keyOptions to tell MakeDEREndEntity
+// which type of crypto.PrivateKey to generate
+type KeyType int
```
jvehent: Does this need to be a public type? seems internal only to me

signer/xpi/x509.go
```diff
@@ -104,3 +133,99 @@ func (s *PKCS7Signer) MakeEndEntity(cn string) (eeCert *x509.Certificate, eeKey
...
+		return &keyOptions{
+			keyType: keyTypeRSA,
+			rsaBits: 2048, // TODO: fixme
+		}
+	default:
+		return nil
+	}
+}
+
+// keyOptions
```
jvehent: :+1:

signer/xpi/x509.go
```diff
@@ -104,3 +133,99 @@ func (s *PKCS7Signer) MakeEndEntity(cn string) (eeCert *x509.Certificate, eeKey
...
+	rsaBits int
+
+	// ecdsaCurve is the elliptic curve to generate an ECDSA key with only applies when keyType is keyTypeECDSA
+	ecdsaCurve elliptic.Curve
+}
+
+// MakeDEREndEntity generates an EE cert and private key like
+// MakeEndEntity but returns the DER encoded certificate bytes instead
+// of a X.509 certificate and takes additional options to generate EEs
+// using different key types or params than the issuer cert
```
jvehent: the only difference between a DER and an X.509 is the call to `eeCert, err = x509.ParseCertificate(derCert)`, so.... dedup?

signer/xpi/xpi.go
```diff
@@ -58,7 +63,7 @@ type PKCS7Signer struct {
 	rsaCache chan *rsa.PrivateKey
 }
 
-// New initializes an XPI signer using a configuration
+// New initializes an new signer using a configuration
```
jvehent: it's still an XPI signer ;)

signer/xpi/xpi.go
```diff
@@ -142,20 +154,145 @@ func (s *PKCS7Signer) Config() signer.Configuration {
...
-func (s *PKCS7Signer) SignFile(input []byte, options interface{}) (signer.SignedFile, error) {
+// isCOSEAlgBad checks whether a COSE algorithm is recognized as a
+// valid IANA algorithm name, and supported and enabled in autograph
+func isCOSEAlgBad(algName string) (alg *cose.Algorithm, err error) {
+	alg, algErr := cose.GetAlgByName(algName)
+	if algErr != nil {
+		err = errors.Wrapf(algErr, "xpi: unrecognized algorithm COSE Signing algorithm")
+		return
+	}
+	if _, exists := supportedCOSEAlgNames[alg.Name]; exists == false {
```
jvehent: totally a nit, but given there are only 4 values we accept, wouldn't an if statement with string equality be more readable?

signer/xpi/xpi.go
```diff
@@ -142,20 +154,145 @@ func (s *PKCS7Signer) Config() signer.Configuration {
...
+
+// SignFile takes an unsigned zipped XPI file and returns a signed XPI file
+func (s *PKCS7Signer) SignFile(input []byte, options interface{}) (signedFile signer.SignedFile, err error) {
 	var (
-		signedFile []byte
+		pkcs7Manifest []byte
+		manifest []byte
+		metas = []Metafile{}
+		opt Options
+		coseSigAlgs []*cose.Algorithm // COSE algorithms to sign with
```
jvehent: nit: gofmt?

signer/xpi/xpi.go
```diff
@@ -142,20 +154,145 @@ func (s *PKCS7Signer) Config() signer.Configuration {
...
+			coseSigners = append(coseSigners, signer)
+
+			// create a COSE Signature holder
+			sig := cose.NewSignature()
+			sig.Headers.Protected["alg"] = alg.Name
+			sig.Headers.Protected["kid"] = [][]byte{eeDERCert[:]}
+			msg.AddSignature(sig)
+		}
+
+		external := []byte("")
```
jvehent: what's `external` and why does it have an empty `[]byte` ?

signer/xpi/xpi.go
```diff
@@ -142,20 +154,145 @@ func (s *PKCS7Signer) Config() signer.Configuration {
...
+
+		err = msg.Sign(randReader, external, cose.SignOpts{
+			GetSigner: func(index int, signature cose.Signature) (cose.Signer, error) {
+				return *coseSigners[index], nil
+			},
+		})
+		if err != nil {
+			return nil, errors.Wrap(err, "xpi: COSE signing failed")
+		}
+		// don't include signature in payload
```
jvehent: is that because we use a detached signature? some more details in comments would be useful

signer/xpi/xpi.go
```diff
@@ -142,20 +154,145 @@ func (s *PKCS7Signer) Config() signer.Configuration {
...
+	}
+	if _, exists := supportedCOSEAlgNames[alg.Name]; exists == false {
+		err = fmt.Errorf("xpi: COSE algorithm %s is not supported", algName)
+		return
+	}
+	return alg, nil
+}
+
+// SignFile takes an unsigned zipped XPI file and returns a signed XPI file
+func (s *PKCS7Signer) SignFile(input []byte, options interface{}) (signedFile signer.SignedFile, err error) {
```
jvehent: I feel like a lot of this code could be refactored into a separate function to be easier to follow. Unsure how, but the COSE part is much longer than the PKCS7 one for some reason, which feels unbalanced. Thoughts?

franziskuskiefer:
Is there no good go cbor library that could be used?
I'll look at the cbor and cose code next week.

g-k:
> Is there no good go cbor library that could be used?

We're using https://github.com/ugorji/go 
Not sure if there's a better option.

I didn't find a COSE library.

signer/xpi/xpi.go
g-k: We aren't using it, but it's `external_aad` from https://tools.ietf.org/html/rfc8152#section-4.3

signer/xpi/xpi.go
g-k: Yeah, I was thinking it'd be easier to add new algorithms.

signer/xpi/xpi.go
g-k: Yeah, it's null per the grand unified signing doc and https://github.com/franziskuskiefer/cose-rust/blob/b53d999253321c0b20dd22434f74047d933fd4c2/src/decoder.rs#L213

signer/xpi/xpi.go
g-k: I'd like to do some of the following to simplify it:

* update the go-cose library (need to do this anyway since it was using math/rand instead of crypto/rand)
  * don't use signer and verifier opts
  * compress headers before signing
  * pass signature params to the message sign function so we don't have to construct and add them
  * I have WIP to have `NewSignMessage` return a pointer, ran into some errors though
* move COSE signing config to signer options
  * this lets us remove the `isCOSEAlgBad` / build `coseSigAlgs` loop
* defined types for a manifest and JAR/XPI that let us add files one by one and not read the ZIP add them in SignFile

There will probably be more inherent complexity in generating multiple EE certs and signatures, but it'd be cleaner to move that logic out of the function.

vendor/go.mozilla.org/cose/core.go
```diff
@@ -0,0 +1,241 @@
...
+
+// I2OSP converts a nonnegative integer to an octet string of a specified length
+// https://tools.ietf.org/html/rfc8017#section-4.1
+//
+// implementation from
+// https://github.com/r2ishiguro/vrf/blob/69d5bfb37b72b7b932ffe34213778bdb319f0438/go/vrf_ed25519/vrf_ed25519.go#L206
+// (Apache License 2.0)
+func I2OSP(b *big.Int, n int) []byte {
+	os := b.Bytes()
+	if n > len(os) {
```
franziskuskiefer: I don't know how worried you are about timing attacks. But this looks like it could be exploitable.

vendor/go.mozilla.org/cose/sign_verify_test.go
```diff
@@ -0,0 +1,212 @@
...
+
+	opts := SignOpts{
+		HashFunc: crypto.SHA256,
+		GetSigner: func(index int, signature Signature) (Signer, error) {
+			return *signer, ErrNoSignerFound
+		},
+	}
+
+	sig := NewSignature()
+	sig.Headers.Protected[algTag] = -41 // RSAES-OAEP w/ SHA-256 from [RFC8230]
```
franziskuskiefer: Maybe I'm missing something but I don't see tests with intermediate certs in the protected body header. Note that the Firefox code expects there always to be a (potentially empty) list of intermediate certs.

vendor/go.mozilla.org/cose/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+Running tests:
+
+```console
+make godep golint  # skip if you already have them
+make install
+go test # note that the rust tests will fail
+```
+
+The [cose-rust](https://github.com/g-k/cose-rust) tests run in CI. To run them locally:
```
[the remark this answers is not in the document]
g-k: I'm using my fork since I have some rather gross changes to rust-cose [here](https://github.com/franziskuskiefer/cose-rust/compare/master...g-k:test-verify-cli?expand=1) and I figured you wouldn't want them upstream.

Some sort of FFI like [this](https://github.com/medimatrix/rust-plus-golang) might be cleaner.

refs: https://github.com/franziskuskiefer/cose-rust/issues/48

vendor/go.mozilla.org/cose/core.go
g-k: I'm not sure if we should be worried.

* we could use the constant time functions in [crypto/subtle](https://golang.org/pkg/crypto/subtle/)

For the other libraries:

* crypto/rsa says "The RSA operations in this package are not implemented using constant-time algorithms."
* crypto/elliptic for the curves say "The cryptographic operations are implemented using constant-time algorithms." but crypto/ecdsa doesn't mention anything

vendor/go.mozilla.org/cose/README.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+## Development
+
+Running tests:
+
+```console
+make godep golint  # skip if you already have them
+make install
+go test # note that the rust tests will fail
```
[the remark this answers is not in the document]
g-k: The go tests shell out to cose-rust.

vendor/go.mozilla.org/cose/core.go
franziskuskiefer: I'd always be worried about timing attacks if there is the possibility of an attacker providing data and/or listening in on operations. This might not be the case here though. I don't know how the infrastructure that runs autograph looks like.

vendor/go.mozilla.org/cose/README.md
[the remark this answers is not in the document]
franziskuskiefer: Yeah I never got around to actually fix that interop issue. Maybe it's worth doing now.

vendor/go.mozilla.org/cose/core.go
g-k: Yeah, it'd be worth doing for other users of the library too.

:ulfr or :bobm will know for sure, but my understanding is outside requests go through at least another layer first `attacker -> load balancer -> AMO backend  -> autograph` or `autograph edge -> autograph`. There might be a load balancer in front of autograph (server) too. AMO devs or ops might be able to hit autograph directly too.

Also, since this is running on AWS someone might be able to provision an instance on the same hypervisor to minimize network jitter and make a timing attack easier.

vendor/go.mozilla.org/cose/core.go
```diff
@@ -0,0 +1,241 @@
...
+		}
+		return sig, nil
+	case *ecdsa.PrivateKey:
+		// https://tools.ietf.org/html/rfc8152#section-8.1
+		r, s, err := ecdsa.Sign(rand, key, digest)
+		if err != nil {
+			return nil, fmt.Errorf("ecdsa.Sign error %s", err)
+		}
+
+		// TODO: assert r and s are the same length will be
```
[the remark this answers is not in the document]
g-k: Added a check on bit lengths (haven't updated the version here yet) and the bit lengths vary by up to six bits. I looked at [the source](https://golang.org/src/crypto/ecdsa/ecdsa.go?s=4660:4743#L144) briefly, but it wasn't clear to me why that was the case.

From an example test run:

```
Bit lengths of integers r and s (256 and 256) do not match the key length 254
Bit lengths of integers r and s (256 and 256) do not match the key length 254
test_cose_sign_verify_two_signatures_tampered_payload
Bit lengths of integers r and s (255 and 256) do not match the key length 254
test_cose_sign_verify_two_signatures
Bit lengths of integers r and s (256 and 256) do not match the key length 254
test_cose_sign_verify_rsa_tampered_signature  
test_cose_sign_verify_rsa_modified_payload                                                                                                 test_cose_sign_verify_rsa                                                                                                                  test_cose_sign_verify_tampered_signature                                                                                   
Bit lengths of integers r and s (255 and 256) do not match the key length 254
test_cose_sign_verify_modified_payload               
Bit lengths of integers r and s (256 and 256) do not match the key length 254
test_cose_sign_verify_P256
Bit lengths of integers r and s (256 and 256) do not match the key length 254
test_cose_sign_verify_P256_no_other_certs
Bit lengths of integers r and s (256 and 255) do not match the key length 254
test_cose_sign_verify_P384_no_other_certs
Bit lengths of integers r and s (384 and 384) do not match the key length 378
test_cose_sign_verify_P521_no_other_certs
Bit lengths of integers r and s (520 and 521) do not match the key length 521
```

vendor/go.mozilla.org/cose/sign_verify_test.go
g-k: Yeah I still need to write Fx/XPI specific tests and checks.

vendor/go.mozilla.org/cose/core.go
jvehent: Autograph never accept requests from untrusted clients, there is always a layer of intermediary between an untrusted client and autograph (AMO, autograph-edge, admin panels, etc.).

jvehent:
nit from golint: `src/go.mozilla.org/autograph/signer/xpi/jar.go:196:14: should replace errors.New(fmt.Sprintf(...)) with fmt.Errorf(...)` (from my code but if you wouldn't mind fixing)

jvehent:
two golint nits:
* autograph/signer/xpi/xpi.go:270:9: if block ends with a return statement, so drop this else and outdent its block
* autograph/signer/xpi/xpi.go:377:10: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary)
go install go.mozilla.org/autograph

signer/xpi/cose.go
```diff
@@ -0,0 +1,259 @@
...
+	if !ok {
+		resultErr = fmt.Errorf("Expected SignMessage must have kid in Protected Headers")
+		return
+	}
+	// check that all kid values are bytes and decode into certs
+	kidArray, ok := kidValue.([]interface{})
+	if !ok {
+		resultErr = fmt.Errorf("Expected SignMessage Protected Headers kid value to be an array got %+v with type %T", kidValue, kidValue)
+		return
+	}
```
jvehent: might it be worth extracted the 19 lines above into a `GetKid()` function since it is repeated here from isSupportedCOSEAlgValue ?

signer/xpi/cose.go
```diff
@@ -0,0 +1,259 @@
...
+	for i, cert := range kidArray {
+		certBytes, ok := cert.([]byte)
+		if !ok {
+			resultErr = fmt.Errorf("Expected SignMessage Protected Headers kid value %d to be a byte slice got %+v with type %T", i, cert, cert)
```
jvehent: do we want to return `%+v` in standard errors? it feel more like debugging details to me, and would probably make logs massive and hard to parse in production.

signer/xpi/cose.go
```diff
@@ -0,0 +1,259 @@
...
+		return fmt.Errorf("failed to read META-INF/cose.sig from signed zip: %v", err)
+	}
+	pkcs7ManifestBytes, err := readFileFromZIP(signedFile, "META-INF/manifest.mf")
+	if err != nil {
+		return fmt.Errorf("failed to read META-INF/manifest.mf from signed zip: %v", err)
+	}
+	coseManifest := string(coseManifestBytes)
+	pkcs7Manifest := string(pkcs7ManifestBytes)
+
+	if !strings.Contains(pkcs7Manifest, "cose") {
```
jvehent: couple things:
1. you could avoid switching to a string type by using `bytes.Contains(manifest, "\x63\x6f\x73\x65")`
2. manifests can be large and this is probably going to be slow, not sure if we care
3. this isn't exactly a deterministic match, any word containing `cose` will match, and [there are quite a few](https://www.thefreedictionary.com/words-containing-cose). I would suggest a deterministic match on the path in the manifest instead: `Name: META-INF/cose.sig` or `META-INF/cose.manifest`.

signer/xpi/cose.go
```diff
@@ -0,0 +1,259 @@
...
+	if err != nil {
+		return errors.Wrap(err, "cose.sig is not a valid COSE SignMessage")
+	}
+
+	return nil
+}
+
+// coseSignature returns a CBOR-marshalled COSE SignMessage
+// after generating EE certs and signatures for the COSE algorithms
+func coseSignature(cn string, manifest []byte, algs []*cose.Algorithm, s *PKCS7Signer) (coseSig []byte, err error) {
```
jvehent: nit: maybe rename PKCS7Signer to XPISigner since it does more than PKCS7 now?

signer/xpi/cose.go
```diff
@@ -0,0 +1,259 @@
...
+	var (
+		coseSigners []cose.Signer
+		tmp         = cose.NewSignMessage()
+		msg         = &tmp
```
jvehent: wat?

signer/xpi/cose.go
jvehent: nit: s/coseSignature/issueCoseSignature/

signer/xpi/cose.go
g-k: This is only called from the go client, but it can't hurt to improve it.

g-k:
Think I fixed everything from the second review.

nb: I did change the pkcs7 digest to SHA256 and not sure if we want to make that an option

r? @jvehent

signer/xpi/xpi.go
```diff
@@ -143,60 +148,121 @@ func (s *PKCS7Signer) Config() signer.Configuration {
...
+		}
+	}
+
+	sigfile, err := makeJARSignature(pkcs7Manifest)
+	if err != nil {
+		return nil, errors.Wrap(err, "xpi: cannot make JAR manifest signature from XPI")
+	}
+
+	opt.COSEAlgorithms = []string{}
+	p7sig, err := s.signData(sigfile, opt)
```
jvehent: should rename `s.signData` to `s.signDataWithPKCS7` to be explicit that you're not issuing cose signatures there.

signer/xpi/xpi.go
```diff
@@ -143,60 +148,121 @@ func (s *PKCS7Signer) Config() signer.Configuration {
...
-	if s.EndEntityCN != "" {
-		cn = s.EndEntityCN
+	coseSigAlgs, err := opt.Algorithms()
+	if err != nil {
+		return nil, errors.Wrap(err, "xpi: cannot make JAR manifest from XPI")
 	}
-	if cn == "" {
-		return nil, errors.New("xpi: missing common name")
+	if len(coseSigAlgs) > 0 {
+		return nil, fmt.Errorf("xpi: cannot use /sign/data for COSE signatures. Use /sign/file instead")
```
jvehent: I don't think this check belongs here. I'd suggest moving it to `SignData` directly, so this function doesn't have to know anything about COSE.

signer/xpi/xpi.go
```diff
@@ -143,60 +148,121 @@ func (s *PKCS7Signer) Config() signer.Configuration {
...
 	}
+
 	toBeSigned, err := pkcs7.NewSignedData(sigfile)
 	if err != nil {
 		return nil, errors.Wrap(err, "xpi: cannot initialize signed data")
 	}
-	// XPIs are signed with SHA1
-	toBeSigned.SetDigestAlgorithm(pkcs7.OIDDigestAlgorithmSHA1)
+	// XPIs are signed with SHA256
+	toBeSigned.SetDigestAlgorithm(pkcs7.OIDDigestAlgorithmSHA256)
```
jvehent: This must be reverted before merge.

signer/xpi/jar.go
```diff
@@ -7,20 +7,58 @@ import (
...
+// makeJARManifest calculates a sha1 and sha256 hash for each zip entry and writes them to a manifest file
+func makeJARManifest(input []byte) (manifest []byte, err error) {
 	inputReader := bytes.NewReader(input)
 	r, err := zip.NewReader(inputReader, int64(len(input)))
 	if err != nil {
 		return
 	}
 
-	// first generate the manifest file by calculating a sha1 and sha256 hash for each zip entry
+	// generate the manifest file by calculating a sha1 and sha256 hash for each zip entry
```
jvehent: nit: `isSignatureFile` further down should probably be renamed to `isJARSignatureFile` since it doesn't check for COSE signature files.

signer/xpi/xpi.go
```diff
@@ -248,19 +352,37 @@ func (sig *Signature) Marshal() (string, error) {
...
 
-// Unmarshal takes the base64 representation of a PKCS7 detached signature
-// and the content of the signed data, and returns a PKCS7 struct
+// Unmarshal parses a PKCS7 struct from the base64 representation of a
+// PKCS7 detached and content of the signed data or it parses a COSE
+// Sign Message struct from the base64 representation of a CBOR
+// encoded Sign Message
 func Unmarshal(signature string, content []byte) (sig *Signature, err error) {
+	// signMessagePrefix is the prefix CBOR Encoded COSE SignMessages
+	signMessagePrefix := []byte{'\xd8', '\x62'}
```
jvehent: should this be a constant from the cose package?

signer/xpi/xpi.go
```diff
@@ -248,19 +352,37 @@ func (sig *Signature) Marshal() (string, error) {
...
 	sig = new(Signature)
 	sig.Data, err = base64.StdEncoding.DecodeString(signature)
 	if err != nil {
 		return sig, errors.Wrap(err, "xpi.Unmarshal: failed to decode base64 signature")
 	}
-	sig.p7, err = pkcs7.Parse(sig.Data)
-	if err != nil {
-		return sig, errors.Wrap(err, "xpi.Unmarshal: failed to parse pkcs7 signature")
+
+	if bytes.HasPrefix(sig.Data, signMessagePrefix) {
```
jvehent: how about `cose.IsSignMessage(sig.Data)` ?

signer/xpi/cose.go
```diff
@@ -0,0 +1,339 @@
...
+// 2) the PKCS7 manifest is present
+// 3) the COSE and PKCS7 manifests do not include COSE files
+// 4) we can decode the COSE signature and it has the right format for an XPI
+// 5) the right number of signatures are present and all intermediate and end entity certs parse properly
+// TODO: 6) there is a trusted path from the included COSE EE certs to the signer cert using the provided intermediates
+//
+func verifyCOSESignatures(signedFile signer.SignedFile, truststore *x509.CertPool, signOptions Options) error {
+	coseManifest, err := readFileFromZIP(signedFile, "META-INF/cose.manifest")
+	if err != nil {
+		return fmt.Errorf("xpi: failed to read META-INF/cose.manifest from signed zip: %v", err)
```
jvehent: we  use `errors.Wrap` in other places so maybe we should continue using it here for consistency?

signer/xpi/cose.go
```diff
@@ -0,0 +1,339 @@
...
+	// check that we can verify EE certs with the provided intermediates
+	intermediates := x509.NewCertPool()
+	for _, intermediateCert := range intermediateCerts {
+		intermediates.AddCert(intermediateCert)
+	}
+	cndigest := sha256.Sum256([]byte(signOptions.ID))
+	dnsName := fmt.Sprintf("%x.%x.addons.mozilla.org", cndigest[:16], cndigest[16:])
+
+	for i, eeCert := range eeCerts {
+		opts := x509.VerifyOptions{
```
jvehent: you should also verify that the signOptions.ID matches the subject.CommonName of the EE

signer/xpi/cose.go
```diff
@@ -0,0 +1,339 @@
...
+		Unprotected: map[interface{}]interface{}{},
+		Protected: map[interface{}]interface{}{
+			kidHeaderValue: nil,
+			algHeaderValue: nil,
+		},
+	}
+)
+
+// isValidCOSESignature checks whether a COSE signature is valid for XPIs
+func isValidCOSESignature(sig *cose.Signature) (eeCert *x509.Certificate, err error) {
```
jvehent: I was expecting this function to actually verify a signature, but it looks like it's only checking some parameters and returning the EE cert. I'd suggest renaming it.

signer/xpi/cose.go
```diff
@@ -0,0 +1,339 @@
...
+		err = errors.Wrapf(err, "xpi: failed to parse X509 EE certificate from COSE Signature")
+		return
+	}
+
+	return
+}
+
+// isValidCOSEMessage checks whether a COSE SignMessage is a valid for
+// XPIs and returns parsed intermediate and end entity certs
+func isValidCOSEMessage(msg *cose.SignMessage) (intermediateCerts, eeCerts []*x509.Certificate, err error) {
```
jvehent: same as above, the naming isn't very clear. maybe `extractCertsFromCOSEMessage` ?

signer/xpi/cose.go
```diff
@@ -0,0 +1,339 @@
...
+
+	var (
+		coseSigners []cose.Signer
+		tmp         = cose.NewSignMessage()
+		msg         = &tmp
+	)
+	msg.Payload = manifest
+
+	// Add list of DER encoded intermediate certificates as message key id
+	msg.Headers.Protected["kid"] = [][]byte{s.issuerCert.Raw[:]}
```
jvehent: you already verify that `issuerCert != nil` above, so this is more of a nit, but maybe check that `len(s.issuerCert.Raw) > X` where X is a minimal value for a cert?

signer/xpi/xpi.go
```diff
@@ -213,10 +279,46 @@ func (s *PKCS7Signer) signData(sigfile []byte, options interface{}) ([]byte, err
...
+	cn = o.ID
+	if s.EndEntityCN != "" {
+		cn = s.EndEntityCN
+	}
+
+	if cn == "" {
+		err = errors.New("xpi: missing common name")
+	}
+	return
+}
```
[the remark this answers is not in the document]
g-k: Yeah that code is confusing. I ended up with something slightly different.

signer/xpi/cose.go
```diff
@@ -0,0 +1,339 @@
...
+// generateIssuerEEKeyPair returns a public and private key pair for
+// the provided COSEAlgorithm
+func (s *XPISigner) generateCOSEKeyPair(coseAlg *cose.Algorithm) (eeKey crypto.PrivateKey, eePublicKey crypto.PublicKey, err error) {
+	var signer *cose.Signer
+
+	switch coseAlg {
+	case nil:
+		err = fmt.Errorf("Cannot generate private key for nil cose Algorithm")
+	case cose.PS256:
+		const size = 2048
```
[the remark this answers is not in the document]
g-k: yeah, I think this is a holdover from when we were discussing min key size. But that's enforced in COSE now.

signer/xpi/cose.go
```diff
@@ -0,0 +1,339 @@
...
+		eeKey, err = s.getRsaKey(size)
+		if err != nil {
+			err = errors.Wrapf(err, "failed to generate rsa private key of size %d", size)
+			return
+		}
+		eePublicKey = eeKey.(*rsa.PrivateKey).Public()
+	case cose.ES256:
+		fallthrough
+	case cose.ES384:
+		fallthrough
```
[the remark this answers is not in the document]
g-k: :+1:  I always forget about this syntax.

signer/xpi/xpi.go
g-k: yeah this removes the need to set algs to an empty list before calling it from SignFile too

signer/xpi/xpi.go
g-k: Done. That'll be a separate PR.

signer/xpi/cose.go
g-k: to me validate != verify so these just check structure and not contents, and I need to add COSE verification using go-cose to the parent `verifyCOSESignatures` (since we need the cose manifest bytes and I didn't want to pass them everywhere).

g-k:
TODO:

- [ ] rebase
- [ ] add cose sig/sig byte verification

jvehent:
How religious do you want to be about golint? It currently fails on 
`/home/ulfr/src/go.mozilla.org/autograph/signer/xpi/xpi.go:43:6: type name will be used as xpi.XPISigner by other packages, and that stutters; consider calling this Signer`
I don't have an opinion, it's up to you.

signer/xpi/cose.go
```diff
@@ -51,7 +51,7 @@ func (s *XPISigner) generateCOSEKeyPair(coseAlg *cose.Algorithm) (eeKey crypto.P
 	case nil:
 		err = fmt.Errorf("Cannot generate private key for nil cose Algorithm")
 	case cose.PS256:
-		const size = 2048
+		size := s.issuerKey.(*rsa.PrivateKey).N.BitLen()
```
jvehent: I think you could end up with an ECDSA key in `s.issuerKey` that would make this panic

g-k:
Yeah I saw that. I think it’s OK to make it more distinct from the
autograph.Signer type and easier to grep for. I'd be OK renaming it to
something else too (maybe AddonSigner or ExtensionSigner or something) but
Signer might be confusing.

signer/xpi/cose.go
g-k: :+1: 

https://github.com/mozilla-services/autograph/pull/76/commits/f9dc49baf186474bff9a1b5d0401f1609cbe4a0f

signer/xpi/cose.go
```diff
@@ -49,9 +49,19 @@ func (s *XPISigner) generateCOSEKeyPair(coseAlg *cose.Algorithm) (eeKey crypto.P
...
-		err = fmt.Errorf("Cannot generate private key for nil cose Algorithm")
+		err = errors.New("Cannot generate private key for nil cose Algorithm")
 	case cose.PS256:
-		size := s.issuerKey.(*rsa.PrivateKey).N.BitLen()
+		if s.issuerKey == nil {
+			err = errors.New("Cannot generate COSE key pair with nil issuerKey")
+			return
+		}
+		var size int
+		issuerRSAKey, ok := s.issuerKey.(*rsa.PrivateKey)
```
jvehent: wait, no, that's not what I meant. I meant that s.issuerKey could be an `*ecdsa.PrivateKey`, which would make this whole code block fail

g-k: Wouldn't this fall back to the minKeySize?

I thought this test would cover that https://github.com/mozilla-services/autograph/blob/979c9c790375365563910b99d2d1e56944b389a0/signer/xpi/cose_test.go#L60-L71

g-k:
Updated the COSE verification function to actually verify COSE signature bytes and tapped AJ to look at the go-cose PRs, so this is ready for another review.

g-k:
@franziskuskiefer Julien went on leave on Monday. Do you think you'll have time to review this?

franziskuskiefer:
> Do you think you'll have time to review this?

I'll try to look at it this week.

tools/autograph-client/build_test_xpis.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env sh
```
franziskuskiefer: Did you test if the zips get verified by Firefox again?
For Firefox unit testing it would be good to have (in addition to xpis in here) test cases with intermediate certs and one with RSA only.

signer/xpi/cose.go
g-k: d'oh yeah, fixed to handle ecdsa and other types in https://github.com/mozilla-services/autograph/pull/76/commits/9bc34836847144f80737749520f425f0bc3facb2

signer/xpi/xpi.go
```diff
@@ -143,54 +148,116 @@ func (s *PKCS7Signer) Config() signer.Configuration {
...
+
+	manifest, err = makeJARManifest(input)
+	if err != nil {
+		return nil, errors.Wrap(err, "xpi: cannot make JAR manifest from XPI")
+	}
+
+	// don't issue COSE signatures if we cannot parse valid and
+	// supported COSE Algorithms from signing request options
+	if len(coseSigAlgs) < 1 {
+		pkcs7Manifest = manifest
```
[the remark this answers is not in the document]
g-k: We do error when COSE Algorithm parsing and validation fails here: https://github.com/mozilla-services/autograph/pull/76/files/a9033a46e2e67b2cdc50bfa23a3786454da48f01#diff-0bf2ee5528414c7d2ac0e24b14a748daR170

This is handling the case where no COSE Algorithms are provided and we only issue a PKCS7 signature. Since we don't need to add entries for cose.sig and cose.manifest to the PKCS7 manifest file, we pass through the manifest from the ZIP to avoid recalculating it.

I'll update the comment to make that clearer.

signer/xpi/xpi.go
```diff
@@ -143,54 +148,116 @@ func (s *PKCS7Signer) Config() signer.Configuration {
...
+	cn, err := opt.CN(s)
+	if err != nil {
+		return nil, err
+	}
+	coseSigAlgs, err := opt.Algorithms()
+	if err != nil {
+		return nil, errors.Wrap(err, "xpi: cannot make JAR manifest from XPI")
+	}
+	if len(coseSigAlgs) > 0 {
+		return nil, fmt.Errorf("xpi: cannot use /sign/data for COSE signatures. Use /sign/file instead")
```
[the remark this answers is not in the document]
g-k: The error msg above is wrong. Julien wants to only issue COSE signatures from `/sign/file` to avoid breaking `/sign/data` which is the only endpoint in use. Updated to skip parsing and validating the `cose_algorithms` from the JSON entirely and just check for a non-empty array.

tools/autograph-client/build_test_xpis.sh
g-k: Haven't tested against Fx since https://bugzilla.mozilla.org/show_bug.cgi?id=1466683, which was against an earlier commit on this branch. I'll add the other test cases.

g-k: Actually, I'm not sure what test cases to add. Multiple intermediate certs? An ecdsa intermediate (I don't think we can do this)? For the RSA only one, do you mean just a pk7 signature and COSE PS256?

g-k:
> Do you want me to look at the go-cose as well?

If you have time that'd be great, depending on when you last looked at it the API may not have changed much.

tools/autograph-client/build_test_xpis.sh
franziskuskiefer: We support using intermediate certs or the root directly when verifying. I want to make sure that we test both (if you support signing with the root that is :). It doesn't matter what keys the intermediates use. We're only restricted in what key is used in the COSE signature.

> For the RSA only one, do you mean just a pk7 signature and COSE PS256?

Yep

g-k: OK I added the pk7, PS256 test case here: https://github.com/mozilla-services/autograph/pull/76/commits/238d19f93424c2207d5536fa6b8d1303d320f534#diff-c93506e8ef6730d1fc392b838221a31dR20

We only support one intermediate and don't sign with the root.

So I think we're only missing tests against an ECDSA intermediate with an ECDSA pk7 EE (we decided not to support generating pk7 EEs that don't match the intermediate private key type) and the various COSE options (just pk7, ES256, PS256, ES256 and ES384, ES256 and PS256).

I don't think we have an ECDSA intermediate in the stage or prod configs, so I'll have to find someone with access to the roots to generate those.

vendor/go.mozilla.org/cose/core.go
```diff
@@ -0,0 +1,364 @@
...
+		// https://tools.ietf.org/html/rfc8152#section-8.1
+		r, s, err := ecdsa.Sign(rand, key, digest)
+		if err != nil {
+			return nil, fmt.Errorf("ecdsa.Sign error %s", err)
+		}
+
+		// These integers (r and s) will be the same length as
+		// the length of the key used for the signature
+		// process.
+		if !(s.BitLen() == r.BitLen() && s.BitLen() == key.D.BitLen()) {
```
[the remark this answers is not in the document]
g-k: No and I don't know why. Is there a weaker constraint we could check?

I played with adding a tolerance (like within 8 or 16 bits in length).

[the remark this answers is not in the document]
franziskuskiefer: Well, I guess `Sign` only does the math. I'd recommend to prepend zeroes up to the required bit-length and only check that it's within 1 byte of the length.

franziskuskiefer:
> @franziskuskiefer should we try to debug the PS256 signatures or are you OK landing this and opening an issue to investigate?

This is actually what I'd expect. The Firefox COSE code sees a signature algorithm it doesn't know and thus rejects the entire COSE signature. The rfc doesn't specify how to handle the case where one algorithm isn't known. We went for the safe way. I filed franziskuskiefer/cose-rust/issues/56 to reconsider how we handle this.

franziskuskiefer:
My review might not be enough because I'm not a collaborator?

tools/autograph-client/build_test_xpis.sh
g-k: Testing the roots and multiple intermediates might be implemented as an optional "COSE message key id prefix" param taking multiple args.

Issue #79, opened 2018-04-17
intermittent failures for TestRsaCaching


g-k:
I'm seeing TestRsaCaching fail intermittently in CI and local envs. I've only checked the branch I'm working on.

```
--- FAIL: TestRsaCaching (12.87s)
        xpi_test.go:349: key retrieval from cache took more than 10ms
```

Is anyone else seeing or has seen this test fail? If so, can we switch it to a https://golang.org/pkg/testing/#Benchmark or test with large timeout and benchmark?

g-k:
This is still failing in CI so maybe we should test RSA key gen from a populated cache is faster than key gen without the cache instead of having a hard time limit.

g-k:
This seems to be fixed.

PR #81, opened 2018-04-23
tests: double RSA cache timeout from 10 to 20ms

g-k:
refs: #79 

r? @jvehent


jvehent:
a bold move but I feel adventurous!

PR #102, opened 2018-06-25
xpi: add the add-on ID to the dns names of the EE cert

jvehent:
This solves Go 1.10's `issuer has name constraints but leaf doesn't have a SAN extension`.

The issue is the add-ons signing intermediates of the Firefox PKI are constrained to not be usable for content signature, but those constraints also require the EE to have a SAN, which is not needed by Firefox. We work around this by adding the add-on ID to the SAN of the EE.

Note that this works with both email and UUIDs.


jvehent:
I'll add some tests in the morning

PR #103, opened 2018-06-27
Differentiate between email and uuid in XPI end entity certs

jvehent:
cc @entequak since this broke AMO CI

### DIFF
--- a/signer/xpi/x509.go
+++ b/signer/xpi/x509.go
@@ -5,8 +5,10 @@ import (
 	"crypto/ecdsa"
 	"crypto/rand"
 	"crypto/rsa"
+	"crypto/sha256"
 	"crypto/x509"
 	"crypto/x509/pkix"
+	"fmt"
 	"log"
 	"math/big"
 	"time"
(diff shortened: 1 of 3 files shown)


signer/xpi/x509.go
```diff
@@ -78,6 +73,16 @@ func (s *PKCS7Signer) MakeEndEntity(cn string) (eeCert *x509.Certificate, eeKey
 		NotAfter:           time.Now().Add(8760 * time.Hour), // one year
 		SignatureAlgorithm: s.issuerCert.SignatureAlgorithm,
 	}
+
+	// having a dnsname or an email address field is required to comply with webpki
+	// requirement due to the intermediates having constraints
+	// see https://github.com/golang/go/issues/24151
+	idEmailRegex := regexp.MustCompile("^[a-zA-Z0-9._~-]+@[a-zA-Z0-9._~-]+")
```
g-k: Not really a fan of regexes, maybe https://golang.org/pkg/net/mail/#AddressParser.Parse instead?

jvehent:
After much trial and error, and discovering the golang core team is trying to tighten up certificate parsing in future release, I concluded the sane path forward is to hash the CN (so we don't choke on weird characters) and store that as a valid fqdn in the dnsname extensions of the cert.

This has zero impact on security, since Firefox doesn't care for the SAN extensions, and this cert isn't ever going to be signed by the webpki, and it makes PKIX parsers happy, so I'm happy.

Issue #111, opened 2018-08-10
log hawk TS skews


g-k:
To help debug load test errors for https://bugzilla.mozilla.org/show_bug.cgi?id=1471195 log received vs. local timestamp and delta of (received - local).

PR #113, opened 2018-08-11
log amount of hawk TS skew

g-k:
fixes: #111 

results in log statements like:

{"Timestamp":1533951098316162668,"Time":"2018-08-11T01:31:38Z","Type":"app.log","Logger":"autograph","Hostname":"gguthe-ThinkPad-X1-Carbon-6th","EnvVersion":"2.0","Pid":9024,"Severity":6,"Fields":{"msg":"Hawk skew 22297h14m1.316154302s exceeded MaxTimestampSkew of 10m0s"}}                                                                                                         

r? @milescrabill

PR #116, opened 2018-08-13
statsd metrics

g-k:
fixes https://github.com/mozilla-services/autograph/issues/111

Changes:

* add optional statsd client per discussion in https://github.com/mozilla-services/autograph/pull/113 (implemented as in tigerblood if there's a better or more up to date way to do this let me know) 
* log hawk TS skew as histogram for all authenticated requests for #111
* adds optional statsd logging for go-runtime metrics (CPU, Memory, GC) using https://github.com/bmhatfield/go-runtime-metrics/ (impl also from tigerblood and again interested in newer & better ways to do this)   
* documents *noncecachesize* server conf option
* update vendored deps; add stats relevant packages
* fix the extra override log when starting without a port cli arg 

Notes: 

* I couldn't get the statsd dial checks to fail on start
* @milescrabill might prefer to not have the statsd and runtime configured in the config file and instead via env vars  
* it might be better not to log the TS `abs(skew)` or treat it as a histogram, but we can tweak that
* we don't set statsd tags for anything; and I have no idea how datadog will handle the statsd output (i.e. whether the names and prefixes are correct)
* changes the signature request timing to be relative to request start time (i.e. start of middleware not handler entrance)


authorize.go
```diff
@@ -44,6 +53,13 @@ func (a *autographer) authorize(r *http.Request, body []byte) (userid string, au
 	}
 	hawk.MaxTimestampSkew = a.auths[userid].hawkMaxTimestampSkew
 	err = auth.Valid()
+	if a.stats != nil {
+		skew := abs(auth.ActualTimestamp.Sub(auth.Timestamp))
+		sendSkewErr := a.stats.Histogram("hawk_timestamp_skew_ns", float64(skew.Nanoseconds()), nil, 1.0)
```
g-k: Might want to rename this. I'm not sure what to rename it to.

Also, we might want to track whether it exceeded the skew error or not. As a counter or tag?

g-k:
@milescrabill feedback on the API / do you see yourself tweaking the statsd and runtime configs and prefer a config file or env vars for them?

main.go
```diff
@@ -83,16 +105,61 @@ func parseArgsAndLoadConfig(args []string) (conf configuration, listen string, a
...
+		return
+	}
+
+	s, err := g2s.Dial("udp", conf.Statsd.Addr)
+	if err != nil {
+		log.Fatalf("Unable to connect to Statsd on %s - %s", conf.Statsd.Addr, err)
+	}
+
+	gaugeFunc := func(key string, val uint64) {
+		s.Gauge(1.0, conf.Statsd.Namespace+key, strconv.FormatUint(val, 10))
```
milescrabill: If you set the prefix per above you can get rid of the `conf.Statsd.Namespace` bit

main.go
```diff
@@ -83,16 +105,61 @@ func parseArgsAndLoadConfig(args []string) (conf configuration, listen string, a
...
 
 	return
 }
 
+func startRuntimeCollector(conf configuration) {
+	if conf.Statsd.Addr == "" {
+		return
+	}
+
+	s, err := g2s.Dial("udp", conf.Statsd.Addr)
```
milescrabill: I think using datadog's statsd is probably better. Then you can just write a wrapper that replaces the `g2s` functionality you're using atm.

If you do that, you can share a client too.

g-k:
I'll cherry-pick https://github.com/mozilla-services/autograph/pull/116/commits/3120feb0986ef4d7f1fb0eeeff6de8db8c0c68e4 out since @entequak says we broke AMO tests.

main.go
```diff
@@ -34,7 +35,10 @@ import (
 	"go.mozilla.org/sops"
 	"go.mozilla.org/sops/decrypt"
 
+	"github.com/DataDog/datadog-go/statsd"
```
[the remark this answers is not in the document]
g-k: `g2s` was a workaround for getting the go runtime pkg working. I think I'll switch to go-metrics.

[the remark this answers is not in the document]
g-k: Though go-metrics looks kind of unmaintained since there are open PRs from 2014.

g-k:
Switched the skew histogram to a timing (since it takes a duration and can compute many of the same metrics). Also, lmk if you want to add timing at other points in the app.

g-k:
Thanks @milescrabill !

PR #119, opened 2018-08-14
XPI compat

g-k:
Changes:

* add MD5 hashes to XPI PKCS7 manifest entries (this is for testing backwards compat)
* sort files using the logic from the python signing clients (according to signing clients this lets Fx optimize loading; it also minimizes the change AMO switching to `/sign/file` will incur)
* add some test XPIs from the `signing_clients` lib and make sure they pass

The test XPI (variant A) for backwards specifies, MD5 and SHA2 hashes for manifest file entries, but doesn't say anything about whether including SHA1 is OK. @franziskuskiefer will Fx try to use the strongest hash it can (SHA2 > SHA1 > MD5) for manifest entry validation?

If not, we could make manifest entry hashes a request option.

open questions:

* rename `test/` files to something with fixtures in the name or inline them as a `[]byte` array?

r? @jvehent


g-k:
Ah, CI is still failing.

franziskuskiefer:
>  will Fx try to use the strongest hash it can (SHA2 > SHA1 > MD5) for manifest entry validation?

Nope, but Fx will read the hash that it is configured to use. Meaning that COSE currently always uses SHA256 and PK7 uses whatever is defined in the PK7 signature, which is then only considered valid iff that digest is the same or better than what is required in the config (has to be sha2 if only sha2 allowed, otherwise sha2 or sha1).

g-k:
Closing in favor of the more minimal changes in: https://github.com/mozilla-services/autograph/pull/121

PR #120, opened 2018-08-15
Validate auth header before reading body

### LINKED ISSUE #114: investigate validating HAWK header before reading request body
From working on #111 https://github.com/mozilla-services/autograph/pull/113, we should try reading and validating the parts of the HTTP header before the whole response.

ajvb:
* Split authorize into authorizeHeader and authorizeBody
* Add metrics counter on authorizeHeader failing (there by exiting early)
* Add a check within httpError to read the request body if it has not been read yet. (golang/go#15789)

Fixes #114 

r? @g-k


authorize.go
```diff
@@ -88,9 +91,19 @@ func (a *autographer) authorize(r *http.Request, body []byte) (userid string, au
 		}
 	}
 	if !auth.ValidHash(payloadhash) {
-		return "", false, fmt.Errorf("payload validation failed")
+		return false, fmt.Errorf("payload validation failed")
+	}
+	return true, nil
+}
+
+// authorize combines authorizeHeader and authorizeBody into one function.
```
g-k: :+1: good idea

authorize_test.go
```diff
@@ -58,7 +58,8 @@ func TestBadPayload(t *testing.T) {
 	}
 	authheader := getAuthHeader(req, ag.auths[conf.Authorizations[0].ID].ID, ag.auths[conf.Authorizations[0].ID].Key, sha256.New, id(), "application/json", []byte(`9247oldfjd18weohfa`))
 	req.Header.Set("Authorization", authheader)
-	_, authorize, err := ag.authorize(req, body)
+	auth, _, _, err := ag.authorizeHeader(req)
+	authorize, err := ag.authorizeBody(auth, req, body)
```
g-k: Could these reuse the `authorize` fn?

authorize_test.go
```diff
@@ -159,10 +160,10 @@ func TestDefaultSignerNotFound(t *testing.T) {
 // Two authorizations sharing the same ID should fail
 func TestAddDuplicateAuthorization(t *testing.T) {
 	var authorizations = []authorization{
-		authorization{
+		{
```
g-k: :+1:

errors.go
```diff
@@ -21,6 +23,10 @@ func httpError(w http.ResponseWriter, r *http.Request, errorCode int, errorMessa
 	}).Errorf(errorMessage, args...)
 	msg := fmt.Sprintf(errorMessage, args...)
 	msg += "\r\nrequest-id: " + rid
+	if r.Body != nil {
```
g-k: nit: might want a comment pointing to the golang issue

handlers.go
```diff
@@ -47,6 +47,17 @@ type signatureresponse struct {
 func (a *autographer) handleSignature(w http.ResponseWriter, r *http.Request) {
 	rid := getRequestID(r)
 	starttime := getRequestStartTime(r)
+	auth, userid, authorized, err := a.authorizeHeader(r)
+	if err != nil || !authorized {
+		if a.stats != nil {
+			sendStatsErr := a.stats.Timing("hawk.authorize_header_failed", time.Since(getRequestStartTime(r)), nil, 1.0)
```
g-k: nit: this can be relative to `starttime` which is already `getRequestStartTime(r)`

errors.go
```diff
@@ -21,6 +23,10 @@ func httpError(w http.ResponseWriter, r *http.Request, errorCode int, errorMessa
...
+		io.Copy(ioutil.Discard, r.Body)
+		r.Body.Close()
```
g-k: also, clever to put this in the error handler. I guess go sync storage does this too, but for some reason I thought we'd end up having to do this all over the place. :1st_place_medal:

ajvb:
@g-k mind re-+1-ing/reviewing this when you have a moment?

authorize.go
```diff
@@ -33,14 +33,11 @@ func abs(d time.Duration) time.Duration {
...
 }
 
-// authorize validates the hawk authorization header on a request
-// and returns the userid and a boolean indicating authorization status
-func (a *autographer) authorize(r *http.Request, body []byte) (userid string, authorize bool, err error) {
-	var (
-		auth *hawk.Auth
-	)
+// authorizeHeader validates the hawk authorization header on a request
+// and returns the hawk auth struct, the userid and a boolean indicating authorization status
```
g-k: nit: docstring is out of date

nit: I forgot to mention this earlier, but since it looks like `authorize` is always false when an error is returned and true when an error isn't returned we can drop the authorize param and just check `error == nil` to get authorize/authorized

authorize.go
```diff
@@ -77,8 +74,14 @@ func (a *autographer) authorize(r *http.Request, body []byte) (userid string, au
...
 	if err != nil {
-		return "", false, err
+		return nil, "", false, err
 	}
+	return auth, userid, true, nil
+}
+
+// authorizeBody validates the body within the request and returns
+// a boolean indiciating whether its valid.
+func (a *autographer) authorizeBody(auth *hawk.Auth, r *http.Request, body []byte) (authorize bool, err error) {
```
g-k: same things here docstring out of date and can combine authorize and err return args.

authorize.go
```diff
@@ -88,9 +91,19 @@ func (a *autographer) authorize(r *http.Request, body []byte) (userid string, au
...
-		return "", false, fmt.Errorf("payload validation failed")
+		return false, fmt.Errorf("payload validation failed")
+	}
+	return true, nil
+}
+
+// authorize combines authorizeHeader and authorizeBody into one function.
+func (a *autographer) authorize(r *http.Request, body []byte) (userid string, authorize bool, err error) {
+	auth, userid, authorize, err := a.authorizeHeader(r)
+	if err != nil || !authorize {
```
g-k: with the above changes can drop the second clause on `authorize` here

errors.go
```diff
@@ -21,6 +23,13 @@ func httpError(w http.ResponseWriter, r *http.Request, errorCode int, errorMessa
 	}).Errorf(errorMessage, args...)
 	msg := fmt.Sprintf(errorMessage, args...)
 	msg += "\r\nrequest-id: " + rid
+	// when nginx is in front of go, nginx requires that the entire
+	// request body is read before writing a response.
+	// https://github.com/golang/go/issues/15789
```
g-k: :+1:

authorize.go
```diff
@@ -33,14 +33,14 @@ func abs(d time.Duration) time.Duration {
...
-// and returns the userid and a boolean indicating authorization status
-func (a *autographer) authorize(r *http.Request, body []byte) (userid string, authorize bool, err error) {
-	var (
-		auth *hawk.Auth
-	)
+// authorizeHeader validates the existence of the Authorization header as a hawk
+// authorization header and makes sure that it is valid. It does not validate the
+// body of the request, that is done within authorizeBody. This function returns
+// the hawk auth struct, the userid, and an error which will indicate whether
+// validation was successful.
```
ajvb: @g-k how does this look?

ajvb:
@g-k I rebased into a single commit as I didn't have gpg commit signing turned on.

authorize.go
g-k: Looks good!

authorize.go
```diff
@@ -77,8 +77,14 @@ func (a *autographer) authorize(r *http.Request, body []byte) (userid string, au
...
 	}
 	if err != nil {
-		return "", false, err
+		return nil, "", err
 	}
+	return auth, userid, nil
+}
+
+// authorizeBody validates the body within the request and returns
+// an error which will be nil if the authorization is successful
```
g-k: :+1:

hwine:
Only thing I can think of off hand is commit signing. ALL the commits on the from branch must be signed, not just the latest. See 
  https://wiki.mozilla.org/GitHub/Repository_Security/Problems_and_Options#Problem:_Needing_to_setup_GPG_for_use_on_GitHub_might_dissuade_contributors

for workarounds

g-k:
yeah aj squashed this down to one signed commit though

g-k:
maybe there's a race in our CI jobs reporting their statuses?

PR #124, opened 2018-08-16
Fix go xpi verification

g-k:
Changes:

* remove vendored autograph
* add roots param tools autograph-client (go) to pass a truststore for verifying signed XPIs

Functional Test:

- [x] sign an XPI with the prod truststore and it verifies instead of errorring out.

r? @ajvb


tools/autograph-client/client.go
```diff
@@ -130,6 +130,7 @@ examples:
 	flag.StringVar(&cn, "cn", "", "when signing XPI, sets the CN to the add-on ID")
 	flag.Var(&algs, "c", "a COSE Signature algorithm to sign an XPI with can be used multiple times")
 	flag.StringVar(&pk7digest, "pk7digest", "sha1", "an optional PK7 digest algorithm to use for XPI file signing. Defaults to 'sha1'")
+	flag.StringVar(&rootPath, "r", "/path/to/root", "Path to a PEM file of root certificates")
```
ajvb: nit: I don't remember if the default value shows up in the help or usage output, but if it does, should it be `/path/to/root.pem`?

g-k: Good call! I think it does, good to clarify that root is file and not a dir in code regardless :+1:

ajvb:
@g-k I think something may be going on with travis-ci or I introduced something weird in my last PR? Just got an email that the build for the `authorize` PR failed (https://travis-ci.org/mozilla-services/autograph/builds/417025854)

ajvb:
And there was a build for the same commit that did pass.

g-k:
@ajvb yeah, I've had to retry them recently.

PR #141, opened 2018-09-17
Update vendor

g-k:
Might be breaking changes in the CBOR encoder/decoder (doesn't panic, returns encode/decode errors instead) for XPI COSE sigs: https://github.com/mozilla-services/autograph/pull/141/files#diff-c119e9b9863ccd8f5153654568a13a0e

However, tests pass.

r? @jvehent


Makefile
```diff
@@ -14,6 +14,8 @@ vendor:
 	#dep ensure -update
 	# https://github.com/ThalesIgnite/crypto11/issues/9
 	git checkout -f 2210ea80470825094edf8235b35f9565c7940555 vendor/github.com/ThalesIgnite/crypto11/
+	rm -rf vendor/go.mozilla.org/autograph/  # don't vendor ourselves
```
jvehent: wtf? how is that even possible?

PR #142, opened 2018-09-19
signer: remove MD5 hashes from XPI manifest entires and sig files

g-k:
Per email thread with dveditz, these only applied to "ancient
pre-legacy add-on signing" and are now bloat.

reverts:

https://github.com/mozilla-services/autograph/pull/121
https://github.com/mozilla-services/autograph/pull/128

r? @jvehent


g-k:
But I learned something along the way

PR #145, opened 2018-09-27
Fix apk sf mf line length

g-k:
refs prior email thread with rbarker

r? @jvehent

### DIFF
--- a/signer/apk/jar.go
+++ b/signer/apk/jar.go
@@ -9,7 +9,56 @@ import (
 	"io"
 	"io/ioutil"
 	"strings"
+	"unicode/utf8"
+	"github.com/pkg/errors"
 )
+
+// consts and vars for formatFilename
+const (
+	maxLineByteLen = 72
+	maxContinuedByteLen = 70 // -1 for leading space and -1 for trailing \n newline
+)
+var maxFirstLineByteLen = maxLineByteLen - (len([]byte("Name: ")) + 1)  // + 1 for a \n newline
+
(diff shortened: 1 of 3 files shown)


signer/apk/jar.go
```diff
@@ -9,8 +9,56 @@ import (
...
+		err = errors.Errorf("apk: invalid UTF8 in filename %s", filename)
+		return
+	}
+
+	const (
+		maxLineByteLen = 72
+		maxContinuedByteLen = 70 // -1 for leading space and -1 for trailing \n newline
+	)
+	var (
+		maxFirstLineByteLen = maxLineByteLen - (len([]byte("Name: ")) + 1)  // + 1 for a \n newline
```
jvehent: nit: recalculating this at every execution seems like a waste

signer/apk/jar.go
```diff
@@ -9,8 +9,56 @@ import (
...
+		fileBytes = len(filename)
```
jvehent: nit: is `fileBytes` the right name for this, or should it be `filenameLen`?

g-k:
I do wonder if this applies to XPIs too or if XPIs aren't that strict about following the JAR format.

jvehent:
This is definitely something we should test. I'm inclined to think Firefox doesn't care, since we have hundreds of thousands of addons and never had a report of breakage, but data is better than intuition.

g-k:
Yeah, I'll ask on #amo and check the python signer source.

PR #148, opened 2018-09-28
APK dont compress media

g-k:
Continuing with the issue from #145.

Update the APK signer to not compress media files and `resources.arsc` to match the behavior devs expect from gradle and Android Studio.

```
// This matches the behavior of Android Studio using Gradle 3.1.4,
// which does not compress a number of media formats and the app
// resource table (at: /resources.arsc).
//
// https://developer.android.com/studio/build/building-cmdline
```

NB: this breaks aligning ZIPs, so I'll try to fix that in this or a later PR before deploying. Failing that, we can tell devs to run zipalign themselves.

r? @jvehent


g-k:
OK took some time to think about this. I think the way to go is:

1. provide a per-APK signing request compression option. Default to the current "compress everything" behavior with the option to not change the compression method (i.e. pass-through per :catlee's recommendation).

2. for the APKs we broke, pin the pass-through option in autograph edge 

3. migrate other APKs from "compress everything" to pass-through or find out why this isn't feasible and coordinate with mobile teams to change the default to pass-through; i.e. autograph just signs it

4. possibly deprecate and remove the "compress everything" option

This also suggests it'd be worth keeping a corpus of test files to verify we don't break things (new, separate issue).

Also, for the currently broken APKs https://github.com/mozilla-services/zipalign outputs correctly aligned ZIPs, which suggests, per :ulfr's hunch, the problem is with trying to write and align in the same loop and splitting this will fix the alignment problem for ZIPs containing uncompressed files.

g-k:
Had a nice chunk of time to look into this on Friday and over the weekend. 

Turns out the go zipalign just compresses everything, which trivially counts as aligned: https://github.com/mozilla-services/zipalign/blob/5bc8740e5ecec8141d09f88591881bcfe8e06011/main.go#L71

And the autograph implementation does the same thing, so we'll need to fix that to have the pass through option output aligned APKs.

In addition to writing and aligning in the same loop as being a possible cause, I think we need to read the go data offset since the C++ impl uses it.  We should also 1) verify that [go's archive/zip File.DataOffset](https://golang.org/pkg/archive/zip/#File.DataOffset) is returning the same offset and 2) make sure the `bias` var is necessary and giving the right results (probably split out a function to compute aligned offsets)

Fortunately, aligning things mod 2 is a common problem:

* this was in the coursera HWSW interface class bit-wise problem set (though I've since forgotten it)
* https://en.wikipedia.org/wiki/Data_structure_alignment#Computing_padding

* screenshot_2018-10-08_04-32-09 from Hacker's Delight 2nd Ed.

Issue #156, opened 2018-10-12
Support APK PCKS7 signature SHA1 digests


g-k:
Some APKs need a SHA1 digest to match their cert, but [we have this hardcoded to SHA256
](https://github.com/mozilla-services/autograph/blob/master/signer/apk/apk.go#L146-L147
).

We can copy the digest option for XPIs to support this (and default to SHA256):

```json
"options": {
                     "pkcs7_digest": "SHA256" // or "SHA1"
                },
```

https://github.com/mozilla-services/autograph/tree/master/signer/xpi#2signature-request

However, differing from the XPI option it won't be required and we'll allow it to be blank (meaning default/SHA256).

This would also require a change in signingscript possibly by adding a new format like `autograph_apk_sha1`. If we need support from autograph edge then we should add a way to pass the option through from there too.

jvehent:
> Some APKs need a SHA1 digest to match their cert

Do they? If android doesn't require the hash algorithm of the pkcs7 sigalg to match the signing cert sigalg, then enforcing SHA256 here is preferable.

g-k:
Yes, per @JohanLorenzo on irc to avoid breaking compat w/ old android versions that don't support SHA2+: https://bugzilla.mozilla.org/show_bug.cgi?id=1332916

g-k:
Fixed in #162.

PR #157, opened 2018-10-12
Add apk digest option

g-k:
fixes: #156

Functional Tests:

- [x] sign apk file with default / SHA256 results in signature with correct digest
- [x] sign apk file with SHA1 results in signature with correct digest

r? @jvehent


jvehent:
I'm an idiot... I didn't see this pull request and ended up reimplementing it in #162

PR #158, opened 2018-10-17
set Read and Write timeouts for http.Server

milescrabill:
These are obviously pretty significant timeouts, I just picked 60 seconds arbitrarily.

This is because of the following doc from the stdlib: "ReadTimeout is the maximum duration for reading the entire request, including the body."

We reverse proxy with nginx in production, so this isn't a huge factor, but if autograph is exposed without a reverse proxy these timeouts will help protect against a nasty set of issues. See this blog post for more info: https://blog.cloudflare.com/exposing-go-on-the-internet/.

### DIFF
--- a/main.go
+++ b/main.go
@@ -165,7 +165,9 @@ func run(conf configuration, listen string, authPrint, debug bool) {
 	router.HandleFunc("/sign/hash", ag.handleSignature).Methods("POST")
 
 	server := &http.Server{
-		Addr: listen,
+		ReadTimeout:  60 * time.Second,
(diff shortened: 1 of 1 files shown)


jvehent:
I don't think you meant to include those two binaries.

g-k:
How about we make the `{Read,Write,Idle}Timeout` timeouts configurable and update the readme / docs?

g-k:
Landed in https://github.com/mozilla-services/autograph/pull/223 but we still need to figure out reasonable settings.

PR #162, opened 2018-10-29
apk: add DSA and ECDSA support

### LINKED ISSUE #155: Support signing APKs with DSA private keys?
Certain older APK keys that will not be named are not RSA nor ECDSA.

To support signing APKs with them we'll probably have to:

* change signer key loading to recognize and https://github.com/mozilla-services/autograph/blob/master/signer/signer.go#L101 to return dsa keys 
* likely change the apk signer
* possibly change [the Mozilla golang pkcs7 lib](https://godoc.org/go.mozilla.org/pkcs7) too

jvehent:
Fixes #155 #156


signer/signer.go
```diff
@@ -98,7 +102,9 @@ func (cfg *Configuration) GetPrivateKey() (crypto.PrivateKey, error) {
 }
 
 // ParsePrivateKey takes a PEM blocks are returns a crypto.PrivateKey
-func ParsePrivateKey(keyPEMBlock []byte) (crypto.PrivateKey, error) {
+// It tries to parse as many known key types as possible before failing and
+// returning all the errors it encountered.
```
g-k: Any reason to do this instead of returning the first error?

tools/autograph-client/client.go
```diff
@@ -142,7 +142,7 @@ examples:
 	flag.StringVar(&cn, "cn", "", "when signing XPI, sets the CN to the add-on ID")
 	flag.IntVar(&sa, "sa", 0, "when signing MAR hashes, sets the Signature Algorithm")
 	flag.Var(&algs, "c", "a COSE Signature algorithm to sign an XPI with can be used multiple times")
-	flag.StringVar(&pk7digest, "pk7digest", "sha1", "an optional PK7 digest algorithm to use for XPI file signing. Defaults to 'sha1'")
+	flag.StringVar(&pk7digest, "pk7digest", "", "an optional PK7 digest algorithm to use for XPI file signing, either 'sha1' (default) or 'sha256'.")
```
g-k: :+1: yes this is better

tools/autograph-client/client.go
```diff
@@ -190,6 +193,12 @@ examples:
 			ZIP: zipMethodOption,
 		}
 	}
+	// if we have a pk7digest but no cn, assume we're signing an APK
+	if pk7digest != "" && cn == "" {
```
g-k: nit: I think it'd be cleaner to make this an `else`

signer/signer.go
jvehent: there's a lot of trial and error happening in that function. We start with the most common type and continue parsing until it succeeded or we give it. In that last case, I think returning all the errors is the best option to help debugging, but the formatting is also really terrible so I'm open to any better idea.

tools/autograph-client/client.go
jvehent: I don't think it works, because if it's not an xpi, then it could be one on of 3 other types.

I think we should rewrite this client entirely, to be honest, and have type detection via filename or explicit flags.

signer/signer.go
g-k: Cool just wondering why.

tools/autograph-client/client.go
g-k: opened: https://github.com/mozilla-services/autograph/issues/163

g-k:
lgtm

tools/autograph-client/client.go
g-k: I think separate flags are better for now (was debating this at the time and thought a consistent CLI interface would be better but setting options shouldn't get this complicated).

Issue #122, opened 2018-08-16
port metainf case insensitivity tests for xpi sign file


g-k:
https://github.com/mozilla/signing-clients/blob/29c0130c0d0765d62acbfb294f8ff6ecd102dc13/signing_clients/tests/__init__.py#L282-L286

g-k:
https://github.com/mozilla-services/autograph/pull/173#issuecomment-464112072

PR #173, opened 2018-11-14
xpi: Add failing test to skip mixed case manifest.mf files

g-k:
fixes: #122 

This should fail ...


jvehent:
is this still a thing we need?

g-k:
Not if we get XPIs from AMO or well-formatted XPIs from non-AMO clients.

PR #176, opened 2018-11-15
signer: remove the sleep populating rsa keys for xpi signing

g-k:
Per IRC discussion about XPI load testing.

I functional tested locally against softhsm2 to make sure it uses the HSM RNG. 

I assume we'll want to decide on whether to deploy to prod based on perf numbers from dev / stage.

We can also tweak:

* the timeout before generating our own rsa keys
* the number of goroutines populating the caching

r? @milescrabill

### DIFF
--- a/signer/signer_test.go
+++ b/signer/signer_test.go
@@ -7,6 +7,7 @@
 package signer
 
 import (
+	"io"
 	"crypto/rsa"
 	"testing"
 )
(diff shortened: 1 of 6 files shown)


signer/signer_test.go
```diff
@@ -203,3 +204,37 @@ func TestNoSuitableKeyFound(t *testing.T) {
 		t.Fatalf("expected to fail with no suitable key found but failed with: %v", err)
 	}
 }
+
+func TestGetRandReaderNoHSM(t *testing.T) {
```
milescrabill: Won't this test pass regardless of which reader type we get back? You can type assert down to the actual type to prevent this.

signer/signer_test.go
```diff
@@ -203,3 +204,37 @@ func TestNoSuitableKeyFound(t *testing.T) {
...
+	if randReader == nil {
+		t.Fatal("expected io.Reader but got nil")
+	}
+	switch randReader.(type) {
+	case io.Reader:
+		break
+	default:
+		t.Fatalf("expected io.Reader but got %T", randReader)
+	}
+	numBytesRead, err := randReader.Read([]byte("000"))
```
milescrabill: What are we testing here? Does this differentiate between the readers because the HSM reader errors on this operation? This [works](https://play.golang.org/p/fCJ9U3EoUwr) for the `crypto/rand` `Reader` case.

g-k:
Cool thanks I'll make those changes. I think we still have the dev cloud hsm somewhere to test against too, so I can try that.

g-k:
Apparently the sleep is to not deplete the entropy pool for other signers https://github.com/mozilla-services/autograph/pull/74#discussion_r179576886 but the timeout waiting for the cache can go away https://github.com/mozilla-services/autograph/pull/74#discussion_r179575126

g-k:
Going to open a new branch.

PR #182, opened 2018-11-21
Add cache metrics

g-k:
As discussed in IRC #autograph.

Added the histogram on key fetching since it sometimes results in key gen too (and should be bimodal if we aren't pulling from cache).

Also, thought it'd be good to tag metrics with signer ID, Type, and Mode so we can filter on those, identify problematic signers, etc. (this PR and to do more generally in the future).

r? @milescrabill

### DIFF
--- a/autograph.yaml
+++ b/autograph.yaml
@@ -90,7 +90,7 @@ signers:
         numgenerators: 2
         generatorsleepduration: 1m
         fetchtimeout: 100ms
-        monitorsamplerate: 1s
+        statssamplerate: 1s
       certificate: |
           -----BEGIN CERTIFICATE-----
           MIIH0zCCBbugAwIBAgIBATANBgkqhkiG9w0BAQsFADCBvDELMAkGA1UEBhMCVVMx
--- a/autograph.yaml
+++ b/autograph.yaml
@@ -90,7 +90,7 @@ signers:
         numgenerators: 2
         generatorsleepduration: 1m
         fetchtimeout: 100ms
-        statssamplerate: 1s
+        statssamplerate: 1m
       certificate: |
           -----BEGIN CERTIFICATE-----
           MIIH0zCCBbugAwIBAgIBATANBgkqhkiG9w0BAQsFADCBvDELMAkGA1UEBhMCVVMx
--- a/autograph.yaml
+++ b/autograph.yaml
@@ -90,6 +90,7 @@ signers:
         numgenerators: 2
         generatorsleepduration: 1m
         fetchtimeout: 100ms
+        monitorsamplerate: 1s
       certificate: |
           -----BEGIN CERTIFICATE-----
           MIIH0zCCBbugAwIBAgIBATANBgkqhkiG9w0BAQsFADCBvDELMAkGA1UEBhMCVVMx
--- a/main_test.go
+++ b/main_test.go
@@ -285,7 +285,12 @@ func TestStartMain(t *testing.T) {
 
 	os.Args = []string{"test-autograph", "-D"}
 	go main()
-	time.Sleep(200 * time.Millisecond)
+	if os.Getenv("CI") == "true" {
+		// sleep longer when running in continuous integration
+		time.Sleep(3 * time.Second)
+	} else {
+		time.Sleep(200 * time.Millisecond)
+	}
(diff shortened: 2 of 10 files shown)


main_test.go
```diff
@@ -285,7 +285,12 @@ func TestStartMain(t *testing.T) {
 
 	os.Args = []string{"test-autograph", "-D"}
 	go main()
-	time.Sleep(200 * time.Millisecond)
+	if os.Getenv("CI") == "true" {
```
jvehent: maybe `!= ""` ?

signer/signer.go
```diff
@@ -42,6 +42,10 @@ type RSACacheConfig struct {
 	// FetchTimeout is how long a consumer waits for the cache
 	// before generating its own key
 	FetchTimeout time.Duration
+
+	// MonitorSampleRate is how frequently the monitor reports the
+	// cache size and capacity
+	MonitorSampleRate time.Duration
```
jvehent: Since this has nothing to do with the `autograph-monitor`, I think we should use a different naming to avoid confusing. How about `StatsSamplingRate` ?

autograph.yaml
```diff
@@ -90,6 +90,7 @@ signers:
         numgenerators: 2
         generatorsleepduration: 1m
         fetchtimeout: 100ms
+        monitorsamplerate: 1s
```
jvehent: isn't 1s a bit aggressive?

signer/xpi/xpi.go
```diff
@@ -146,15 +155,26 @@ func New(conf signer.Configuration) (s *XPISigner, err error) {
 		if issuerPrivateKey.N.BitLen() < 2048 {
 			return nil, errors.Errorf("xpi: issuer RSA key must be at least 2048 bits")
 		}
+		if conf.RSACacheConfig.MonitorSampleRate < 5*time.Millisecond {
```
jvehent: a lower limit of 5ms is very low. I'd think anything lower than 5 seconds is already too frequent.

signer/xpi/xpi.go
```diff
@@ -303,6 +323,42 @@ func (s *XPISigner) signDataWithPKCS7(sigfile []byte, cn string, digest asn1.Obj
...
+	}
+	var (
+		err        error
+		signerTags []string = []string{s.ID, s.Type, s.Mode}
+	)
+	err = s.stats.Histogram(name, float64(value), signerTags, 1)
+	if err != nil {
+		log.Warnf("Error sending histogram %s: %s", name, err)
+	}
+}
```
jvehent: I can think of other uses for these functions outside of the xpi signer. Would it be useful to export them from the `go.mozilla.org/autograph/signer` package?

autograph.yaml
g-k: Yeah that was just to see it was working locally. Bumped to 1 min.

@milescrabill any idea what we'll want to use in stage and prod?

signer/xpi/xpi.go
g-k: Yeah I'd like to add the signer ID, type, and mode tags to all stats we send from signers.

I think we can wrap the datadog statsd client and export from the signer pkg.

I was also have a branch that adds the cache as autograph-shared resource, but we end up specifying key sizes which might be worse.

main_test.go
g-k: Going to leave as is and we can change them both in another PR. Does seem weird to depend on CI setting this specific value.

g-k:
re r? @jvehent

signer/xpi/xpi.go
```diff
@@ -146,15 +154,26 @@ func New(conf signer.Configuration) (s *XPISigner, err error) {
 		if issuerPrivateKey.N.BitLen() < 2048 {
 			return nil, errors.Errorf("xpi: issuer RSA key must be at least 2048 bits")
 		}
+		if conf.RSACacheConfig.StatsSampleRate < 5*time.Second {
+			log.Warnf("xpi: sampling rsa cache as rate of %s (less than 5ms)", conf.RSACacheConfig.StatsSampleRate)
```
jvehent: nit: should be `less than 5s`

g-k: ah good catch pushed a fix

PR #183, opened 2018-11-26
Rsa reuse xpi

jvehent:
WIP

### DIFF
--- a/signer/xpi/x509.go
+++ b/signer/xpi/x509.go
@@ -69,8 +69,6 @@ type rsaKey struct {
 	key   *rsa.PrivateKey
 }
 
-var currentRsaKey rsaKey
-
 // getRsaKey applies some intelligence to key management. It will return the
 // current key if its lifetime or usage count aren't exceeded, or will get a
 // new key from the cache. If the cache is unresponsive, it will generate a new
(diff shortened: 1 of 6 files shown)


signer/xpi/x509.go
```diff
@@ -56,30 +56,45 @@ func (s *XPISigner) monitorRsaCacheSize() {
 	}
 }
 
-// retrieve a key from the cache or generate one if it takes too long
-// or if the size is wrong
+var currentRsaKeyAge time.Time
+var currentRsaKeyUsage int
+var currentRsaKey *rsa.PrivateKey
```
g-k: nit: this can be one var block or maybe we want to throw the age, usage, key, and chan in an RSAKeyCache struct?

signer/xpi/x509.go
```diff
@@ -56,30 +56,45 @@ func (s *XPISigner) monitorRsaCacheSize() {
...
 func (s *XPISigner) getRsaKey(size int) (*rsa.PrivateKey, error) {
 	var (
-		err error
-		key *rsa.PrivateKey
+		err   error
 		start time.Time
 	)
+	if currentRsaKeyAge.Add(s.rsaKeyMaxAge).After(time.Now()) && // if current key hasn't reached max lifetime
+		currentRsaKeyUsage < s.rsaKeyMaxUsage { // if current key hasn't reached max usage
+		currentRsaKeyUsage++
```
g-k: for multiple concurrent requests will this be atomic?

g-k: we can test this with https://golang.org/doc/articles/race_detector.html#Race_on_loop_counter

signer/xpi/x509.go
```diff
@@ -61,33 +62,63 @@ func (s *XPISigner) monitorRsaCacheSize() {
...
+	// only allow one execution of this function at a time
+	currentRsaKey.lock.Lock()
+	defer currentRsaKey.lock.Unlock()
+
+	// see if we can reuse the current key
+	if currentRsaKey.key != nil &&
+		currentRsaKey.age.Add(s.rsaKeyMaxAge).After(time.Now()) && // if current key hasn't reached max lifetime
+		currentRsaKey.usage < s.rsaKeyMaxUsage { // if current key hasn't reached max usage
+		currentRsaKey.usage++
+		return currentRsaKey.key, nil
```
g-k: nit: we should either hoist `start` to above the lock and send to the histogram from this early exit or remove sending it later since it'll be incomplete

signer/xpi/x509.go
```diff
@@ -61,33 +62,63 @@ func (s *XPISigner) monitorRsaCacheSize() {
 	}
 }
 
-// retrieve a key from the cache or generate one if it takes too long
-// or if the size is wrong
+type rsaKey struct {
+	lock  sync.Mutex
+	age   time.Time
```
g-k: nit: I think we should rename this to `creationTime` or `createdAt` since age implies a duration (like the other age vars) to me

signer/xpi/x509.go
```diff
@@ -61,33 +62,63 @@ func (s *XPISigner) monitorRsaCacheSize() {
...
 	)
+
+	// we're messing with pointers and counters shared across goroutines, so
+	// only allow one execution of this function at a time
+	currentRsaKey.lock.Lock()
+	defer currentRsaKey.lock.Unlock()
+
+	// see if we can reuse the current key
+	if currentRsaKey.key != nil &&
+		currentRsaKey.age.Add(s.rsaKeyMaxAge).After(time.Now()) && // if current key hasn't reached max lifetime
```
g-k: won't this always be true?

jvehent: if the key is 20 minutes old and max age is 10 minutes then it will be before the current time by 10 minutes and the test will fail.

jvehent:
Since this is fairly controversial and we didn't need it, closing it.

Issue #115, opened 2018-08-13
switch to go modules


g-k:
~~https://github.com/golang/dep~~ https://github.com/golang/go/wiki/vgo

Benefits:

* don't need to vendor all deps
* can write lockfile of recent

Downsides:

* need to trust the tool
* tool is relatively new
* requires newer versions of git and access to internet in builds
* downloading deps can incur CI failures when they fail and make CI flakey

g-k:
thoughts @ajvb @milescrabill @jvehent ?

ajvb:
I'd switch to vgo as it will become the new standard (https://github.com/golang/go/wiki/vgo)

moz examples:
https://github.com/mozilla-services/autograph-edge/blob/master/Makefile#L12
https://github.com/mozilla-services/cloudtrail-streamer/blob/master/Makefile#L35

ajvb:
In that wiki page, it does state `For any production workloads, use dep, or migrate to it if you have not done so already.` + `The [vgo] proposal has been accepted and will be merged in the Go tree and replace dep at a later date.`

But I've been using vgo on two projects with great success + ease.

g-k:
oh right, that's better!  I want say the one I was thinking of @ajvb. E_TOO_MANY_TOOLS

ajvb:
>need to trust the tool

I trust `vgo`, and it is now under the same development management / code review process as `go` and all standard tooling, but yeah it is very new.

>tool is relatively new

I've seen some bugs appear and be fixed since I have started using it, but so far they have all resulted in the initial build blowing up, so it's very obvious.

>requires newer versions of git and access to internet in builds
>downloading deps can incur CI failures when they fail and make CI flakey

Yeah absolutely. Is there anything in autograph's release/test pipeline that may make this a bigger problem than it would be on another project?

g-k:
> Is there anything in autograph's release/test pipeline that may make this a bigger problem than it would be on another project?

I have run into vgo needing a newer version of git (>= 2.18 or something), but that's it.

ajvb:
>I have run into vgo needing a newer version of git (>= 2.18 or something), but that's it.

Oh yeah, I do believe that is the case.

g-k:
Not sure how far out it is, but maybe waiting for vgo to land / go 1.11 is best.

g-k:
OK let's do this just not right away.

g-k:
fixed in https://github.com/mozilla-services/autograph/pull/330

PR #189, opened 2018-12-04
vgo attempt 2 issue 115

g-k:
Fixes: #115 ~~based off #188~~ 

NB: 

* the crypto11 problem that forced us to pin an older version was fixed in https://github.com/ThalesIgnite/crypto11/pull/10 but we should test for that issue in HSM stage
* ~~The CCI build job is failing with "go: cannot find main module; see 'go help modules'" I believe this is due to not having `go.mod` and `go.sum` published at go.mozilla.org. I think we can land this branch then replay the build and it should pass, otherwise we might need to split out monitor into its own repo.~~ edit: miles pointed out we weren't installing the local checkout

@milescrabill this might require pipeline changes to deploy monitor

r? @jvehent @milescrabill


jvehent:
Do we want to get rid of the vendor directory? Specifically, how much value is there in having local copies of the dependencies? How high is the risk of one of those disappearing from the internet, which would break our builds?

g-k:
> Do we want to get rid of the vendor directory?
> Specifically, how much value is there in having local copies of the dependencies?

The vendor dir:

* gives us diffs of updated deps (frequently too big to view in github though), which is theoretically easier to audit
* makes it easier to install and build the app without changes

vgo / 1.11 modules:

* makes easier update and pin individual packages, which could make dep change review easier (and frees us from things like the makefile hack to checkout the old crypto11 version for a known working commit)

Also, discussed some in https://github.com/mozilla-services/autograph/issues/115 

> How high is the risk of one of those disappearing from the internet, which would break our builds?

I think some downtime is acceptable. Especially if we cache built docker images in the dockerflow pipeline allowing us to redeploy existing tags, but I'd expect the risk of breakage to be the probability of the host for any single package being down at build time. 

There might be a way to have the best of both worlds and keep the module cache in the app dir.

milescrabill:
We can use modules and still vendor our deps if we want, an approach is described here: https://arslan.io/2018/08/26/using-go-modules-with-vendor-support-on-travis-ci/.

Essentially: `GO111MODULE=on`, `go mod vendor`, `go test -mod=vendor`, `go build -mod=vendor`

g-k:
Closing in favor a new branch that I can be more deliberate in and cherry pick out of this branch.

PR #204, opened 2018-12-28
widevine signer

g-k:
# WIP

:ulfr mentioned this is a just a RSA PSS hash signer.

Currently based off the gpg-cli branch.


jvehent:
This looks good. I'm wondering if instead of calling it widevine, we could simply call it "rsapss" and make it a generic signer (since that's effectively what it is).

g-k:
Yeah, I called it widevine since that's what signingscript called it, but RSA-PSS makes sense.

signer/rsapss/rsapss.go
```diff
@@ -0,0 +1,165 @@
...
+	// Type of this signer is "rsapss"
+	Type = "rsapss"
+)
+
+// RSAPSSSigner holds the configuration of the signer
+type RSAPSSSigner struct {
+	signer.Configuration
+
+	// key is the parsed RSA private key to sign hashes
+	key *rsa.PrivateKey
```
jvehent: would you like to make this hsm-proof right away? in which case, you should manipulate a crypto.PrivateKey instead of an *rsa one.

signer/rsapss/rsapss.go
```diff
@@ -0,0 +1,165 @@
...
+	s.PrivateKey = conf.PrivateKey
+	parsedPrivateKey, err := signer.ParsePrivateKey([]byte(conf.PrivateKey))
+	if err != nil {
+		return nil, errors.Wrap(err, "rsapss: failed to parse private key")
+	}
+	rsaKey, ok := parsedPrivateKey.(*rsa.PrivateKey)
+	if !ok {
+		return nil, errors.Errorf("rsapss: parsed private key is not RSA")
+	}
+	s.key = rsaKey
```
jvehent: nit: can use s.key on line 57 directly

signer/rsapss/rsapss.go
```diff
@@ -0,0 +1,165 @@
...
+		ID:         s.ID,
+		Type:       s.Type,
+		PrivateKey: s.PrivateKey,
+		PublicKey:  s.PublicKey,
+	}
+}
+
+// SignData takes data hashes it and returns a signed base64 encoded hash
+func (s *RSAPSSSigner) SignData(data []byte, options interface{}) (signer.Signature, error) {
+	h := sha1.New()
```
jvehent: premature optimization: maybe support other hash types?

tools/autograph-client/client.go
```diff
@@ -146,6 +150,7 @@ examples:
 	flag.IntVar(&maxworkers, "m", 1, "maximum number of parallel workers")
 	flag.StringVar(&cn, "cn", "", "when signing XPI, sets the CN to the add-on ID")
 	flag.IntVar(&sa, "sa", 0, "when signing MAR hashes, sets the Signature Algorithm")
+	flag.StringVar(&rsapssHash, "wa", "base64(sha1(data))", "for RSA-PSS Base64 SHA1 hash to sign using the /sign/hash endpoint")
```
jvehent: why is this a separate flag instead of just using the `-a` parameter?

signer/rsapss/rsapss.go
g-k: I tried that but I get `expected identifier on left side of :=` and switching to `=` gives `undefined: ok`, but we can do that assignment with a `var ok bool` declared beforehand.  ¯\\_(ツ)_/¯

signer/rsapss/rsapss.go
g-k: opened: https://github.com/mozilla-services/autograph/issues/206

tools/autograph-client/client.go
g-k: `-a` says it's for the sha256 hash

g-k:
ok tests are passing.

@jvehent do you want to r? again or should I just land this?

signer/rsapss/rsapss.go
```diff
@@ -0,0 +1,170 @@
...
+	if conf.PublicKey == "" {
+		return nil, errors.New("rsapss: missing public key in signer configuration")
+	}
+	s.key, _, s.rng, s.PublicKey, err = conf.GetKeysAndRand()
+	if err != nil {
+		return nil, errors.Wrapf(err, "rsapss: error fetching key and rand from signer configuration")
+	}
+
+	switch s.key.(type) {
+	case *rsa.PrivateKey, *crypto11.PKCS11PrivateKeyRSA:
```
jvehent: nit: crypto11 isn't really needed here. Signers can simply rely on the crypto.Signer interface those keys implement, and you can actually rewrite this as follows:
```go
switch s.pub.(type) {
case *rsa.PublicKey:
```

PR #235, opened 2019-01-23
XPI keys in hsm 210

### LINKED ISSUE #210: support HSM storage of XPI signing keys
blocks: https://bugzilla.mozilla.org/show_bug.cgi?id=1515173

g-k:
Changes:

* store issuer keys in HSM to fix #210 (per https://github.com/mozilla-services/autograph/issues/210#issuecomment-456455626 we'll still generate EEs outside the HSM due to stdlib `x509.CreateCertificate` only accepting keys from the rsa and ecdsa package)
* use HSM RNG for XPI signing when HSM is available
* add hsm ECDSA signer to softhsm docker image and autograph dev softhsm config
* fix the sign XPI CI job to fail when the build script fails
* add tests to the XPI signer to get coverage >75%
* combine CI integration tests
* migrate APK signing tests to script and run them against the docker image
* in CI save and load gzipped docker tarballs (since it looks like CPU is cheaper than network IO)

Planning to work on EE key gen and deletion using the HSM in another PR based off this one (which has gotten large enough).

r? @jvehent


signer/xpi/cose.go
```diff
@@ -65,41 +66,47 @@ func intToCOSEAlg(i int) (v *cose.Algorithm) {
 func (s *XPISigner) generateCOSEKeyPair(coseAlg *cose.Algorithm) (eeKey crypto.PrivateKey, eePublicKey crypto.PublicKey, err error) {
 	var signer *cose.Signer
 
+	switch key := s.issuerKey.(type) {
+	case *rsa.PrivateKey, *ecdsa.PrivateKey, *crypto11.PKCS11PrivateKeyRSA, *crypto11.PKCS11PrivateKeyECDSA:
+		// ok
+	default:
+		err = errors.Errorf("xpi: Cannot generate COSEKeypair for issuerKey type %T", key)
+		return
```
g-k: We could allow this, but might as well check signer validity.

tools/softhsm/extensions-ecdsa-pk8.pem
```diff
@@ -0,0 +1,6 @@
+-----BEGIN PRIVATE KEY-----
```
g-k: this is the result of `openssl pkcs8 -topk8 -nocrypt -in extensions-ecdsa.pem -out extensions-ecdsa-pk8.pem` where extensions-ecdsa.pem is the private key field from the signer `autograph.yaml`

signer/xpi/cose.go
```diff
@@ -65,41 +66,47 @@ func intToCOSEAlg(i int) (v *cose.Algorithm) {
...
+
 	switch coseAlg {
-	case nil:
-		err = errors.New("Cannot generate private key for nil cose Algorithm")
 	case cose.PS256:
 		var size = 2048
 		switch key := s.issuerKey.(type) {
-		case *ecdsa.PrivateKey:
-			if key.D.BitLen() > size {
-				size = key.D.BitLen()
```
g-k: This is never true for the common NIST curves defined in `crypto/elliptic` since the ECDSA private key / D param bit len will always be much less than 2048: https://play.golang.org/p/_DrG2-g2ZxF

It doesn't make sense to use this directly since it's 

It's also a weird case since we're generating an RSA EE for an ECDSA issuer.

Instead of dropping this case and just using the default like we do here we could:

a) try to compute the effective RSA key size for a give ECDSA key size and generate a key of that size
b) bump the default up to 4096 (but this would hurt perf if we start signing with a bunch of PS256)
c) make the default RSA key size configurable at a signer or mode level (e.g. webext's use 2048, but system addons use 4096)
d) keep the above case which is effectively using the default size

signer/xpi/x509.go
```diff
@@ -133,14 +133,32 @@ func (s *XPISigner) generateIssuerEEKeyPair() (eeKey crypto.PrivateKey, eePublic
 			return
 		}
 
+		newKey, ok := eeKey.(*rsa.PrivateKey)
```
g-k: we could try to dedup these case statements with very similar logic for rsa and ecdsa key types. e.g. pull out fns for `getIssuerRSAKeySize` and `getIssuerECDSACurve`

signer/xpi/cose.go
g-k: per chat with ulfr: we can worry about it later is we start using PS256 sigs.

signer/xpi/x509.go
g-k: fixed in 8c283b1c5ad8cb84f8985d71df9cf46f0212d4d0

g-k:
cert in f5024d3d0b6a1ad8002da3c61dffff8dcf2842e8 generated with:

`openssl req -newkey rsa:1024 -nodes -keyout rsa-1024.pem -extensions mysection -config myconfig.cnf -x509 -days 365 -out rsa-1024-certificate.pem`

where myconfig.cnf was:

```
[req]
req_extensions = v3_req
prompt = no
distinguished_name = req_distinguished_name
[req_distinguished_name]
C = US
ST = CA
L = Mountain View
O = Mozilla Addons
OU = Unit Test
CN = example.com
[v3_req]

[mysection]
keyUsage         = digitalSignature,keyCertSign
extendedKeyUsage = codeSigning
basicConstraints = critical,CA:true
```

signer/xpi/x509.go
```diff
@@ -117,12 +118,69 @@ func (s *XPISigner) makeTemplate(cn string) *x509.Certificate {
 	}
 }
 
+// getIssuerPubKey returns the public key for ECDSA or RSA keys from
+// the crypto stdlib or crypto11
+func (s *XPISigner) getIssuerPubKey() (pubKey crypto.PublicKey, err error) {
```
g-k: This and the following three functions could move to signer.

jvehent: Yeah, why not use the existing `signer.GetKeysAndRand` instead, and store the pubkey in the `XPISigner`?

signer/xpi/cose.go
```diff
@@ -4,15 +4,16 @@ import (
...
 	"crypto/ecdsa"
-	"crypto/rand"
 	"crypto/rsa"
 	"crypto/sha256"
 	"crypto/x509"
 	"encoding/base64"
 	"fmt"
 	"strings"
 
+	"github.com/ThalesIgnite/crypto11"
```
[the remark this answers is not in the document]
g-k: :+1: fixed in 69bdb3fd0d4b0f6867d65f15cde4adae6ee6b895

We're using it in the mar signer, but I can add a lint for that (probably just grepping for crypto11 in `signer/**/`) after switching that one to `GetKeysAndRand`.

g-k:
re r? @jvehent

PR #240, opened 2019-02-01
Add tool to make CSR using key in HSM

jvehent:


jvehent:
Good point, added a readme. re-r? plz

tools/makecsr/README.md
```diff
@@ -0,0 +1,8 @@
+makecsr
```
g-k: :+1:

Issue #147, opened 2018-09-27
nit: enforce JAR header length and max headers


g-k:
Per the spec:

> Implementations should support 65535-byte (not character) header values, and 65535 headers per file. They might run out of memory, but there should not be hard-coded limits below these values.

https://docs.oracle.com/javase/8/docs/technotes/guides/jar/jar.html#Notes_on_Manifest_and_Signature_Files

We're probably doing something wrong if our APKs are that big, but it's a potential DoS vector (memory usage).

Switching to APK v2 or v3 sigs would also address the problem, since I don't believe they require unpacking and reading the whole APK.  

Also, check whether Fx XPIs expect that and line length to be enforced (per https://github.com/mozilla-services/autograph/pull/145#issuecomment-425114934).

jvehent:
Is there more work to do here?

g-k:
For each of our two jar packing signers: apk v1 and XPI, there are two checks:

- [ ] headers are <= 65535 bytes long
- [ ] <= 65535 headers per file

apk v1 is deprecated, so we probably don't care.

for XPI:

- [x] header line length check added in https://github.com/mozilla-services/autograph/pull/241 added
- [ ] headers per file not checked but there is an implicit limit from memory and request size (of ~4GB of headers for one)
  - [ ] maybe find out whether the Fx implementation checks this?

I think max header lines check would be easy to add for XPIs to make that explicit, but also nice-to-have / nitpicky.

PR #241, opened 2019-02-07
Fix xpi jar line length

g-k:
Fix for https://github.com/mozilla/addons-server/issues/10588 and the line length part of #147

r? @jvehent

### DIFF
--- a/signer/apk/jar.go
+++ b/signer/apk/jar.go
@@ -19,7 +19,8 @@ import (
 // consts and vars for formatFilename
 const (
 	maxLineByteLen      = 72
-	maxContinuedByteLen = 70 // -1 for leading space and -1 for trailing \n newline
+	maxContinuedByteLen = 70    // -1 for leading space and -1 for trailing \n newline
+	maxHeaderBytes      = 65535 // max length for wrapped / multiline headers is 2 << 15 - 1
 )
 
 var maxFirstLineByteLen = maxLineByteLen - (len([]byte("Name: ")) + 1) // + 1 for a \n newline
@@ -42,6 +43,10 @@ func formatFilename(filename []byte) (formatted []byte, err error) {
 		filenameLen      = len(filename)
 		writtenFileBytes = 0 // number of bytes of the filename we've written
 	)
+	if filenameLen > maxHeaderBytes {
+		err = errors.Errorf("apk: filename length %d exceeds the wrappable limit %d", filenameLen, maxHeaderBytes)
+		return
+	}
 	if filenameLen <= maxFirstLineByteLen {
 		formatted = filename
 		return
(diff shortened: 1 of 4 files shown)


signer/xpi/jar.go
```diff
@@ -11,8 +11,62 @@ import (
 	"io"
 	"io/ioutil"
 	"strings"
+	"unicode/utf8"
 )
 
+// consts and vars for formatFilename
```
g-k: Might be good to dedup this code. I can make an issue for that.

signer/xpi/jar.go
```diff
@@ -86,13 +140,20 @@ func makeJARManifest(input []byte) (manifest []byte, err error) {
...
+			return manifest, err
+		}
 		h1 := sha1.New()
 		h1.Write(data)
-		fmt.Fprintf(mw, "SHA1-Digest: %s\n", base64.StdEncoding.EncodeToString(h1.Sum(nil)))
 		h2 := sha256.New()
 		h2.Write(data)
-		fmt.Fprintf(mw, "SHA256-Digest: %s\n\n", base64.StdEncoding.EncodeToString(h2.Sum(nil)))
+
+		fmt.Fprintf(mw, "Name: %s\nDigest-Algorithms: SHA1 SHA256\nSHA1-Digest: %s\nSHA256-Digest: %s\n\n",
```
g-k: Noticed that the apk jar packer uses `SHA-256` instead of `SHA256` and also doesn't specify digest-algorithms: https://github.com/mozilla-services/autograph/blob/master/signer/apk/jar.go#L103

signer/apk/jar.go
```diff
@@ -19,7 +19,8 @@ import (
 // consts and vars for formatFilename
 const (
 	maxLineByteLen      = 72
-	maxContinuedByteLen = 70 // -1 for leading space and -1 for trailing \n newline
+	maxContinuedByteLen = 70    // -1 for leading space and -1 for trailing \n newline
+	maxHeaderBytes      = 65535 // max length for wrapped / multiline headers is 2 << 15 - 1
```
jvehent: wut? 😱

signer/xpi/jar.go
jvehent: yes, xpi and apk use different manifest syntaxes

signer/apk/jar.go
g-k: :woman_shrugging: yeah hopefully we never hit that #147 https://searchfox.org/mozilla-central/source/security/apps/AppSignatureVerification.cpp#304

signer/xpi/jar.go
g-k: :+1: ok de-duping this might not be the best idea then.

PR #253, opened 2019-03-21
remote-settings dev config

jvehent:
FInally got around to this. I generated a new PKI using `tools/genpki` and make a new dev config using a `contentsignaturepki` signer.

You can use this via the hawk credentials for `alice` by calling the `remote-settings` signer.


jvehent:
So, turns out `monitor_test.go` breaks with `contentsignaturepki`. We didn't see that before because we didn't have a contentsignaturepki config in `autograph.yaml` (only in the softhsm config).
I'll refactor some code to make unit testing nicer here.

In the meantime, @leplatrem, you can simply copy/paste this patch into your local config to issue valid signatures for remote-settings (don't forget to change firefox prefs though).

g-k:
> We had an unexpected error preparing a VM for this build, potentially due to our infrastructure or cloud provider.  Please retry the build in a few minutes

CCI

:( restarted

jvehent:
> have the chains published somewhere public so that it works with our current client code

I somewhat naively assumed you'd use this for local dev and Firefox would follow the `file://` path. Is that not the case?

leplatrem:
> I somewhat naively assumed you'd use this for local dev and Firefox would follow the file:// path. Is that not the case?

I run docker in a container, so IDK, are they present inside the Docker filesystem?
If it's not much too much hassle having them in a public URL makes it easier (less params when running Docker :D)

jvehent:
Try `mkdir /tmp/autograph` on your local machine, then add `-v /tmp/autograph:/tmp/autograph` to the docker run command line.
The problem with making those certs available publicly is that we need a place to upload them into. Unlike the normandy config which uses statically configured certs, this configuration generates a signing cert at runtime. It's the future, but it does make local dev a little tricky. If mounting the local `/tmp/autograph` path works, it's a easier solution.

monitor_test.go
```diff
@@ -86,6 +91,29 @@ func verifyRsapssSignature(b64Sig, b64Key string) error {
...
+	// parse the json signature
+	sig, err := contentsignaturepki.Unmarshal(signature)
+	if err != nil {
+		return err
+	}
+	// make a templated hash
+	_, inputHash := contentsignaturepki.MakeTemplatedHash([]byte(input), sig.Mode)
+	if !ecdsa.Verify(eePubKey, inputHash, sig.R, sig.S) {
+		return fmt.Errorf("ecdsa signature verification failed")
+	}
```
jvehent: this fails with
```
--- FAIL: TestMonitorPass (2.81s)
    monitor_test.go:82: {Ref:1t9ngg0sz77go3t1rvm8ms4ztg Type:contentsignaturepki Mode:p384ecdsa SignerID:remote-settings PublicKey:-----BEGIN CERTIFICATE-----
        MIICWzCCAeKgAwIBAgIIFY4HHiYeU5AwCgYIKoZIzj0EAwMwXzELMAkGA1UEBhMC
        VVMxCzAJBgNVBAgTAkNBMRYwFAYDVQQHEw1Nb3VudGFpbiBWaWV3MRAwDgYDVQQK
        EwdNb3ppbGxhMRkwFwYDVQQDExBjc3Jvb3QxNTUzMTg2NzQ3MB4XDTE5MDEyMDE3
        NDU0N1oXDTI5MDMyMTE2NDU0N1owYDELMAkGA1UEBhMCVVMxCzAJBgNVBAgTAkNB
        MRYwFAYDVQQHEw1Nb3VudGFpbiBWaWV3MRAwDgYDVQQKEwdNb3ppbGxhMRowGAYD
        VQQDExFjc2ludGVyMTU1MzE4Njc0NzB2MBAGByqGSM49AgEGBSuBBAAiA2IABE/l
        1sHSqAN6/ghHNWS4cNh/ZMNzxiSk0VadD8qic3URaswhY0B6444mCOzDCrT4zYnT
        9M1ydiW2zm1VmHowVgrBgcb82WRNlnnnBvcYrLwmSgUf9fvhObejTqF/QG4J9qNq
        MGgwDgYDVR0PAQH/BAQDAgGGMBMGA1UdJQQMMAoGCCsGAQUFBwMDMA8GA1UdEwEB
        /wQFMAMBAf8wMAYDVR0eAQH/BCYwJKAiMCCCHi5jb250ZW50LXNpZ25hdHVyZS5t
        b3ppbGxhLm9yZzAKBggqhkjOPQQDAwNnADBkAjBPi4uhzbCzoFxYD2LZY+hUQ1kX
        eHE9YmF0NOBbosIY041UrbLQLpQRXi4LFku9008CMGoSf/UGB6qV9KzsCY3s51Lp
        Rw/oK58D2n63FpPnOTICWvWAYsG3HPNCoURfLqlEOQ==
        -----END CERTIFICATE-----
         Signature:4gvaqXhERK8I2JnfJsaJBSCOuAckGKCfpW3n5shn9q3noOYqXWmodxVHzgPBtyW7QHvCGggTHoVskZxDJFFUZKd7U4Ppf1UR2rYEdnkCsGZ9z0ET1fRK_hC5LrRipS6P SignedFile: X5U:file:///tmp/autograph/chains/remotesettingsdev/remote-settings.content-signature.mozilla.org-20190421.chain}
    monitor_test.go:83: verification of monitoring response 4 failed: ecdsa signature verification failed
```

but I'm failing to see why. 302 ☕️

leplatrem:
I've been trying to mount the `/tmp/autograph` folder without luck:

```
$ mkdir /tmp/autograph
$ chmod 777 /tmp/autograph
$ docker run --rm -v /tmp/autograph:/tmp/autograph:rw  --name autograph-server mozilla/autograph
```
gives me ``{"msg":"failed to add signer \"remote-settings\": contentsignaturepki: failed to make chain and x5u for end-entity: failed to upload chain: failed to make directory: mkdir /tmp/autograph/chains: permission denied"}}``

I thought it could come from my limited docker skills, but actually we do mount writable folder for Kinto attachments and it just works. So something else might be wrong

autograph.yaml
```diff
@@ -88,6 +88,58 @@ signers:
...
         -----END EC PRIVATE KEY-----
 
+    # Development remote settings content signer. Will issue a certificate
+    # valid for remote-settings.content-signature.mozilla.org. To use this
+    # with Firefox, set security.content.signature.root_hash to
+    # 5E:36:F2:14:DE:82:3F:8B:29:96:89:23:5F:03:41:AC:AF:A0:75:AF:82:CB:4C:D4:30:7C:3D:B3:43:39:2A:FE
+    #
+    # Note the private key of the root was not saved. Use tools/genpki to
+    # make a new config as needed.
+    - id: remote-settings
```
leplatrem: could we have a dedicated user in the config? :pray:

jvehent:
I'll get ya sorted

leplatrem:
Also, one doubt. 

My local setup involves Autograph, Remote Settings and Normandy. 

Using the specified root hash from this PR in Firefox, will I be able to verify signatures coming from both RemoteSettings-dev and Normandy-dev signers?

Thanks

jvehent:
That's a good point. If want to sign with both at the same time then we
need to do a different configuration. Do you already have a Docker compose?
If so, I could update it to integrate autograph in it and get you the full
config that works with both signers.

leplatrem:
> That's a good point. If want to sign with both at the same time then we need to do a different configuration. 

So right now, I have a working draft where if I change the root_hash between the moment I fetch records from remote-settings and validate the normandy signatures, then it seems to work.

(using [normandy dev hash](https://normandy.readthedocs.io/qa/environments.html#id3)) 

> Do you already have a Docker compose?
> If so, I could update it to integrate autograph in it and get you the full config that works with both signers.

I was running things manually so far. But if you need a Docker compose file I can make one. 
The autograph config I used is slightly modified from the one of this PR, since I had to put the `remote-settings` signer as the first one in `alice` authorizations.

leplatrem:
There are loads of things that would be required for this docker compose file to work from scratch (DB initialization mainly etc.), but this is how it'd look like at least:

https://gist.github.com/leplatrem/de2af43bbe867652e103c827307cc092#file-docker-compose-yaml

@jvehent is it enough for you to show me the light?

Thanks!

monitor_test.go
jvehent: I looked at this code in 10 different ways, printed all the intermediate values, checked the hash, and I still can't find why it fails... @g-k: would you mind having a look to see if I missed something here?

jvehent: FOUND IT! `certs[0]` is the intermediate, not the end-entity... 😫

main_test.go
```diff
@@ -48,7 +48,10 @@ func TestMain(m *testing.M) {
 		}
 	}
 	ag.makeSignerIndex()
-	log.Printf("autographer: %+v\n", ag)
+	// ok this is lame but the contentsignaturepki signer will upload a file dated
+	// at the given second so we don't want anything else to run at the same second
+	// otherwise that file may get rewritten. Easiest way to solve it? Sleep.
+	time.Sleep(time.Second)
```
jvehent: I will hang my head in shame at this one.... if you have any idea how to better handle this, I'm all ears.

autograph.yaml
```diff
@@ -58,35 +58,109 @@ signers:
...
-    - id: appkey3
-      type: contentsignature
+    # Development Normandy content signer. Will issue a certificate
+    # valid for normandy.content-signature.mozilla.org. To use this
+    # with Firefox, set security.content.signature.root_hash to
+    # 5E:36:F2:14:DE:82:3F:8B:29:96:89:23:5F:03:41:AC:AF:A0:75:AF:82:CB:4C:D4:30:7C:3D:B3:43:39:2A:FE
+    #
+    # Note the private key of the root was not saved. Use tools/genpki to
+    # make a new config as needed.
+    - id: normandy
```
leplatrem: This should be `normankey`

autograph.yaml
```diff
@@ -833,6 +907,8 @@ authorizations:
           - appkey2
           - appkey3
           - normankey
+          - normandy
```
leplatrem: This should be either `normankey` or `normandy`, but it should be associated to the `normandev` authorization

autograph.yaml
```diff
@@ -833,6 +907,8 @@ authorizations:
...
+          - remote-settings
```
leplatrem: Could you please add a dedicated user for remote-settings? If the signer is not the first in the list of authorized signers it does not pick it

autograph.yaml
jvehent: I'm removing "normankey" entirely, the "normandy" signer replaces it.

autograph.yaml
jvehent: I added a "kintodev" user that has the "remote-settings" signer.

main_test.go
```diff
@@ -287,8 +290,6 @@ func TestDefaultPort(t *testing.T) {
 }
 
 func TestPortOverride(t *testing.T) {
-	t.Parallel()
```
g-k: why can't we run the parser tests in parallel?

signer/contentsignaturepki/contentsignature.go
```diff
@@ -281,3 +281,29 @@ func (s *ContentSigner) getModeFromCurve() string {
 func (s *ContentSigner) GetDefaultOptions() interface{} {
 	return nil
 }
+
+// Verify takes the location of a cert chain (x5u), a signature in its
+// raw base64_url format and input data. It then performs a verification
+// of the signature on the input data using the end-entity certificate
+// of the chain, and returns a boolean indicating validity.
```
g-k: nit: looks like this is returning an error

main_test.go
jvehent: I disabled this while investigating and forgot the put it back. I added a commit that parallelizes (almost) all the tests

PR #260, opened 2019-04-01
DB ssl

g-k:
Changes:
* fix typo PG ssl mode `enabled` -> `enable`
* disable DB in CI w/ HSM (so we don't have two app instances talking to the same DB)
* in CI, save docker image cache to one file (so we can de-dupe layers) and bump base cache key (see commit message)
* Add SSLRootCert option to config and update docs
* Enabled DB SSL for local dev with a generated root CA and server cert
* Bundle rds cert bundle for use at `/usr/local/share/ca-certificates/rds-combined-ca-bundle.crt`
* use docker-compose in makefile and rename images from `*:build` tag to `autograph-` since `app:build` is super generic

Testing:

- [x] can run autograph-db and autograph-app-hsm locally w/ docker-compose

r? @jvehent


g-k:
:+1: removed the CSR files

For an inexact check with sample size of one it looks like saving one docker tarball to cache shaves ~2 minutes off the CCI build step: 

* 1:11 on this branch https://circleci.com/gh/mozilla-services/autograph/2302 

vs

* 3:21 on master in https://circleci.com/gh/mozilla-services/autograph/2276

PR #264, opened 2019-04-03
parallelize monitor signing

g-k:
Change:

* parallelize monitor signing requests per signer then aggregate the results or errors

Since we have 30+ signers in HSM prod taking ~1 minute. 

NB: I didn't get perf numbers to show a speedup.

r? @milescrabill @jvehent

### DIFF
--- a/monitor.go
+++ b/monitor.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"sync"
 	"time"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


monitor.go
```diff
@@ -35,33 +36,54 @@ func (a *autographer) handleMonitor(w http.ResponseWriter, r *http.Request) {
...
+				Type:      s.Config().Type,
+				Mode:      s.Config().Mode,
+				SignerID:  s.Config().ID,
+				PublicKey: s.Config().PublicKey,
+				Signature: encodedsig,
+				X5U:       s.Config().X5U,
+			}
+		}(i, s)
+	}
+	wg.Wait()
```
g-k: NB: this could hang until we hit the server timeout

Issue #163, opened 2018-10-30
rewrite autograph go CLI client


g-k:
re: https://github.com/mozilla-services/autograph/pull/162#discussion_r229039610

The CLI client should:

* default to trying to detect the file type and using the right signer (e.g. `client sign file`)
* optionally validate and have a validate subcommand (e.g. `client validate file`)
* allow for passing the file type and other params explicitly (e.g. `client sign --format xpi file`)
* compile to a binary users can install with `go get`
* share a common HTTP client library with autograph CLI, edge, and possibly monitor

PR #266, opened 2019-04-05
add verifyxpi tool

jvehent:
small tool that may come in handy.
@g-k should we do something about the options? I'm just passing an empty set for now.


g-k:
> should we do something about the options? I'm just passing an empty set for now.

Yeah probably. This won't verify COSE sigs https://github.com/mozilla-services/autograph/blob/master/signer/xpi/xpi.go#L559

g-k:
refs #163

jvehent:
Could the verify function automatically try COSE if a signature file exist?

g-k:
> Could the verify function automatically try COSE if a signature file exist?

Yep, we'll need to update update the verify logic to get the COSE algorithms from the parsed `cose.sig`

signer/xpi/xpi.go
```diff
@@ -556,16 +557,70 @@ func VerifySignedFile(signedFile signer.SignedFile, truststore *x509.CertPool, o
...
+type XpiManifest struct {
+	BrowserSpecificSettings struct {
+		Gecko struct {
+			ID string `json:"id"`
+		} `json:"gecko"`
+	} `json:"browser_specific_settings"`
+}
+
+// GetXpiID returns the string ID of an extension from its manifest
+func GetXpiID(xpiFile signer.SignedFile) (string, error) {
```
g-k: +1 might be worth validating the ID = OU in the PK7 sig too (we can open an issue for that)

signer/xpi/xpi.go
```diff
@@ -556,16 +557,70 @@ func VerifySignedFile(signedFile signer.SignedFile, truststore *x509.CertPool, o
...
-		err = verifyCOSEManifest(signedFile)
-		if err != nil {
-			return errors.Wrap(err, "xpi: error verifying COSE manifest for signed file")
-		}
-		err = verifyCOSESignatures(signedFile, truststore, opts)
-		if err != nil {
-			return errors.Wrap(err, "xpi: error verifying COSE signatures for signed file")
+	coseSig, err := readFileFromZIP(signedFile, coseSigPath)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "xpi: no COSE signature found, skipping")
```
g-k: nit: worth pulling this through L590 out into a `readOptionsFromCOSESig(signedFile signer.SignedFile) (opts Options, err error)` function to have it be at the same level of abstraction as the other fns called from this function

nit: For calls from autograph-client and unit tests where signing with COSE doesn't write a `cose.sig` this will fail silently, which is undesirable in IMO and should, at a minimum, be documented somewhere.

signer/xpi/xpi.go
jvehent: Yep. It could be done by diving in `xpi.Signature.p7.Certificates`, but that's a bit annoying (have to parse signerInfo to find the right cert) and I'm not sure it's worth it since we want to kill pkcs7 down the road.

signer/xpi/xpi.go
jvehent: I fixed the first nit by moving it to `xpi.ReadCOSEAlgsFromSig`. Not sure what to do about the second one... I did put a comment in there saying we fail silently for COSE, like Firefox does.

jvehent:
FYI this can't be merged in current form because it assumes the add-on is listed in `manifest.json` which is wrong. The canonical place to retrieve the add-on id is in the subject.cn of the end-entity cert in the pkcs7.

PR #269, opened 2019-04-08
tools: update vendored monitor deps

g-k:


g-k:
getting:

```
go test -v -covermode=count -count=1 -coverprofile=coverage_monitor.out go.mozilla.org/autograph/tools/autograph-monitor
panic: sql: Register called twice for driver postgres
``` in CI

Issue #267, opened 2019-04-06
multiple signature files in xpi


cunyx:
More a question than an issue, perhaps a feature request:

Being relatively new in most related areas, locally I tried to address some aspects of
supporting multiple signature files in a XPI archive
created with autograph (v. 3.0.2).

Might this be of interest for this repository?

If yes, I need to do some cleanup and I'll have to inspect how to handle
and work with this github thing.

Otherwise I'll have to keep patching new releases downstream.

g-k:
Hi @cunyx, what's your use case?

Generally speaking autograph supports Mozilla's signing needs first, but depending on the size and applicability of your patch we might be able to accept it.

With regard to supporting multiple signature files I think we'd need to address how to:

* reserve signature filenames to prevent people from to injecting unauthorized signatures
* specify which signature filenames and formats (i.e. PKCS7 or COSE at `META-INF/mozilla.rsa` and `META-INF/cose.sig` respectively) an XPI signer uses in its config

so it might be tricky.

cunyx:
> Hi @cunyx, what's your use case?

Thank you for the reply @g-k !

1.
As AddOn developer I'd like to xpi-sign my extension with the help of an own
instance of autograph to prove to the users, that the code has not changed
in the time between release and signing by Mozilla via addons.mozilla.org .

Currently the one and only signature guarantees integrity since
adding the Mozilla signature but this does not necessarily cover non developer
code changes after the release or edits by the (perhaps compromised)
signing authority.

Tolerating previous signatures needs cooperation from the autograph operated
by Mozilla, to not throw away existing signatures in the submitted xpi.


2.
a.
When distributing a browser, while not being Mozilla, and running an own
autograph instance to sign AddOns with an (additional) own root signing key,
shipped with a patched browser, the browser should trust and run AddOns,
that are signed (only) by Mozilla, (only) with my key or with both.

This would need the ability to have multiple signatures in the xpi and it
shouldn't matter if the xpi is first submitted to AMO or signed with the
browsers' key(s).

If this feature is only available in my autograph version but not in the one
operated by Mozilla, the xpi will loose "my" signatures when Mozilla adds her.
Therefore "my" signatures has always to be the last - that isn't ideal.

b.
As browser distributor, AddOn signer and autograph user, who is not Mozilla,
I'd like to not have to name my signature files "mozilla.sf" and "mozilla.rsa",
because I do not want to pretend to be Mozilla and perhaps violate trademarks etc.


[...]
[...]

PR #285, opened 2019-04-30
XPI owner option

cunyx:
After discussion in issue #267 the file name prefix for the xpi signature files *.sf and *.rsa is now configurable.


g-k:
Thanks @cunyx! I've got a lot on my plate at the moment, but I'll try to get to this in the next few weeks.

jvehent:
It pains me to close this pull request, because it's a significant amount of work and I think it's been very well implemented, but we're also not going to adopt this functionality in the near future.

I want to welcome contributors as much as possible, and rejecting well written code goes against that. I truly hope this does not discourage you from further contributing to Autograph and Mozilla.

My only advise, for next time, would be to make sure we're ready to accept a functionality before jumping into writing it.

Thanks again,
Julien

PR #293, opened 2019-05-24
Various cspki improvements

jvehent:
WIP


g-k:
@jvehent is this ready for review?

jvehent:
yeah let's land this and I'll make other PRs for the following patches. the rewrite of crypto11 means it'll take me a while to get the rest out the door...

g-k:
OK I wasn't sure if it was still WIP

signer/contentsignaturepki/contentsignature.go
```diff
@@ -110,48 +110,48 @@ func New(conf signer.Configuration) (s *ContentSigner, err error) {
...
 		tx, err = s.db.BeginEndEntityOperations()
 		if err != nil {
-			return nil, errors.Wrap(err, "contentsignaturepki: failed to begin end-entity db operations")
+			return nil, errors.Wrapf(err, "contentsignaturepki %q: failed to begin db operations", s.ID)
 		}
 	}
 	err = s.findAndSetEE(conf, tx)
-	if err != nil {
+	if err == nil {
+		log.Printf("contentsignaturepki %q: reusing existing EE %q", s.ID, s.eeLabel)
```
g-k: :+1:

signer/contentsignaturepki/contentsignature.go
```diff
@@ -110,48 +110,48 @@ func New(conf signer.Configuration) (s *ContentSigner, err error) {
...
+	} else {
+		// No suitable end-entity found, making a new chain
 		if err == database.ErrNoSuitableEEFound {
```
g-k: I think we could flatten these down to one `else if` ladder or `switch case` though I'm not sure if that's idiomatic

g-k:
I think we should stick to crypto11 <1.0 and upgrade later since that'll be a major change.

signer/contentsignaturepki/contentsignature.go
jvehent: funny enough, I did this a few minutes ago because I thought it looked ugly

g-k: :100: mind reader

signer/contentsignaturepki/contentsignature.go
```diff
@@ -84,10 +85,14 @@ func New(conf signer.Configuration) (s *ContentSigner, err error) {
 	if conf.ID == "" {
 		return nil, fmt.Errorf("contentsignaturepki %q: missing signer ID in signer configuration", s.ID)
 	}
-	if conf.PrivateKey == "" {
-		return nil, fmt.Errorf("contentsignaturepki %q: missing private key in signer configuration", s.ID)
+	if conf.IssuerPrivKey == "" {
+		return nil, fmt.Errorf("contentsignaturepki %q: missing issuer private key in signer configuration", s.ID)
 	}
-	s.issuerPriv, s.issuerPub, s.rand, _, err = conf.GetKeysAndRand()
+	// we need to parse or retrieve from the hsm the issuer private key,
```
g-k: A handle for the issuer private key, right? We shouldn't require it to be exportable to the autograph CU.

signer/contentsignaturepki/x509.go
```diff
@@ -76,7 +74,7 @@ func (s *ContentSigner) makeChain() (chain string, name string, err error) {
 	// valid for longer than that to account for clock skew
 	notAfter := time.Now().UTC().Add(s.validity + s.clockSkewTolerance)
 
-	block, _ := pem.Decode([]byte(s.PublicKey))
+	block, _ := pem.Decode([]byte(s.IssuerCert))
 	if block == nil {
 		err = errors.New("no pem block found in signer public key configuration")
```
g-k: nit: does this log msg need to be updated?

g-k:
Also, is this PR returning the EE pub key in the sig req response now?

jvehent:
Alright, I ended up adding https://github.com/mozilla-services/autograph/pull/293/commits/9e335e4ff4a2c132cb6776460dfa23468be6f7b2 to this PR after all. It's what we discussed yesterday wrt to multiple instances initializing in parallel without blocking, which will be required if we want to initialize signers on-demand rather that at init time.

jvehent:
@g-k r?

signer/contentsignaturepki/contentsignature.go
```diff
@@ -79,82 +80,109 @@ func New(conf signer.Configuration) (s *ContentSigner, err error) {
...
+			// some other error popped up, exit
+			return err
+		}
+		// create a label and generate the key
+		s.eeLabel = fmt.Sprintf("%s-%s", s.ID, time.Now().UTC().Format("20060102150405"))
+		s.eePriv, s.eePub, err = conf.MakeKey(s.issuerPub, s.eeLabel)
+		if err != nil {
+			return errors.Wrapf(err, "contentsignaturepki %q: failed to generate end entity", s.ID)
+		}
+		// make the certificate and upload the chain
```
g-k: nit: we should log the handle, ID, and label of any keys were generate

signer/contentsignaturepki/contentsignature.go
```diff
@@ -79,82 +80,109 @@ func New(conf signer.Configuration) (s *ContentSigner, err error) {
...
+		err = s.makeAndUploadChain()
+		if err != nil {
+			return errors.Wrapf(err, "contentsignaturepki %q: failed to make chain and x5u", s.ID)
+		}
```
g-k: nit: logging the chain and upload location on success would be useful

jvehent: done!

PR #295, opened 2019-05-29
Add recommendations file

g-k:
refs: https://bugzilla.mozilla.org/show_bug.cgi?id=1548647

Probably still need to (but those could be separate issues too):

- [ ] test resigning an already signed XPI 
- [ ] add support for recommendations request options to tools/client

r? @jvehent


autograph.yaml
```diff
@@ -900,6 +900,129 @@ signers:
...
+      # The signing parameters for each type of add-on are
+      # 'add-on' are signed with the OU 'Production' and the provided ID
+      # 'extension' are signed with the OU 'Mozilla Extensions' and the provided ID
+      # 'system add-on' are signed with the OU 'Mozilla Components' and the provided ID
+      # 'hotfix' are signed with the OU 'Production' and the ID 'firefox-hotfix@mozilla.org'
+      mode: add-on-with-recommendation
+      # RSA key gen is slow and CPU intensive, so we can optionally
+      # pregenerate and cache keys with a worker pool
+      recommendation:
+        path: "test-recommendation.json"
```
jvehent: perhaps this shouldn't be configurable?

signer/xpi/README.rst
```diff
@@ -33,17 +34,57 @@ Each signer must have a type, a mode and the certificate and private key of
...
-When a signature is requested, autograph will generate a private key and issue
-an end-entity certificate specifically for the signature request. The certificate
-is signed by the configured intermediate CA. The private key is thrown away
-right after the signature is issued.
+When a signature is requested, autograph will generate a private key
+and issue an end-entity certificate specifically for the signature
+request. The certificate is signed by the configured intermediate
+CA. The private key is thrown away right after the signature is
+issued. Note that if one XPI signer is including a recommendations
+file the other XPI signers should reserve that file path.
```
jvehent: Why? Are you concerned about signers overriding each other's recommendations, or something else?

signer/xpi/jar.go
```diff
@@ -320,6 +321,132 @@ func isCOSESignatureFile(name string) bool {
...
+	// Create a buffer to write our archive to.
+	buf := new(bytes.Buffer)
+
+	// Create a new zip archive.
+	w := zip.NewWriter(buf)
+
+	// Iterate through the files in the archive,
+	for _, f := range r.File {
+		if f.Name == filepath {
+			// reserved signature and other files do not get included in the manifest
```
jvehent: nit: comment need updating

signer/xpi/recommendation.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	if _, ok := r.Validity["not_before"]; !ok {
+		return errors.Errorf("xpi: recommendation validity with missing not_before field")
+	}
+	if _, ok := r.Validity["not_after"]; !ok {
+		return errors.Errorf("xpi: recommendation validity with missing not_after field")
+	}
+	if !r.Validity["not_before"].Before(r.Validity["not_after"]) {
+		return errors.Errorf("xpi: recommendation validity not_after must be after not_before field")
+	}
```
jvehent: I think it would also make sense to refuse recommendation that are valid for less than, say, 24 hours

signer/xpi/xpi.go
```diff
@@ -143,6 +171,8 @@ func New(conf signer.Configuration, stats *signer.StatsClient) (s *XPISigner, er
 	switch conf.Mode {
 	case ModeAddOn:
 		s.OU = "Production"
+	case ModeAddOnWithRecommendation:
+		s.OU = "Production"
```
jvehent: nit: collapse with case above:
```go
case ModeAddOn, ModeAddOnWithRecommendation:
```

signer/xpi/xpi.go
```diff
@@ -157,6 +187,19 @@ func New(conf signer.Configuration, stats *signer.StatsClient) (s *XPISigner, er
...
+		s.recommendationValidityRelativeStart = conf.RecommendationConfig.ValidityRelativeStart
+		s.recommendationValidityDuration = conf.RecommendationConfig.ValidityDuration
+		log.Infof("xpi: set recommendation options allowed states: %+v, path: %s, start duration: %s duration: %s",
+			s.recommendationAllowedStates,
+			s.recommendationFilePath,
+			s.recommendationValidityRelativeStart,
+			s.recommendationValidityDuration)
+	}
+	s.recommendationFilePath = conf.RecommendationConfig.FilePath
+	log.Infof("xpi: signer %s is ignoring recommendation file path %s", s.ID, s.recommendationFilePath)
```
jvehent: nit: use `%q` for strings so we know when an empty string is logged (it'll set `""`)

jvehent: why is it `ignoring` the recommendation file path ?

signer/xpi/xpi.go
```diff
@@ -224,6 +267,22 @@ func (s *XPISigner) SignFile(input []byte, options interface{}) (signedFile sign
...
+		input, err = appendFileToZIP(input, s.recommendationFilePath, recFileBytes)
+		if err != nil {
+			return nil, errors.Wrap(err, "xpi: error append recommendation file to XPI")
+		}
+	} else {
+		input, err = removeFileFromZIP(input, s.recommendationFilePath)
+		if err != nil {
+			return nil, errors.Wrap(err, "xpi: error removing recommendation file from XPI")
+		}
+	}
```
jvehent: perhaps `removeFileFromZIP` should always be invoked, because we want to remove existing recommendations anyway.
```go
	input, err = removeFileFromZIP(input, s.recommendationFilePath)
	if err != nil {
		return nil, errors.Wrap(err, "xpi: error removing recommendation file from XPI")
	}
	if s.Mode == ModeAddOnWithRecommendation {
		recFileBytes, err := s.makeRecommendationFile(opt, cn)
		if err != nil {
			return nil, errors.Wrap(err, "xpi: error making recommendation file from options")
		}
		input, err = appendFileToZIP(input, s.recommendationFilePath, recFileBytes)
		if err != nil {
			return nil, errors.Wrap(err, "xpi: error append recommendation file to XPI")
		}
	}
```

autograph.yaml
```diff
@@ -900,6 +900,129 @@ signers:
...
+        states:
+          standard: true
+          recommended: true
+          partner: true
```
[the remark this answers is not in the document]
g-k: I was thinking it'd be good to not let the client insert arbitrary strings in the recommendation file.

autograph.yaml
g-k: Depending on whether the proposed filename is used in a lot of existing XPIs my plan was to allow AMO to change it without requiring a code change. :woman_shrugging:

signer/xpi/README.rst
g-k: If we don't filter out the recommendations file someone could use a signer in mode "addon" and effectively get a recommended addon since it's the same OU.

addons-server should also filter out the file, but autograph shouldn't rely on that control.

signer/xpi/recommendation.go
g-k: Why? Doesn't seem like it's useful to an attacker.

signer/xpi/xpi.go
g-k: https://github.com/mozilla-services/autograph/pull/295#discussion_r289018352

PR #301, opened 2019-07-19
database: error instead of dying on monitor failure

g-k:
Code changes:

* change DB monitor to log errors instead of dying (only the csigpki signer uses the DB and it sets up and loads its config at init, so it's OK to briefly lose DB connectivity and we shouldn't crash the app)
* factor out DB init to addDB in main
* factor out CheckConnection func
* add connect/monitor tests
* add MonitorPollInterval to DB config
* distinguish the heartbeat handler from LB heartbeat and teach it to check HSM connectivity (it finds the first signer config with a non-PEM private key, caches it, and uses it to check HSM connectivity returning 502 bad gateway when the connection fails to remove it from the health LBs cc @milescrabill @Micheletto re: this behavior) 

Unit tests just check DB connected and HSM disconnected, but I functional tested that running app-hsm and hitting http://localhost:8001/__heartbeat__ returns 

```json
{"hsmAccessible": true}
```

r? @jvehent 

TODO assuming this lands:

- [ ] monitor logs for HSM and DB disconnects
- [ ] add monitorpollinterval to autograph configs

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -238,6 +238,47 @@ func handleLBHeartbeat(w http.ResponseWriter, r *http.Request) {
 	w.Write([]byte("ohai"))
 }
 
+// handleHeartbeat checks whether backing services are enabled and
+// accessible and returns 200 when they are and 502 when the
+// aren't. Currently it only checks whether the HSM is accessible.
+func (a *autographer) handleHeartbeat(w http.ResponseWriter, r *http.Request) {
+	if r.Method != "GET" {
+		httpError(w, r, http.StatusMethodNotAllowed, "%s method not allowed; endpoint accepts GET only", r.Method)
+		return
+	}
+	var (
+		// a map of backing service name to up or down/inaccessible status
+		result = map[string]bool{}
+		status = http.StatusOK
+	)
+
+	// try to fetch the private key from the HSM for the first
+	// signer conf with a non-PEM private key that we saved on
+	// server start
+	conf := a.hsmHeartbeatSignerConf
+	if conf != nil {
+		err := conf.CheckHSMConnection()
+		if err == nil {
+			result["hsmAccessible"] = true
+			status = http.StatusOK
+		} else {
+			log.Errorf("error checking HSM connection for signer %s: %s", conf.ID, err)
+			result["hsmAccessible"] = false
+			status = http.StatusInternalServerError
+		}
+	}
+
+	respdata, err := json.Marshal(result)
+	if err != nil {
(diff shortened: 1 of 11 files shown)


handlers.go
```diff
@@ -229,15 +229,54 @@ func (a *autographer) handleSignature(w http.ResponseWriter, r *http.Request) {
...
+		err := conf.CheckHSMConnection()
+		if err == nil {
+			result["hsmAccessible"] = true
+			status = http.StatusOK
+		} else {
+			log.Errorf("error checking HSM connection for signer %s: %s", conf.ID, err)
+			result["hsmAccessible"] = false
+			status = http.StatusBadGateway
+		}
+	}
```
jvehent: I think it'd be good to call `db.CheckConnection` here as well and just display the boolean status in the response without, for now, failing the heartbeat if it's not healthy.

handlers.go
```diff
@@ -229,15 +229,54 @@ func (a *autographer) handleSignature(w http.ResponseWriter, r *http.Request) {
...
+	conf := a.hsmHeartbeatSignerConf
+	if conf != nil {
+		err := conf.CheckHSMConnection()
+		if err == nil {
+			result["hsmAccessible"] = true
+			status = http.StatusOK
+		} else {
+			log.Errorf("error checking HSM connection for signer %s: %s", conf.ID, err)
+			result["hsmAccessible"] = false
+			status = http.StatusBadGateway
```
jvehent: perhaps a 500 internal server error is more appropriate to return than a 502, so those errors can't be confused with the 502s nginx may return?

handlers.go
g-k: I kind of want to return RT latency numbers but maybe that's a bad side channel to expose.

handlers_test.go
```diff
@@ -373,19 +373,78 @@ func TestHeartbeat(t *testing.T) {
...
+		}
+	}
+}
+
+func TestHeartbeatChecksHSMStatusFails(t *testing.T) {
+	// NB: do not run in parallel with TestHeartbeat
+	ag.hsmHeartbeatSignerConf = &ag.signers[0].(*contentsignature.ContentSigner).Configuration
+
+	expectedStatus := http.StatusBadGateway
+	expectedBody := []byte("{\n  \"hsmAccessible\": false\n}")
```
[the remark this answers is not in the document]
g-k: yeah I guess we should parse it as JSON and assert values, but I don't think it'll change much.

PR #307, opened 2019-07-22
rsapss: make hash and salt mode configurable

jvehent:


signer/rsapss/rsapss.go
```diff
@@ -64,6 +66,20 @@ func New(conf signer.Configuration) (s *RSAPSSSigner, err error) {
...
+		s.rsaPssSaltLength = rsa.PSSSaltLengthAuto
+	}
+	switch conf.RSAPSSHash {
+	case "sha1":
+		s.rsaPssHash = crypto.SHA1
+	case "sha256":
+		s.rsaPssHash = crypto.SHA256
+	default:
+		// default to sha256
+		s.rsaPssHash = crypto.SHA256
```
g-k: nit: existing signers use SHA1 so I'd prefer this to error as misconfigured instead or default to SHA1

signer/rsapss/rsapss.go
```diff
@@ -79,24 +95,23 @@ func (s *RSAPSSSigner) Config() signer.Configuration {
...
-	}
-
-	opts := &rsa.PSSOptions{
-		SaltLength: rsa.PSSSaltLengthEqualsHash,
-		Hash:       crypto.SHA1,
-	}
-
-	sigBytes, err := s.key.(crypto.Signer).Sign(s.rng, digest, opts)
+	// sha1 is 20, sha256 is 32
+	if len(digest) != 20 && len(digest) != 32 {
```
g-k: nit: using the pkg constants like sha1.Size would be clearer e.g. https://play.golang.org/p/_ncgf_SxFpT and let us drop the comment

g-k:
Closing in favor of #319

PR #320, opened 2019-08-12
Fix 317 hb timeouts

### LINKED ISSUE #317: add timeout for heartbeat DB and HSM connection checks

g-k:
fixes #317 

r? @jvehent

### DIFF
--- a/autograph.yaml
+++ b/autograph.yaml
@@ -31,6 +31,12 @@ statsd:
 #    maxidleconns: 10
 #    monitorpollinterval: 10s
 
+# When the DB or HSM is enabled use these timeouts for the
+# /__heartbeat__ handler
+heartbeat:
+    hsmchecktimeout: 10ms
+    dbchecktimeout: 15ms
+
 # The keys below are testing keys that do not grant any power
 signers:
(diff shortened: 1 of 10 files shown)


autograph.yaml
```diff
@@ -31,6 +31,12 @@ statsd:
 #    maxidleconns: 10
 #    monitorpollinterval: 10s
 
+# When the DB or HSM is enabled use these timeouts for the
+# /__heartbeat__ handler
+heartbeat:
+    hsmchecktimeout: 10ms
+    dbchecktimeout: 15ms
```
jvehent: isn't that a bit aggressive and prone to failing when the db or hsm slow down during maintenance?
I feel like 10x, even 100x, these values would still provided the functionality without the flakiness.

main.go
```diff
@@ -67,23 +67,20 @@ type configuration struct {
 	Signers        []signer.Configuration
 	Authorizations []authorization
 	Monitoring     authorization
+	Heartbeat      HeartbeatConfig
```
jvehent: nit: does `HeartbeatConfig` need to be exported?

PR #332, opened 2019-08-22
gpg2: inline stdin write

jvehent:


signer/gpg2/gpg2.go
```diff
@@ -193,6 +199,10 @@ func (s *GPG2Signer) SignData(data []byte, options interface{}) (signer.Signatur
 	defer os.Remove(tmpContentFile.Name())
 	ioutil.WriteFile(tmpContentFile.Name(), data, 0755)
 
+	// take a mutex to prevent multiple invocations of gpg in parallel
+	serializeSigning.Lock()
+	defer serializeSigning.Unlock()
```
g-k: is this lock freed if there's a pipe write or other error?

g-k:
Also, we should make an issue to unit or integration test the failure case.

jvehent:
> Also, we should make an issue to unit or integration test the failure case.

I'm not even sure how we could reproduce it reliably. Perhaps this is similar to https://github.com/mozilla-services/autograph/blob/master/database/queries_test.go#L11 ?

g-k:
Yeah might not be worth the effort.

Issue #247, opened 2019-03-06
add lint for crypto11 use in signers


g-k:
Remove remaining crypto11 usage from signers and add a lint to fail in CI if it's added back.

refs: https://github.com/mozilla-services/autograph/pull/235#discussion_r255174951

Eventually, it'd be nice to mprotect crypto11 as either its own process or to run signers in separate processes.

g-k:
The only uses are in the MAR signer:

```
signer/mar/mar.go:17:   "github.com/ThalesIgnite/crypto11"
signer/mar/mar.go:61:   case *crypto11.PKCS11PrivateKeyRSA:
signer/mar/mar.go:72:   case *crypto11.PKCS11PrivateKeyECDSA:
```

g-k:
> We're using it in the mar signer, but I can add a lint for that (probably just grepping for crypto11 in signer/**/) after switching that one to GetKeysAndRand.

Issue #64, opened 2018-01-31
Sign APK v2


jvehent:
Spec at https://source.android.com/security/apksigning/v2

Example code that verifies at https://github.com/avast/apkverifier

This should be added to the `FileSigner` interface of the apk signer.

g-k:
refs: https://bugzilla.mozilla.org/show_bug.cgi?id=1489950

PR #335, opened 2019-08-29
initial apk2 signer that shells out to apksigner

jvehent:
WIP


g-k:
refs: https://github.com/mozilla-services/autograph/issues/64

monitor.go
```diff
@@ -51,50 +51,63 @@ func (a *autographer) handleMonitor(w http.ResponseWriter, r *http.Request) {
...
+				sigresps[i] = formats.SignatureResponse{
+					Ref:        id(),
+					Type:       s.Config().Type,
+					Mode:       s.Config().Mode,
+					SignerID:   s.Config().ID,
+					PublicKey:  s.Config().PublicKey,
+					SignedFile: signedfile,
+					X5U:        s.Config().X5U,
+					SignerOpts: s.Config().SignerOpts,
+				}
```
jvehent: @g-k This is a new approach to monitor signers that only implement the FileSigner interface. It adds a new `TestFileGetter` that the signer implements to return a valid test file, in this case it's a pruned version of Focus.

We still don't have a way to verify the signed file (that would mean shelling out to apksigner in `autograph-monitor`), but at least this code exercises the signer and makes sure its configuration is good.

g-k:
can you run `make fmt-fix` and push up the changes?

monitor.go
g-k: WFM 

Long term, do we still want to switch the monitor to fargate (or having a separate file signer verifier) and shell out to `apksigner verify`? If so, I can make an issue for that.

.circleci/config.yml
```diff
@@ -33,7 +33,7 @@ jobs:
   unit-test:
     docker:
       - image: circleci/golang:1.13-buster
-      - image: circleci/postgres:11
```
g-k: What version are we running in prod? I think it'd be good to match that.

Are we using any postgres 12 features?

Dockerfile
```diff
@@ -9,17 +9,15 @@ RUN addgroup --gid 10001 app \
...
 # import the RDS CA bundle
 # https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_PostgreSQL.html#PostgreSQL.Concepts.General.SSL
-RUN curl -o /tmp/rds-combined-ca-bundle.pem https://s3.amazonaws.com/rds-downloads/rds-combined-ca-bundle.pem
-RUN openssl x509 -in /tmp/rds-combined-ca-bundle.pem -inform PEM -out /usr/local/share/ca-certificates/rds-combined-ca-bundle.crt
-RUN rm -f /tmp/rds-combined-ca-bundle.pem
-RUN update-ca-certificates
+RUN curl -o /tmp/rds-combined-ca-bundle.pem https://s3.amazonaws.com/rds-downloads/rds-combined-ca-bundle.pem && \
+    openssl x509 -in /tmp/rds-combined-ca-bundle.pem -inform PEM -out /usr/local/share/ca-certificates/rds-combined-ca-bundle.crt && \
+    rm -f /tmp/rds-combined-ca-bundle.pem && \
+    update-ca-certificates
```
g-k: :+1:

docker-compose.yml
```diff
@@ -1,6 +1,8 @@
 version: "3"
 volumes:
-  tmpdir:
+  apptmpdir:
+  hsmtmpdir:
+  hsmconfig:
```
g-k: :+1:

signer/apk2/apk2.go
```diff
@@ -0,0 +1,148 @@
...
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	"go.mozilla.org/autograph/signer"
+)
+
+const (
+	// Type of this signer is "apk2" represents a signer that
+	// shells out to apksigner to sign artifacts
+	Type = "apk2"
```
g-k: @jvehent what do you think about adding a version field to signers in the future?

That way we could avoid adding and removing whole signers for e.g. rsapss to genericrsa, apk to apk2, and contentsignature to contentsignaturepki. Also we could use that field to pick the underlying signing binary for the gpg2 and apk2 signers e.g. for apksigner updates in the future.

signer/apk2/apk2.go
```diff
@@ -0,0 +1,148 @@
...
+)
+
+// APK2Signer holds the configuration of the signer
+type APK2Signer struct {
+	signer.Configuration
+	minSdkVersion string
```
g-k: Is this the installed apksigner SDK version or the minimum client SDK version the signature should support?

signer/apk2/apk2.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "apk2: failed to get private key from configuration")
+	}
+	switch priv.(type) {
+	case *ecdsa.PrivateKey:
+		// ecdsa is only supported in sdk 18 and higher
+		s.minSdkVersion = "18"
+	default:
+		s.minSdkVersion = "9"
+	}
```
g-k: nit: might be useful to log which version we pick

tools/autograph-client/client.go
```diff
@@ -508,6 +515,9 @@ func verifyMAR(signedMAR []byte) bool {
 }
 
 func verifyPGP(input []byte, signature string, pubkey string) bool {
+	return true
```
g-k: nit: we might want to log that these are no-ops

tools/autograph-monitor/monitor.go
```diff
@@ -158,7 +167,7 @@ func Handler() (err error) {
 		case pgp.Type, gpg2.Type:
 			// we don't verify pgp signatures because that requires building a keyring
 			// using the public key which is hard to do using the current openpgp package
-			log.Printf("Skipping verification of PGP signature from signer %q", response.SignerID)
+			log.Printf("Skipping verification of PGP signature from signer %q (we can't verify those)", response.SignerID)
```
g-k: :+1:

docker-compose.yml
```diff
@@ -86,16 +90,18 @@ services:
       context: tools/autograph-monitor/
     environment:
       - AUTOGRAPH_URL=http://autograph-app-hsm:8001/
+      - AUTOGRAPH_KEY=19zd4w3xirb5syjgdx8atq6g91m03bdsmzjifs2oddivswlu9qs
```
g-k: +1 to making this explicit

g-k:
Once this is in stage we should make sure it doesn't blow up for concurrent requests like `gpg2`

.circleci/config.yml
jvehent: I _think_ we run 11 in prod with @Micheletto could confirm. I don't foresee using 12 features any time soon, so happy to roll it back if you think parity is more important than upgrading.

signer/apk2/apk2.go
jvehent: It's the minimum client sdk. I'll add a comment.

signer/apk2/apk2.go
jvehent: signers don't log currently, so while I agree with you, that seems like a bigger problem to solve (and one worth solving)

tools/autograph-client/client.go
jvehent: 👍

signer/apk2/apk2.go
jvehent: I need to think a little bit about how that would work. Wouldn't it make the logic inside the signer a lot more complicated though? And the tests as well?

monitor.go
jvehent: I think that would be worthwhile. And perhaps we could also integrate the work mgoodwin is doing around xpcshell in the monitor runtime.

.circleci/config.yml
g-k: I think it's better to test against the version we have deployed. We could test against postgres 11 and 12.

signer/apk2/apk2.go
g-k: > Wouldn't it make the logic inside the signer a lot more complicated though? And the tests as well?

More paths to cover, but less code :woman_shrugging:

I can split out another issue to discuss.

signer/apk2/apk2.go
g-k: > signers don't log currently

This is news to me and the XPI signer logs.

If there are other constraints or best practices we want to enforce like https://github.com/mozilla-services/autograph/issues/247 it'd be good to document them.

Makefile
```diff
@@ -28,47 +44,21 @@ tag: all
...
-	$(GO) vet go.mozilla.org/autograph/signer/rsapss
+	go vet $(PACKAGES)
 
 fmt-diff:
 	gofmt -d *.go database/ signer/ tools/autograph-client/ $(shell ls tools/autograph-monitor/*.go) tools/softhsm/ tools/hawk-token-maker/ tools/make-hsm-ee/ tools/makecsr/ tools/genpki/
 
 fmt-fix:
-	gofmt -w *.go database/ signer/ tools/autograph-client/ $(shell ls tools/autograph-monitor/*.go) tools/softhsm/ tools/hawk-token-maker/ tools/make-hsm-ee/ tools/makecsr/ tools/genpki/
+	go fmt $(PACKAGES)
+	gofmt -w tools/autograph-client/ $(shell ls tools/autograph-monitor/*.go) tools/softhsm/ tools/hawk-token-maker/ tools/make-hsm-ee/ tools/makecsr/ tools/genpki/
```
g-k: should we change this one to `go fmt` too?

Makefile
```diff
@@ -5,6 +5,22 @@ GO := go
...
+go.mozilla.org/autograph/signer/contentsignature \
+go.mozilla.org/autograph/signer/contentsignaturepki \
+go.mozilla.org/autograph/signer/mar \
+go.mozilla.org/autograph/signer/xpi \
+go.mozilla.org/autograph/signer/mar \
+go.mozilla.org/autograph/signer/pgp \
+go.mozilla.org/autograph/signer/gpg2 \
+go.mozilla.org/autograph/signer/genericrsa \
+go.mozilla.org/autograph/signer/rsapss
+
```
g-k: :+1: overdue

signer/apk2/apk2.go
jvehent: yeah, I don't know why I said that. we obviously log in the signers. 🤒

g-k: np

.circleci/config.yml
```diff
@@ -33,7 +33,7 @@ jobs:
   unit-test:
     docker:
       - image: circleci/golang:1.13-buster
-      - image: circleci/postgres:12
+      - image: circleci/postgres:10.6
```
jvehent: @g-k this is what we use in prod

Makefile
jvehent: go fmt only works on packages, not directories, so I didn't change it here but we could

PR #340, opened 2019-09-10
signer: use HSM RNG when available

g-k:
refs: https://bugzilla.mozilla.org/show_bug.cgi?id=1569471

i.e. not based on private key type

The HSM should produce more and higher quality random numbers and app servers should not block on boot waiting for RNG init (particularly for XPI key gen). Also, it reduces the risk of app server compromise allowing someone to bias the RNG (since an attacker would have to compromise the HSM RNG now).

This also gives us more breathing room for migrating all XPI signers to the HSM, since it depends on getting new certs for stage (for dep) and prod intermediate CAs, which might not happen in time for train-4.

One possible risk is exhausting the HSM RNG somehow, but however it's unlikely we'll hit CloudHSM's limit.

> Q: What is the ‘entropy source’ (source of randomness) for CloudHSM?

> Each HSM has a FIPS-validated Deterministic Random Bit Generator (DRBG) that is seeded by a True Random Number Generator (TRNG) within the HSM hardware module that conforms to SP800-90B. This is a high-quality entropy source capable of producing 20Mb/sec of entropy per HSM.

https://aws.amazon.com/cloudhsm/faqs/

### DIFF
--- a/signer/signer.go
+++ b/signer/signer.go
@@ -198,6 +198,15 @@ type Signature interface {
 // SignedFile is an []bytes that contains file data
 type SignedFile []byte
 
+// GetRand returns a cryptographically secure random number from the
+// HSM if available and otherwise rand.Reader
+func (cfg *Configuration) GetRand() io.Reader {
+	if cfg.isHsmAvailable {
+		return new(crypto11.PKCS11RandReader)
+	}
+	return rand.Reader
+}
+
 // GetKeysAndRand parses a configuration to retrieve the private and public key
 // of a signer, as well as a RNG and a marshalled public key. It knows to handle
 // HSMs as needed, and thus removes that complexity from individual signers.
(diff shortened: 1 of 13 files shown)


g-k:
Note that this only applies to signers using `GetKeysAndRand` for their RNG.

signer/signer.go
```diff
@@ -211,26 +220,24 @@ func (cfg *Configuration) GetKeysAndRand() (priv crypto.PrivateKey, pub crypto.P
 		publicKeyBytes []byte
 		unmarshaledPub crypto.PublicKey
 	)
+	rng = cfg.GetRand()
```
jvehent: I wonder. Since you're touching that code, it is worth renaming `getkeysandrand` to `getkeys` and updating the signers to call `getrand` separately?

Issue #339, opened 2019-09-09
Record hash of signed file


jvehent:
When signing a file, record the sha256 hash of the signed file

Issue #338, opened 2019-09-09
log hashes of signed file output 


g-k:
per at @catlee in meeting

Applies to the signer types:

* gpg
* apk

PR #342, opened 2019-09-12
Log signed hashes

g-k:
* refactor computing hashes for logs
* add output hashes for signed files for #338 and #339 

r? @jvehent

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -37,6 +37,18 @@ type heartbeatConfig struct {
 	hsmSignerConf *signer.Configuration
 }
 
+// hashSHA256AsHex returns the hex encoded string of the SHA256 sum
+// the arg toHash bytes
+func hashSHA256AsHex(toHash []byte) string {
+	h := sha256.New()
+	h.Write(toHash)
+	return asHex(h.Sum(nil))
+}
+
+func asHex(toHex []byte) string {
+	return fmt.Sprintf("%X", toHex)
+}
+
 // handleSignature endpoint accepts a list of signature requests in a HAWK authenticated POST request
 // and calls the signers to generate signature responses.
 func (a *autographer) handleSignature(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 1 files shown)


handlers.go
```diff
@@ -158,9 +170,9 @@ func (a *autographer) handleSignature(w http.ResponseWriter, r *http.Request) {
 				httpError(w, r, http.StatusInternalServerError, "encoding failed with error: %v", err)
 				return
 			}
-			// convert the input hash to hexadecimal for logging
-			hashlog = fmt.Sprintf("%X", input)
-
+			// the input is already a hash just convert it to hex
+			inputHash = asHex(input)
```
jvehent: perhaps this is just me, but I find the `asHex` more confusing than just calling `fmt.Printf` here.

handlers.go
```diff
@@ -195,20 +205,20 @@ func (a *autographer) handleSignature(w http.ResponseWriter, r *http.Request) {
...
-			"user_id":    userid,
-			"t":          int32(time.Since(starttime) / time.Millisecond), //  request processing time in ms
+			"rid":         rid,
+			"options":     sigreq.Options,
+			"mode":        sigresps[i].Mode,
+			"ref":         sigresps[i].Ref,
+			"type":        sigresps[i].Type,
+			"signer_id":   sigresps[i].SignerID,
+			"input_hash":  inputHash,
+			"output_hash": outputHash,
```
jvehent: `input_hash` is common to all three endpoints, but `output_hash` is only set by `/sign/file` and I can see us going through the logs wondering why `output_hash` is empty without noticing the call is for `/sign/hash` or `/sign/data`. I'm wondering if renaming this field to `signed_file_hash` would reduce this confusion. What do you think?

g-k: I was thinking we'd want to log output hash for `/hash` and `/data` and use a standard field name, but was running into errors and punted on the idea thinking we could add them later if desired. What about an explicit string like `"unimplemented"` so it's clearer in the logs?

I can take another shot at `/data` since I think that'd be helpful.

g-k:
:+1: I'm going to rebase and autosquash

PR #344, opened 2019-09-23
cspki: add pubkey initialization check

jvehent:


signer/contentsignaturepki/x509.go
```diff
@@ -37,6 +37,10 @@ func (s *ContentSigner) findAndSetEE(conf signer.Configuration) (err error) {
 		err = errors.Wrapf(err, "found suitable end-entity labeled %q in database but not in hsm", s.eeLabel)
 		return
 	}
+	if len(s.PublicKey) < 50 {
```
g-k: Should we check this in `GetKeys` / will we ever use a public key < 50 bytes?

signer/contentsignaturepki/x509.go
```diff
@@ -37,6 +37,10 @@ func (s *ContentSigner) findAndSetEE(conf signer.Configuration) (err error) {
...
+		err = errors.Errorf("malformed public key in key label %q for signer %q: %q", s.eeLabel, s.ID, s.PublicKey)
```
g-k: Worth mentioning how the keys is malformed in the error message.

signer/contentsignaturepki/x509.go
jvehent: makes sense, moved it there

signer/contentsignaturepki/x509.go
jvehent: changed the message to mention it, but at that point we don't know _why_ it's malformed

g-k: well why we're considering it malformed

PR #345, opened 2019-09-24
handlers: log keyid on successfully signing op

g-k:
For debugging signing requests

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -206,6 +206,7 @@ func (a *autographer) handleSignature(w http.ResponseWriter, r *http.Request) {
 			"ref":        sigresps[i].Ref,
 			"type":       sigresps[i].Type,
 			"signer_id":  sigresps[i].SignerID,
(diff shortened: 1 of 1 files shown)


handlers.go
```diff
@@ -206,6 +206,7 @@ func (a *autographer) handleSignature(w http.ResponseWriter, r *http.Request) {
 			"ref":        sigresps[i].Ref,
 			"type":       sigresps[i].Type,
 			"signer_id":  sigresps[i].SignerID,
+			"key_id":     sigreq.KeyID,
```
jvehent: I'm confused. Under which circumstances would the keyid be different from the signerid?

g-k: I think we were trying to debug signer resolution for the `-` vs `_` csigpki discrepancy.

I mostly want to associate signing HTTP request param with each requested signature in that HTTP request.

g-k:
Closing. Going to rethink.

PR #359, opened 2019-10-09
Bz 1587530 ev csr

hwine:
bit of a hack, compiles locally - can't run until on prod instance


hwine:
I went with `make-ev-csr`, and restored the original, so we don't have to modify the existing docs :)

PR #369, opened 2019-10-10
Dockerfile tweaks

jvehent:


jvehent:
This took me down a rabbit hole that ended in sadness and sorrow... tldr; let's not do this, it's a bad idea.

Pruning the main container works, but it breaks the softhsm container that we build in CI and use in docker-compose, because that container inherits from the main container, and doesn't like having `/go/src` and `/usr/local` deleted.

Trying to be a smart ass, I thought perhaps I could consolidate everything into a single container. So I moved the softhsm and monitor setup into the main autograph container. But Softhsm needs setup, so I moved that too. And softhsm has permissions, so I started adding the `app` user to the `softhsm` group and... well, that's where I stopped. All that shenanigans is for dev, and the main container is used in production, so we really don't want weird groups and unused configs sitting there.

Back to the drawing board.

PR #386, opened 2019-10-25
Test profiles

### LINKED ISSUE #385: support golang runtime monitoring
For debugging and perf tracking it'd be useful to be able to enable golang runtime monitoring in the shipped autograph server container.

specifically profilers from https://github.com/golang/go/wiki/Performance#debugging-performance-issues-in-go-programs (which refs https://blog.golang.org/profiling-go-programs too).

g-k:
Changes:
* update make to detect new unvendored golang packages and add test* and showcoverage* targets for it (without autocomplete)
* refactor makefile to use golang package names and save them to `testprofiles/$PKG_NAME` (renames the following targets (from on the left result on the right)

testautographdb testdatabase
testautographformats testformats
testcs contentsignature
testcspki contentsignaturepki

and similarly for showcoverage* of the same package names). This should make it easier to add other types of test profile.
* env vars to enable the http pprof endpoints and configure a golang runtime variables (fixes #385)
* add data race detector for tests ~~and enable it in CI~~ (todo later once test failures are fixed)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,6 +73,7 @@ test%: PACKAGE_TEST_OUTPUT_DIR = $(subst test,testprofiles/,$@)
 test%:
 	mkdir -p $(PACKAGE_TEST_OUTPUT_DIR)
 	$(GOTEST) -outputdir "testprofiles/$(PACKAGE_NAME)" $(PACKAGE_PATH)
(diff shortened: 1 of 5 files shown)


g-k:
not sure what coveralls is doing but this PR drops coverage 0.8%

r? @jvehent

Issue #96, opened 2018-06-06
Store authorizations in database


jvehent:
Instead of storing authorizations in a config yaml, Autograph should rely on a database to allow for new authorization to be created without redeploying the service.

This would mean storing hawk tokens also in database, so we should consider the security implications of doing so.

g-k:
Per discussion during my 1:1 with ulfr today:

> This would mean storing hawk tokens also in database, so we should consider the security implications of doing so.

This is considered an acceptable risk given proper DB controls (separate RO user, auditing and alerting, etc.).

> Also unlike Digest, this scheme is not intended to protect the key itself (the password in Digest) because the client and server must both have access to the key material in the clear.

https://hapi.dev/family/hawk/?v=7.1.2

However, as noted above the server will still need access to the HAWK keys (and not just a hash of it). We could encrypt authorizations in the DB.

> It would allow us to remove the sops integration, and by doing so also remove the large dependency tree that goes with it.

We would still need sops for HSM creds, and signers with private keys stored in the config, but it does get use closer to removing sops.

In the event of a lost auth token, we could revoke or rotate it without redeploying, which would allow us to continue serving traffic without a full redeploy (so faster IR and reduced downtime).

We'll probably want some sort of way to handle DB migrations. We don't want to ship a full ORM at this point, so a dev tool that generates SQL (up and down scripts) might be useful.

g-k:
For dev configs for downstream services (like AMO and kinto) we'll probably want to continue to support the config file or possibly have a script to populate a sqlite DB in dev and ship a separate dev docker image.

In the config file the `authorizations` block currently stores:

* `id` (the auth ID; should be a string matching `[-_a-z0-9]{1-255}`)
* `key` (the shared HAWK secret: string though it could be bytes; per above we might want to encrypt this)
* `hawktimestampvalidity` a string golang can parse into a `time.Duration` representing the allowed HAWK ts skew
* `signers` an array of signer IDs as strings (this could be a postgres array or a join table to a separate table of signer IDs)

g-k:
perf-wise we probably want autograph to load the authorizations at boot and possibly for each request too (overhead should be <10ms)

jvehent:
> perf-wise we probably want autograph to load the authorizations at boot and possibly for each request too (overhead should be <10ms)

How about querying the DB if the authorization isn't found in memory cache?

g-k:
possibly should consider an extensible design to support #118

g-k:
[...]

PR #401, opened 2019-11-08
Fix 96 auths in db

g-k:
fixes: #96 

Changes:
* add flag to dump auths to sql for migrations (NB: at the moment this only adds new auths it doesn't update existing ones with the same IDs)
* move main `authorization` struct to a shared struct in formats
  * remove `authorization.hawkMaxTimestampSkew` in favor of parsing a `time.Duration` directly and enforcing the 1 minute min skew
* add a flag to the config to read auths from the DB and use it for the HSM integration test
* add and validate hawk auth ID and Key field in `formats` and move signer ID validation there too. `formats` still only uses the stdlib. (moving the signerID was preferable adding `signer` as a dependency)
* adds a dev tool for later use (pulling it out without messing up the makefile history would be messy)

Considerations:
* auths are only loaded on app start (so after upating auths we require a restart to eliminate TOUTOC but also don't incur a perf impact beyond startup time)
* adds a new `autographdbauthsadmin` user in dev
* adds the softhsm auths to the default DB
* a migration procedure for deploys isn't defined
* auths are normalized into a hawk creds, signers, and an authorizations join table with the goal of eventually moving non-confidential signer config to the signer table (i.e. just have a list of keys in the encrypted config). This might prove unwieldy, but the table sizes are tiny at the moment and we could switch to a materialized view if that'd help.
* the signers and hawk creds use the unique IDs as their primary (instead of say an autoincrementing integer PK)
* hawk secrets are stored unencrypted (can't hash because hawk requires the full secret), so we rely on the underlying db and network security


g-k:
Initial feedback from Julien was to hit the DB for each request.

Pending poor perf: possible optimizations include an `updated_at` column and delayed expiration f auths from a local cached but going that way without profiling would be premature

main.go
```diff
@@ -105,6 +106,7 @@ func parseArgsAndLoadConfig(args []string) (conf configuration, listen string, a
 	fset.StringVar(&cfgFile, "c", "autograph.yaml", "Path to configuration file")
 	fset.StringVar(&port, "p", "", "Port to listen on. Overrides the listen var from the config file")
 	fset.BoolVar(&authPrint, "A", false, "Print authorizations matrix and exit")
+	fset.BoolVar(&authSQL, "auth-to-sql", false, "Print authorizations as SQL to stdout")
```
jvehent: does this imply `-A` ?
the naming isn't super clear. I thought it meant autograph would use the db to manage auth.

main.go
```diff
@@ -278,11 +300,16 @@ func (c *configuration) loadFromFile(path string) error {
 	return nil
 }
 
+// isDBEnabled returns whether the database is enabled in the config
+func (c *configuration) isDBEnabled() bool {
```
jvehent: nit: `isDBConfigured` ?

Makefile
```diff
@@ -19,11 +19,23 @@ install-cover:
 install-goveralls:
 	$(GO) get github.com/mattn/goveralls
 
+install-migrate:
+	$(GO) get -tags 'postgres' -u github.com/golang-migrate/migrate/v4/cmd/migrate/
```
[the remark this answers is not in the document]
g-k: fetch with postgres support:

> These examples build the cli which will only work with postgres. In order to build the cli for use with other databases, replace the postgres build tag with the appropriate database tag(s) for the databases desired. The tags correspond to the names of the sub-packages underneath the database package.

https://github.com/golang-migrate/migrate/tree/master/cmd/migrate#notes


```
-tags tag,list
	a comma-separated list of build tags to consider satisfied during the
	build. For more information about build tags, see the description of
	build constraints in the documentation for the go/build package.
	(Earlier versions of Go used a space-separated list, and that form
	is deprecated but still recognized.)
```

https://golang.org/cmd/go/

main.go
g-k: No I'll make the docstring clearer.

jvehent:
Tried this tonight, starting with a blank postgresql database, I ran the following commands:
```bash
$ make database/schema.sql database/auths.sql
$ sudo postgres
> psql
postgres=# CREATE DATABASE autograph;
postgres=# \i database/schema.sql
postgres=# \i database/auths.sql
```
I then enabled the following database config in `autograph.yaml`
```yaml
database:
   name: autograph
   user: myautographdbuser
   password: myautographdbpassword
   host: db:5432
   #sslmode: verify-full
   #sslrootcert: /etc/ssl/certs/db-root.crt
   maxopenconns: 100
   maxidleconns: 10
   monitorpollinterval: 10s
```
and ran autograph, which failed with the following logs:
```bash
$ go run go.mozilla.org/autograph -c autograph.yaml 
{"Timestamp":1581382620514460345,"Time":"2020-02-11T00:57:00Z","Type":"app.log","Logger":"autograph","Hostname":"gator5","EnvVersion":"2.0","Pid":187276,"Severity":6,"Fields":{"msg":"loaded unencrypted config from autograph.yaml with sha256sum 566dc9de372e4f0003abaa1c87a0b533c638693fc68dda1c0cd1699c0aad5ef3"}}
{"Timestamp":1581382620517229815,"Time":"2020-02-11T00:57:00Z","Type":"app.log","Logger":"autograph","Hostname":"gator5","EnvVersion":"2.0","Pid":187276,"Severity":6,"Fields":{"msg":"database connection established"}}
{"Timestamp":1581382620517422740,"Time":"2020-02-11T00:57:00Z","Type":"app.log","Logger":"autograph","Hostname":"gator5","EnvVersion":"2.0","Pid":187276,"Severity":6,"Fields":{"msg":"starting DB monitor polling every 10s"}}
{"Timestamp":1581382620521689810,"Time":"2020-02-11T00:57:00Z","Type":"app.log","Logger":"autograph","Hostname":"gator5","EnvVersion":"2.0","Pid":187276,"Severity":6,"Fields":{"msg":"Statsd enabled at 127.0.0.1:8125 with namespace autograph."}}
{"Timestamp":1581382749709911591,"Time":"2020-02-11T00:59:09Z","Type":"app.log","Logger":"autograph","Hostname":"gator5","EnvVersion":"2.0","Pid":187276,"Severity":2,"Fields":{"msg":"failed to add signer \"normandy\": contentsignaturepki \"normandy\": failed to initialize end-entity: contentsignaturepki \"normandy\": failed to find suitable end-entity: found suitable end-entity labeled \"\" in database but not in hsm: crypto11: could not find PKCS#11 key"}}
exit status 1
```

The `endentities` table is empty, so I'm not sure where autograph is finding a suitable label...
```sql
autograph=# select * from endentities;
 id | label | hsm_handle | signer_id | is_current | x5u | created_at 
----+-------+------------+-----------+------------+-----+------------
(0 rows)
```

g-k:
will revisit after #406

PR #406, opened 2019-11-15
Refactor auth backend

g-k:
Refactoring for #96 

* Move the following autographer data structures:

```golang
signers       []signer.Signer
auths         map[string]authorization
signerIndex   map[string]int
```

to an in-memory implementation of a swappable authbackend interface that hides the signerIndex implementation details and update callers accordingly
* remove `makeSignerIndex` from the `autographer` and index as auths are added
* expose a `getSignerForUser(userID, signerID string) (signer.Signer, error)` method instead of `getSignerID` (which depends on access to the autographer signers and `signerIndex`)

r? @jvehent


authorize.go
```diff
@@ -63,7 +64,11 @@ func (a *autographer) authorizeHeader(r *http.Request) (auth *hawk.Auth, userid
 	if err != nil {
 		return nil, "", err
 	}
-	hawk.MaxTimestampSkew = a.auths[userid].hawkMaxTimestampSkew
+	hawkSkewAuth, err := a.getAuthByID(userid)
+	if err != nil {
+		return nil, "", errors.Wrapf(err, "error finding auth for id %s for hawk.MaxTimestampSkew", userid)
+	}
+	hawk.MaxTimestampSkew = hawkSkewAuth.hawkMaxTimestampSkew
```
jvehent: this doesn't need to be fixed here, but I somewhat question the value of setting the skew parameter differently for each signer vs having a global one for autograph.

handlers.go
```diff
@@ -130,28 +130,31 @@ func (a *autographer) handleSignature(w http.ResponseWriter, r *http.Request) {
...
+		// returns an error if the signer is not found or if
+		// the user is not allowed to use this signer
+		requestedSigner, err := a.authBackend.getSignerForUser(userid, sigreq.KeyID)
+		switch err {
+		case nil:
+			// got the signer and allowed to use it
+		default:
 			httpError(w, r, http.StatusUnauthorized, "%v", err)
 			return
 		}
```
jvehent: did you mean to use `if err != nil` ?

memory_backend.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// getSignerId returns the signer identifier for the user. If a keyid
+// is specified, the corresponding signer is returned. If no signer is
+// found, an error is returned and the signer identifier is set to -1.
+func (b *inMemoryBackend) getSignerID(userid, keyid string) (int, error) {
+	tag := getSignerIndexTag(userid, keyid)
+	if _, ok := b.signerIndex[tag]; !ok {
+		if keyid == "" {
+			return -1, errors.Errorf("%q does not have a default signing key", userid)
```
jvehent: I wonder if this error is accurate. The default is just the first authorized signer for the user, so are we really saying that the user is not permitted to use any signer?

jvehent: and given line 139, it should be impossible to reach this code, right?

authorize.go
g-k: yeah captured that in https://github.com/mozilla-services/autograph/issues/414

memory_backend.go
g-k: This is from https://github.com/mozilla-services/autograph/blob/ebb4dffb0ae0ba89c6a79f58d225c5f920488205/authorize.go#L150-L159

> we really saying that the user is not permitted to use any signer?

Yes, and we said we didn't want any dangling users. I'm open to suggestions for how to reword it.

> and given line 139, it should be impossible to reach this code, right?

Yes, assuming auths aren't modified or reloaded since the app is loaded. This PR exists, because we said (in 1:1 or chat) we wanted to load auths on each request https://github.com/mozilla-services/autograph/pull/401#issuecomment-553929203

handlers.go
g-k: Yep! Had the type switch since I was catching other types in development.

errors.go
```diff
@@ -15,6 +16,9 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+// ErrAuthNotFound is for when autographer.getAuthByID doesn't find an auth
+var ErrAuthNotFound = errors.New("authorization with id %q not found")
```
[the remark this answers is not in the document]
g-k: :+1: good catch. I'll just drop the param since the memory error isn't used.

g-k:
refs: #96

PR #409, opened 2019-11-22
Add troubleshooting instructions

hwine:
Documenting in repo, as some of the steps require a checkout, so the
docs will already be there.


hwine:
@g-k I haven't had a chance to verify all steps on a real box yet, so keeping the PR in draft state for now.

I would appreciate any general comments on format, etc. Especially terminology -- this seems to be another place where ``key_id`` is ambiguous. Any suggestion on clarifying?

PR #410, opened 2019-11-22
Correct doc file locations [ci skip]

hwine:
Image related files belong in the 'statics' directory.


hwine:
fyi, I tried the [ci skip] tag to avoid building a doc-only change.

PR #446, opened 2020-02-25
Move entirely to gomod, no more vendoring

jvehent:
This started with various outdated deps found with `go list -u -f '{{if (and (not (or .Main .Indirect)) .Update)}}{{.Path}}: {{.Version}} -> {{.Update.Version}}{{end}}' -m all 2> /dev/null`

And as I untangled the failures, I ended up removing all the `vendor` folders from autograph and autograph-monitor, moving source code outside GOPATH, and fixing all the bugs found along the way.

So as a result, we're completely done with vendoring, and 100% on gomod. Hence the dramatic file deletion in this PR.


g-k:
wonder if `go mod tidy` / https://bugzilla.mozilla.org/show_bug.cgi?id=1587843 would pick these changes up

Makefile
```diff
@@ -1,40 +1,36 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
-GO := go
-GOLINT := golint -set_exit_status
-GOTEST := $(GO) test -v -coverprofile coverage.out -covermode=count -count=1
-PACKAGE_NAMES := $(shell git grep -E -n --no-color -h '^package [a-z0-9]+$$' -- '*.go' ':!vendor/' ':!tools/' | cut -d ':' -f 2 | sed 's/package //g' | sed 's/main/autograph/g' | sort | uniq) monitor
-PACKAGE_PATHS := $(subst go.mozilla.org/autograph/signer/autograph,go.mozilla.org/autograph,$(subst go.mozilla.org/autograph/signer/monitor,go.mozilla.org/autograph/tools/autograph-monitor,$(subst go.mozilla.org/autograph/signer/signer,go.mozilla.org/autograph/signer,$(subst go.mozilla.org/autograph/signer/formats,go.mozilla.org/autograph/formats,$(subst go.mozilla.org/autograph/signer/database,go.mozilla.org/autograph/database,$(addprefix go.mozilla.org/autograph/signer/,$(PACKAGE_NAMES)))))))
-TEST_TARGETS := $(addprefix test,$(PACKAGE_NAMES))
```
g-k: yeah this is kinda wild

Makefile
```diff
@@ -45,93 +41,29 @@ show-lint:
...
-test%: PACKAGE_PATH = $(addprefix go.mozilla.org/autograph/signer/,$(subst test,,$@))
-test%: PACKAGE_TEST_OUTPUT_DIR = $(subst test,testprofiles/,$@)
-test%:
-	mkdir -p $(PACKAGE_TEST_OUTPUT_DIR)
-	$(GOTEST) -outputdir "testprofiles/$(PACKAGE_NAME)" $(PACKAGE_PATH)
-ifeq ($(RACE_TEST),1)
-	$(GO) test -v -race $(PACKAGE_PATH)
-endif
-ifeq ($(STATIC_CHECK),1)
-	$(GO) staticcheck $(PACKAGE_PATH)
```
g-k: note to @g-k: make issues for adding race and staticchecks back

Makefile
```diff
@@ -45,93 +41,29 @@ show-lint:
...
-showcoverageformats:
-showcoveragemonitor:
-showcoveragesigner:
-
-showcoverageapk:
-showcoveragecontentsignature:
-showcoveragecontentsignaturepki:
-showcoveragegenericrsa:
-showcoveragegpg2:
-showcoveragemar:
```
g-k: I'm OK with dropping non-xpi per package targets. We might want a util or helper script that filters on the go test `-run` arg to only run specific tests.

g-k:
If we're worried about un-vendoring reducing availability and impacting productivity, it's probably worth keeping a local checkout and build around too.

g-k:
The /files page for this PR makes GH UI or the browser unhappy (or both). It was very laggy.

Makefile
jvehent: oh that's a whoops. I'll add them back!

docker-compose.yml
```diff
@@ -11,6 +11,7 @@ services:
       context: database/
     environment:
       - POSTGRES_DB=autograph
+      - POSTGRES_HOST_AUTH_METHOD=trust
```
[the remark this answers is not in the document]
jvehent: yeah it's just for docker-compose so I don't think we have to worry

jvehent:
@g-k I made a few extra fixed to readd `race` and `staticcheck`, and do one more dep upgrade. Please review.

PR #448, opened 2020-02-26
docker: update AWS RDS CA cert

g-k:
Our current RDS cert expires March 6th and we want to update it in RDS and autograph app before then.

Per discussion w/ @Micheletto last night, we're planning to:

* land this and tag a point release
* upgrade the cert in RDS
* do an off-train deploy next week (so we can keep a working live stack around and limit the change set size)
 
Alternatives:

* land these changes in train-8
* lower the sslmode to not verify the cert
* mount the new cert into the image and add it as part of the deploy

r? @Micheletto @jvehent 

Follow up item:

- [ ] fix the cert filename in the docker image and config files


g-k:
Closing in favor of #449

PR #449, opened 2020-02-26
Convert to using the 2019 Root CA for RDS. 

Micheletto:
This change converts to using the 2019 Root CA for RDS (expires Aug 22, 2024) and fails if it expires in less than 90 days.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ RUN addgroup --gid 10001 app \
       apt-get clean
 
 # import the RDS CA Cert, and fail if it expires in < 90 days.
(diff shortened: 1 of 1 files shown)


g-k:
Chatted with @jvehent about this and looked at some docs.

* pq takes a path to a _bundle_ of PEM-encoded certs in the `sslrootcert` param
* the openssl command is popping off the first cert in the bundle and converting it from PEM to PEM
* https://manpages.debian.org/jessie/ca-certificates/update-ca-certificates.8.en.html is updating the OS certs, but pq doesn't

So we can:

- [ ] drop the openssl command
- [ ] fetch and save the bundle to `/usr/local/share/` without passing it through openssl
- [ ] point pq at the sslrootcert
- [ ] drop the call to `update-ca-certificates` since we don't want the OS trusting AWS for everything anwyay

PR #450, opened 2020-02-27
docker: just fetch the AWS RDS CA bundle

g-k:
per discussion in https://github.com/mozilla-services/autograph/pull/449


jvehent:
late r+

PR #454, opened 2020-03-09
in autograph-monitor, force gc to run after lambda invocation

jvehent:


tools/autograph-monitor/monitor.go
```diff
@@ -87,6 +88,11 @@ func main() {
 	if os.Getenv("LAMBDA_TASK_ROOT") != "" {
 		// we are inside a lambda environment so run as lambda
 		lambda.Start(Handler)
+		// force gc run
+		// https://bugzilla.mozilla.org/show_bug.cgi?id=1621133
+		t1 := time.Now()
+		runtime.GC()
+		log.Println("Garbage collected in", time.Now().Sub(t1))
```
jvehent: I don't think this will work because `lambda.Start` does not return after being called. I'll implement a wrapper instead.

PR #455, opened 2020-03-09
force gc in monitor, fix bug 1621133

jvehent:


g-k:
refs: https://bugzilla.mozilla.org/show_bug.cgi?id=1621133

g-k:
Feels silly that we have to do this and pay AWS for the extra execution time since the point of serverless is that this is taken care of.

PR #460, opened 2020-03-23
tcp healthcheck initial prototype

### LINKED ISSUE #459: TCP listener for NLB heartbeats
Expose a TCP listener that follows the HTTP heartbeat and fails when autograph is unhealthy, so that TCP load balancers like AWS NLB.

jvehent:
early prototype to fix #459 . Mostly looking for feedback on the approach at this stage.

f? @g-k @Micheletto

### DIFF
--- a/main.go
+++ b/main.go
@@ -191,7 +191,7 @@ func run(conf configuration, listen string, debug bool) {
 	}
 
 	ag.startCleanupHandler()
-	ag.startTcpHealthCheckHandler()
+	ag.startTCPHealthCheckHandler()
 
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -80,6 +81,7 @@ type autographer struct {
 	debug         bool
 	heartbeatConf *heartbeatConfig
 	authBackend   authBackend
+	isReady       bool
```
g-k: Might want to rename this to `lastHeartbeatOK` or at least document what it is / how it works

Micheletto:
Can probably close this one out?

PR #469, opened 2020-04-07
database: fix error handling in GetLabelOfLatestEE

jvehent:


database/connect.go
```diff
@@ -69,7 +69,11 @@ func Connect(config Config) (*Handler, error) {
 	if config.MaxIdleConns > 0 {
 		dbfd.SetMaxIdleConns(config.MaxIdleConns)
 	}
-	return &Handler{dbfd}, nil
+	h := &Handler{dbfd}
+	dbCheckCtx, dbCancel := context.WithTimeout(context.Background(), 5*time.Second)
```
g-k, line 73: +1 for adding a timeout

database/queries.go
```diff
@@ -53,8 +53,11 @@ func (db *Handler) GetLabelOfLatestEE(signerID string, youngerThan time.Duration
 				WHERE is_current=TRUE AND signer_id=$1 AND created_at > $2
 				ORDER BY created_at DESC LIMIT 1`,
 		signerID, maxAge).Scan(&label, &nullableX5U)
-	if err == sql.ErrNoRows {
-		return "", "", ErrNoSuitableEEFound
+	if err != nil {
+		if err == sql.ErrNoRows {
+			return "", "", ErrNoSuitableEEFound
+		}
+		return
```
g-k, line 60: ah non-`ErrNoRows` errors weren't returned

docker-compose.yml
```diff
@@ -49,6 +49,8 @@ services:
     volumes:
       - hsmtmpdir:/tmp/
       - hsmconfig:/app/
+    depends_on:
+      - db
```
g-k, line 53: :+1: 

does this let us drop the wait-for-it scripts?

jvehent: I don't think so. It'll just start the container after the DB, but provisioning might take time, so the `wait-for-it` script is still needed.

PR #498, opened 2020-07-14
Add test case for `formatFilename()` with control char

willdurand:
As per [this comment](https://github.com/mozilla/addons-server/issues/14817#issuecomment-658302096), I propose this test case that I initially wrote to determine whether control characters in filenames would be allowed.

### DIFF
--- a/signer/xpi/jar_test.go
+++ b/signer/xpi/jar_test.go
@@ -46,6 +46,22 @@ func TestFormatFilenameInvalidUTF8(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


willdurand:
I guess the failure is because my fork does not have a token for the coverage tool.

g-k:
Yeah, I just gave you write on the repo. Hopefully pushing directly makes goveralls happy.

willdurand:
@g-k ha yes, I pushed the same commit on this repo and circle-ci kicked in and updated the status of this PR!

g-k:
oh nice!

PR #509, opened 2020-08-06
added sponsored, verified, and line to states in config

eviljeff:
and removed partner because it was never used.  There might be other changes needed?


g-k:
Landing since unit tests pass, but fail to publish coverage data.

PR #506, opened 2020-07-31
Add monitor struct/handler

gterzian:
FIX #464

PR #522, opened 2020-08-24
gterzian master

### LINKED ISSUE #464: reduce monitor overhead
Currently, monitoring via the `/__monitor__` endpoint scales linearly with <# of monitor requests> and <# of signers>.

This:

* consumes CPU and memory representing a potential DoS vector 
* introduces latency for real signing requests especially for signers that shell out or lock and shell out like the `gpg2` signer (TODO: measure actual CPU and memory overhead)
* adds complexity to deployments i.e. what succeeds with one config might not work with another

g-k:
rebase of #506 (which fixes #464) with:

* 14bd70a to update the softhsm config for integration tests
* a0301e1 to parse to a `time.Duration` and use a hardcoded 5m default fixes: #524

This PR will also have passing unit tests since I'm kicking it off and the coverage secret will be reported.


g-k:
ugh blocked on https://github.com/mozilla-services/autograph/issues/523 now

PR #526, opened 2020-08-26
add recommended-android recommendation state; drop sponsored

eviljeff:
https://bugzilla.mozilla.org/show_bug.cgi?id=1661230 - to support mozilla/addons-server#15315


g-k:
unit test CI failing w/ the coverage CI token missing is expected since we don't inject secrets into CI runs from forks.

PR #603, opened 2021-02-22
change test addon to ublock origin and vendor it

g-k:
Change:

* Add a test XPI fixture to the built image and change it to be the default for `make integration-test`

Fixes existing CI jobs failing with:

```console
Creating project_app_run ... 
Creating project_app_run ... done--2021-02-21 05:23:56--  https://addons.mozilla.org/firefox/downloads/file/935711/pomodoro_clock-1.1.1-an+fx-windows.xpi
Resolving addons.mozilla.org (addons.mozilla.org)... 52.41.196.34, 44.235.202.42, 34.223.236.119
Connecting to addons.mozilla.org (addons.mozilla.org)|52.41.196.34|:443... connected.
HTTP request sent, awaiting response... 404 Not Found
```

and removes the CI dependency on AMO being up and serving the test addon


diox:
Hey @g-k, this probably started failing because we deleted duplicate files on add-ons in https://github.com/mozilla/addons-server/issues/16464 - relatively few public add-ons were affected, but I suspect the one you were using was.

I think you should avoid hardcoding AMO URLs - there is no guarantee that a particular XPI will always be available, and no guarantee that the download URL will stay the same either (it's not part of our supported API and could in theory change at any time without prior notice - in fact we were discussing doing that earlier today!). Why don't you commit a test xpi in the autograph repos directly ?

g-k:
> Why don't you commit a test xpi in the autograph repos directly ?

Yeah, that's a good idea. I think we were originally trying to avoid copying test XPIs into the images and working around CircleCI's lack of volume mounts

PR #626, opened 2021-03-08
more monitor refactoring

g-k:
All changes in `tools/autograph-monitor`

Changes:
* refactor csig test fixtures
* extract autograph CS dev root to `constants.go` 
* document csigpki `verifyCertChain`

### DIFF
--- a/tools/autograph-monitor/contentsignaturepki_test.go
+++ b/tools/autograph-monitor/contentsignaturepki_test.go
@@ -74,7 +74,7 @@ func TestVerifyWronglyOrderedChain(t *testing.T) {
 }
 
 func TestVerifyFirefoxRoot(t *testing.T) {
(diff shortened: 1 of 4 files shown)


tools/autograph-monitor/contentsignaturepki_test.go
```diff
@@ -136,12 +136,6 @@ var ValidMonitoringContentSignature = formats.SignatureResponse{
 	X5U:       "http://127.0.0.1:64320/normandychain",
 }
 
-// this is the trusted root ca for the firefox pki
-var FirefoxPKIRootPEM = []byte(firefoxPkiProdRoot)
```
g-k, line 140: These are public, but I don't think anything else is using the monitor (e.g. it's not an importable module via go.mozilla.org)

Issue #674, opened 2021-04-26
monitor S3 fetch connection resets


g-k:
We've seen connection resets for the monitor fetching the chain file e.g. 

```
Errors found during monitoring:
1. failed to retrieve x5u: Get https://content-signature-2.cdn.mozilla.net/chains/onecrl.content-signature.mozilla.org-2021-06-11-15-04-26.chain: read tcp 169.254.76.1:50044->99.84.79.65:443: read: connection reset by peer: errorString
null
```

- [ ] collect more data on when this occurs
- [ ] consider retrying the x5u fetch before failing or increasing timeouts or using the first party aws client (namely https://github.com/mozilla-services/autograph/blob/main/tools/autograph-monitor/contentsignaturepki.go#L44 which uses the default `http.Client` https://github.com/mozilla-services/autograph/blob/ab0219f9b8d3c9f5d800972387a2263d101c961c/signer/contentsignaturepki/upload.go#L70)

g-k:
refs: #594

g-k:
The first party aws client would tie us to S3 specifics and we'd like to be generic of HTTP accessible URLs, so pulling the config options from the S3 client might be the best option:

```
	// Default connection pool options
	DefaultHTTPTransportMaxIdleConns        = 100
	DefaultHTTPTransportMaxIdleConnsPerHost = 10

	// Default connection timeouts
	DefaultHTTPTransportIdleConnTimeout       = 90 * time.Second
	DefaultHTTPTransportTLSHandleshakeTimeout = 10 * time.Second
	DefaultHTTPTransportExpectContinueTimeout = 1 * time.Second

	// Default to TLS 1.2 for all HTTPS requests.
	DefaultHTTPTransportTLSMinVersion uint16 = tls.VersionTLS12
```

https://github.com/aws/aws-sdk-go-v2/blob/main/aws/transport/http/client.go#L12-L25

(used at https://github.com/aws/aws-sdk-go-v2/blob/main/service/s3/api_client.go#L207)

Issue #594, opened 2021-02-02
get x5u connection issues 


g-k:
From the monitor logs:

```
Errors found during monitoring:
1. failed to retrieve x5u: Get https://content-signature-2.cdn.mozilla.net/chains/onecrl.content-signature.mozilla.org-2021-03-22-17-58-02.chain: read tcp 169.254.76.1:53494->99.84.79.91:443: read: connection reset by peer: errorString
null
```

```
Errors found during monitoring:
1. failed to retrieve x5u: Get https://content-signature-2.cdn.mozilla.net/chains/onecrl.content-signature.mozilla.org-2021-03-22-17-58-02.chain: dial tcp [2600:9000:2163:f800:a:da5e:7900:93a1]:443: socket: address family not supported by protocol
2. failed to retrieve x5u: Get https://content-signature-2.cdn.mozilla.net/chains/normandy.content-signature.mozilla.org-2021-03-22-17-58-03.chain: dial tcp [2600:9000:2163:b400:a:da5e:7900:93a1]:443: socket: address family not supported by protocol
3. failed to retrieve x5u: Get https://content-signature-2.cdn.mozilla.net/chains/pinning-preload.content-signature.mozilla.org-2021-03-22-17-58-04.chain: dial tcp [2600:9000:2163:f800:a:da5e:7900:93a1]:443: socket: address family not supported by protocol
4. failed to retrieve x5u: Get https://content-signature-2.cdn.mozilla.net/chains/remote-settings.content-signature.mozilla.org-2021-03-22-17-58-04.chain: dial tcp [2600:9000:2163:ba00:a:da5e:7900:93a1]:443: socket: address family not supported by protocol: errorString
null
```

"failed to retrieve x5u" occurs in `GetX5U` in one location: https://github.com/mozilla-services/autograph/blob/master/signer/contentsignaturepki/upload.go#L84

which uses the stdlib http.Client https://github.com/mozilla-services/autograph/blob/master/signer/contentsignaturepki/upload.go#L76

@hwine suggested disabling ipv6 entirely, so we could use https://stackoverflow.com/questions/53055808/golang-force-net-http-client-to-use-ipv4-ipv6

Not going to take action now, but we'll see if we get more issues.

PR #698, opened 2021-05-18
monitor: build x5u client with timeouts

g-k:
fix: #674 #594

Set timeouts closer to the aws sdk go v2 s3 client https://github.com/mozilla-services/autograph/issues/674#issuecomment-828510317

We'll need to monitor changes to see if it fixes the issue.


g-k:
Merging since I believe the test failure is just related to gotest using a stale autograph ref and the unit test should be fixed on main.

Issue #620, opened 2021-03-03
test autograph starts with an expired XPI intermediate


g-k:
More generally when configured with an expired intermediate or root cert (for the Addon or content signature signers) autograph should:

1. start
2. log an error at the highest non-fatal severity 
3. continue running

Also, when requests are made to the a signer with an expired intermediates or root autograph should return a 4xx or 5xx http status code. An already running server should start returning the error when its intermediate or root expires.

g-k:
> Also, when requests are made to the a signer with an expired intermediates or root autograph should return a 4xx or 5xx http status code. An already running server should start returning the error when its intermediate or root expires.

Instead for XPIs in #722 autograph will continue signing with the expired chain and Fx clients will make the call to disable the addon or not.

g-k:
I don't think there's a use case for using an expired chain with content signature signers, since we resign the old data.

PR #722, opened 2021-07-16
signer: allow XPI signers to use an expired intermediate

g-k:
refs: #620 https://bugzilla.mozilla.org/show_bug.cgi?id=1549018

The new test dev PKI was generated using tools/genpki (`go run genpki.go
-no-hsm`) with backdated `NotBefore` and `NotAfter` and content signature
restrictions removed.

Changes:
* pin generated EE cert validity to between 2020-01-01 and 2030-01-01 (Fx ignores the EE cert validity but having it overlap with the rest of the chain lets us use the golang stdlib chain verification code)
* disable check that the XPI signer issuer cert / intermediate is valid at the current time when an XPI signer is initialized https://github.com/mozilla-services/autograph/pull/722/files#diff-7da932eaccc702303470a66346796e68c5cb5194b61720038306989dabcd8355L158
* update chain verification for PKCS7 and COSE XPI signatures to take a specific time to verify against
* add the `-vt` arg to set a verification time to `tools/autograph-client` XPI signing requests the client verifies after signing (#163 would split this apart). The client is used for integration tests.
* add names to `TestVerifyCOSESignaturesErrs` testcases
* add a new XPI signer to the autograph config and its key to load with softhsm (note that it uses a different root)
* decrease the minimum XPI cert size to accommodate the new signer

r? @ajvb @kkleemola


signer/xpi/cose.go
```diff
@@ -294,35 +294,37 @@ func verifyCOSESignatures(signedFile signer.SignedFile, truststore *x509.CertPoo
...
 	}
 
-	// check that we can verify EE certs with the provided intermediates
-	intermediates := x509.NewCertPool()
-	for _, intermediateCert := range intermediateCerts {
-		intermediates.AddCert(intermediateCert)
-	}
-	cndigest := sha256.Sum256([]byte(signOptions.ID))
-	dnsName := fmt.Sprintf("%x.%x.addons.mozilla.org", cndigest[:16], cndigest[16:])
+	// // check that we can verify EE certs with the provided intermediates
```
kkleemola, line 303: Why are things being commented out rather than deleted?

signer/xpi/cose_test.go
```diff
@@ -507,22 +507,23 @@ func TestVerifyCOSESignaturesErrs(t *testing.T) {
 	}
 
 	cases := []struct {
+		name    string
```
kkleemola, line 510: :tada:

signer/xpi/cose.go
```diff
@@ -5,7 +5,7 @@ import (
 	"crypto"
 	"crypto/ecdsa"
 	"crypto/rsa"
-	"crypto/sha256"
+	// "crypto/sha256"
```
ajvb, line 8: Why comment out rather than just remove?

signer/xpi/cose_test.go
```diff
@@ -507,22 +507,23 @@ func TestVerifyCOSESignaturesErrs(t *testing.T) {
 	}
```
ajvb, line 508: Should there be new tests around allowing expired certs?

signer/xpi/cose.go
```diff
@@ -347,8 +349,8 @@ func (s *XPISigner) issueCOSESignature(cn string, manifest []byte, algs []*cose.
 	if s.issuerCert == nil {
 		return nil, fmt.Errorf("xpi: cannot issue COSE Signature when XPISigner.issuerCert is nil")
 	}
-	if len(s.issuerCert.Raw) < 1100 {
-		return nil, fmt.Errorf("xpi: cannot issue COSE Signature DER encoded XPISigner.issuerCert should be at least 1100 bytes long")
+	if len(s.issuerCert.Raw) < 600 {
```
ajvb, line 352: Why this change? and is there a test for this?

go.mod
```diff
@@ -26,7 +26,7 @@ require (
 	go.mozilla.org/hawk v0.0.0-20190327210923-a483e4a7047e
 	go.mozilla.org/mar v0.0.0-20200124173325-c51ce05c9f3d
 	go.mozilla.org/mozlogrus v2.0.0+incompatible
-	go.mozilla.org/pkcs7 v0.0.0-20200128120323-432b2356ecb1
+	go.mozilla.org/pkcs7 v0.0.0-20210722210306-21e054180621
```
g-k, line 29: This is https://github.com/mozilla-services/pkcs7/pull/56

signer/xpi/cose.go
g-k: The value is a heuristic for "probably a real cert." It was an ask in the original COSE PR:

> you already verify that issuerCert != nil above, so this is more of a nit, but maybe check that len(s.issuerCert.Raw) > X where X is a minimal value for a cert?

https://github.com/mozilla-services/autograph/pull/76/files#r200813777



From the commit message history:

https://github.com/mozilla-services/autograph/pull/722/commits/8b0550205afcc2b3a58a38edeb849978044a88fb:

> ECDSA certs can be shorter
>
> The main check is that it's not empty and reasonable length the previous value based on a ballpark guess of dev certs at the time: 6365141

https://github.com/mozilla-services/autograph/commit/636514147dd70291bbec2a86519715bf983a20e7:

> I don't know the minimum offhand, but the dev ones are 2007 for the rsa one and 1132 for the ecdsa.
> 
> refs:
> https://github.com/mozilla-services/autograph/pull/76/files#r202278633
> https://github.com/mozilla-services/autograph/pull/76/files#r200813777

signer/xpi/cose.go
g-k: Yeah I forgot to finish fixing the verification code. It requires an upstream change https://github.com/mozilla-services/pkcs7/pull/56

signer/xpi/cose.go
g-k: Yeah I forgot to finish fixing the verification code. It requires an upstream change https://github.com/mozilla-services/pkcs7/pull/56

signer/xpi/cose.go
ajvb: Ok sounds good ty!

signer/xpi/cose_test.go
g-k: Yeah, I had to update the verification logic and they're buried a bit:

* the PKCS7 verification unit tests are in that PR
* `TestVerifyCOSESignaturesErrs` checks https://github.com/mozilla-services/autograph/pull/722/files#diff-c7f05b3e8291c266520101a71a41cd340754722852e691f2f05f6214be12a50cR481
  * verification at UTC now still returns an expired cert error https://github.com/mozilla-services/autograph/pull/722/files#diff-c7f05b3e8291c266520101a71a41cd340754722852e691f2f05f6214be12a50cR748
  * verification at `2019-01-01` returns a different error for invalid CN instead of an expired cert error https://github.com/mozilla-services/autograph/pull/722/files#diff-c7f05b3e8291c266520101a71a41cd340754722852e691f2f05f6214be12a50cL744-R796
* https://github.com/mozilla-services/autograph/pull/722/files#diff-59d350c2c572889a80240593671b12f226cbe506045eb575c8ddce67eb90e7bcL1170 removes a signer config with an invalid intermediate was previously expected to fail on signer init
* https://github.com/mozilla-services/autograph/pull/722/files#diff-59d350c2c572889a80240593671b12f226cbe506045eb575c8ddce67eb90e7bcR1105-R1107 adds a passing signer config testcase with the expired chain from the dev config and:
  * `TestSignFile` checks all valid configs initialize and sign a file it can verify https://github.com/mozilla-services/autograph/blob/ba655f880a8b076fbd346ef1b8215f90d8a237c7/signer/xpi/xpi_test.go#L20-L23 
* the client verifies everything it signs in the XPI integration tests https://github.com/mozilla-services/autograph/pull/722/files#diff-f71fc751da93e58e7dd840baf6efa9bd4de3c6e4126c0ce9b513776d80020130R296

ajvb: Awesome, ty!

ajvb (approved):

g-k:
Thanks AJ!

PR #740, opened 2021-08-26
signer: revert XPI EE validity to the interval UTC now and 87600h later

### LINKED ISSUE #739: Use signing time as notBefore in EE certificate of XPI signer
The add-ons blocklist relies on the notBefore of the end-entity certificate to make blocklist decisions:
- `notBefore` read at https://searchfox.org/mozilla-central/rev/28c8f793f9f8c80de24a179b4472098692a7e9a4/toolkit/mozapps/extensions/internal/XPIInstall.jsm#680
- Time used at https://searchfox.org/mozilla-central/rev/28c8f793f9f8c80de24a179b4472098692a7e9a4/toolkit/mozapps/extensions/Blocklist.jsm#1254-1259

In #722, the `notBefore` was changed to a fixed time in the past (2020). Consequently, newly signed add-ons had a chance of incorrectly considered to be blocked by Firefox. To resolve this, we should be using the current time as `notBefore` again, and add a unit test to prevent the time from regressing again in the future.

Note that the validity of `notBefore` is not considered in determining the validity of the certificate in Firefox (https://searchfox.org/mozilla-central/rev/28c8f793f9f8c80de24a179b4472098692a7e9a4/security/apps/AppSignatureVerification.cpp#637-659). The only purpose of `notBefore` in the EE cert is to determine the signing date for the purpose of the blocklist. In the future we may use a fixed date to verify the whole certificate chain (https://bugzilla.mozilla.org/show_bug.cgi?id=1549018).

g-k:
fix #739

Changes:
* set XPI EE validity to UTC Now and ~10 years later and add a unit test since apparently that was untested. Fx uses the `notBefore` for a blocklist input, so it isn't completely ignored and wasn't safe to hardcode it as mentioned in https://bugzilla.mozilla.org/show_bug.cgi?id=1549018#c21. Fixes #739.
* add a `-noverify` option to the golang client used in integration tests (refs #163)
* disable verification for the expired chain in integration tests (we test that autograph signs XPIs with an expired chain but don't verify the result). Golang's [Certificate.Verify](https://pkg.go.dev/crypto/x509#Certificate.Verify) does not have an option to ignore the EE cert and match Fx/NSS behavior. We'd need to hack together our own version, which is out of scope for this PR.

NB: With this change and #722, it's possible to create an EE with validity that doesn't overlap the rest of an expired chain:

```
EE                              |<----->|
Inter       |<----->|
Root  |<----------->|
```

It that occurs it might break monitoring.

### DIFF
--- a/signer/xpi/x509.go
+++ b/signer/xpi/x509.go
@@ -108,8 +108,8 @@ func (s *XPISigner) makeTemplate(cn string) *x509.Certificate {
 			Province:           []string{"CA"},
 			Locality:           []string{"Mountain View"},
 		},
-		NotBefore:          EENotBefore,
-		NotAfter:           EENotBefore.Add(87600 * time.Hour), // ten year
+		NotBefore:          time.Now().UTC(),
+		NotAfter:           time.Now().UTC().Add(87600 * time.Hour), // roughly ten years later
 		SignatureAlgorithm: s.issuerCert.SignatureAlgorithm,
 		KeyUsage:           x509.KeyUsageDigitalSignature,
 		ExtKeyUsage:        []x509.ExtKeyUsage{x509.ExtKeyUsageCodeSigning},
--- a/signer/xpi/x509_test.go
+++ b/signer/xpi/x509_test.go
@@ -7,6 +7,7 @@ import (
 	"crypto/sha256"
 	"fmt"
 	"testing"
+	"time"
 
 	"go.mozilla.org/cose"
 )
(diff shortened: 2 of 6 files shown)


signer/xpi/x509.go
```diff
@@ -108,8 +108,8 @@ func (s *XPISigner) makeTemplate(cn string) *x509.Certificate {
 			Province:           []string{"CA"},
 			Locality:           []string{"Mountain View"},
 		},
-		NotBefore:          EENotBefore,
-		NotAfter:           EENotBefore.Add(87600 * time.Hour), // ten year
+		NotBefore:          time.Now().UTC(),
```
ajvb, line 111: Is there any chance for a race condition here? Like should it be `now() - 5 seconds` or similar?

ajvb (approved):

signer/xpi/x509.go
g-k: Good question! Which events are you worried about racing?

I believe the expected order of events is:

1. addon signed / NotBefore generated
1. (optional) addon installed (possibly multiple installs)
1. blocklist updated. This assumes it knows about all addons signed at the time (ideally I think we'd use the PK7 SigningTime attr but that's a client side change)

I don't think we'd run into problems backdating by five seconds, but there's a chance that a concurrent backdated addon signing and a blocklist update could generate a blocklist that doesn't include the backdated NotBefore time.

Generally, backdating certs is frowned upon (AFAIK it's a violation of the web CA baseline requirements), but we do backdate content signatures to account for client clock skew so if there's an edge case we're missing it'd be good to consider that.

g-k:
Going to land this and we can revert and revisit later.

signer/xpi/x509.go
ajvb: Ok totally! With that order, it makes sense that it is extremely unlikely for there to be a race condition. As well, this was more of a general curiosity rather than me being specifically concerned.

g-k: Got it! I appreciate having to think it through.

As we chatted about we'll have to be careful with unit tests that sign and verify quickly since those could race depending on the resolution the verifier uses.

PR #749, opened 2021-09-08
Fix 738 endpoint for keyid lookup

### LINKED ISSUE #737: autograph should make it possible to look-up available keyids for a signer
As I understand it, keyids are not secrets. As an autograph user, it would be useful to be able to look up which ones are available for a particular signer, and perhaps be able to look up available signers as well. @escapewindow had a couple of ideas, including:
1) Simply document them in this repo
2) Have an available endpoint that returns them (eg: `/signers` to get signers and/or `/signers/foo` to get keyids for the `foo` signer)

Feel free to close this as undesirable if we the current obfuscation is wanted :)

### LINKED ISSUE #738: log signed XPI or addon version number
Extract and log the top-level `.version` field from the `/manifest.json` path in the XPI.

Recommended addons will also include it in the recommendation file.

I think we can implement this by:

1. add `manifest.json` as a path to the consts in xpi.go
2. read the version key from `manifest.json` by either
          A. `readFileFromZIP` then `json.Unmarshal` to parse the manifest as JSON then extract the `version` key (worse perf, lower complexity)
          B. hook into `makeJARManifest`, `repackJARWithMetafiles`, or another function that iterates through the zip then wait until the manifest is read, parse it, and extract and return the `version` key as another return value (better perf, more complex)
3. log the version by either
          A. logging the version key directly from the xpi signer then if we need to correlate after the fact with the signing request IDs
          B. adding a new SignFile interface and logging the extra data as fields with [the "signing operation succeeded" lines](https://github.com/mozilla-services/autograph/blob/1b01419c138ff4e3ee47d0819fdc44ccac285266/handlers.go#L228-L239) (might help with #214 too)

g-k:
fix #737 using the first proposed approach

Changes:
* define a user/auth ID format (like the signer id format); validate users at boot; and refactor `autographer.addAuthorizations` tests in https://github.com/mozilla-services/autograph/pull/749/commits/97ce5b27d13bd98a195cefb99abdd04aefa63af8
* implement `/auths/:auth_id/keyids` more or less as proposed in #737; add support to the golang client and integration tests (handler tests don't exercise the mux router vars)

### DIFF
--- a/authorize.go
+++ b/authorize.go
@@ -10,6 +10,7 @@ import (
 	"crypto/sha256"
 	"fmt"
 	"net/http"
+	"regexp"
 	"time"
 
 	log "github.com/sirupsen/logrus"
@@ -23,6 +24,11 @@ type authorization struct {
 	Key     string
 	Signers []string
 }
+
+// authIDFormat is a regex for the format Authorization IDs must follow
+const authIDFormat = `^[a-zA-Z0-9-_]{1,255}$`
+
+var authIDFormatRegexp = regexp.MustCompile(authIDFormat)
 
 func abs(d time.Duration) time.Duration {
 	if d < 0 {
--- a/authorize_test.go
+++ b/authorize_test.go
@@ -187,28 +187,104 @@ func TestDefaultSignerNotFound(t *testing.T) {
 	}
 }
 
-// Two authorizations sharing the same ID should fail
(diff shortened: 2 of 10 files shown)


ajvb (approved):

g-k:
Thanks AJ!

PR #768, opened 2021-09-20
add missing indirect deps for CI then tidy

g-k:
Fix nightly CI failures.

Ignore coveralls reporting failures while they're down for maintenance.


g-k (commented):
👀

PR #770, opened 2021-09-22
remove DSA key parsing

### LINKED ISSUE #667: remove DSA key support
refs: #637

go stdlib DSA support is deprecated.

This is a breaking change that would allow us to remove some code in https://github.com/mozilla-services/autograph/blob/main/signer/signer.go

g-k:
fixes #667

breaking change

finishes up changes from #666 that removed the first apk signer. 

was added in #162


g-k (commented):
👀

PR #771, opened 2021-09-22
Fix 738 log addon version

### LINKED ISSUE #738: log signed XPI or addon version number

g-k:
fix #738 

Changes:
* rename test vars `PASSINGTESTCASES` to `validSignerConfigs` and `FAILINGTESTCASES` to `invalidSignerConfigs` to avoid confusion with other vars named `testcases`
* use go embed directives (1.16+) https://pkg.go.dev/embed to remove bytestrings from source files and expose the files for other tools to inspect (e.g. `unzip -l`) 
    * add `.zip` to the `omni.ja` files so zip tools recognize them as zips 
* test sign file against the ublock and `omni.ja` fixtures
* run `TestSign{Data,File}` test cases in parallel saved ~6s on local testing of the package 
* log the addon `manifest.json` top-level `version` field and `.browser_specific_settings.gecko.id` field using
   * `extractGeckoIDAndVersionFromWebextManifest`, which assumes the version field is a string per [the MDN docs](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/version). This was the laziest option to fix #738 since it doesn't modify the repack func return values or the sign file interface.


signer/xpi/jar.go
```diff
@@ -623,3 +633,19 @@ func verifyAndCountManifest(signedXPI []byte, manifestPath string) (int, int, er
...
 }
+
+func extractGeckoIDAndVersionFromWebextManifest(manifestBytes []byte) (geckoID string, version string, err error) {
+	var manifest struct {
+		Version                 string `json:"version"`
+		BrowserSpecificSettings struct {
+			Gecko struct {
+				ID string `json:"id"`
+			} `json:"gecko"`
+		} `json:"browser_specific_settings"`
```
willdurand, line 644: I think `applications` is an alias for `browser_specific_settings`. In addition, this prop is optional.

Is it OK to not log the add-on ID all the time?

willdurand: Some more information:

1. the `version` should be mandatory
2. the GUID (same value as the Gecko ID) is somehow inserted in the signature of a signed XPI so I suppose there is a way to _always_ get the GUI somewhere?

g-k: > I think `applications` is an alias for `browser_specific_settings`. In addition, this prop is optional.

TIL! I didn't see that key on MDN or in [the webext spec](https://browserext.github.io/browserext/#list-of-keys), but these pages suggest Fx <48 required it:

* https://stackoverflow.com/questions/37127330/unrecognized-manifest-key-applications-warning-for-google-chrome
* https://support.mozilla.org/en-US/questions/1217302

so it probably predates standardization and we might have to sign addons with that key.

> Is it OK to not log the add-on ID all the time?

Yes, I have [a few test cases for partial matches](https://github.com/mozilla-services/autograph/pull/771/files#diff-10e1d7b6567d37db6b1a0eafed0dba17a3581c73a1822a914e860ad38d34f828R373-R407) and XPIs signed in other modes (privileged, hotfix, etc.) probably won't include the an ID or might use the other key.

8b4f49197c732b0eee8cfc0b1d03e134506ae2cc adds test cases for:

- [x] a manifest with `applications` set to an empty obj
- [x] a manifest with `applications.gecko` set to an empty obj
- [x] a manifest with `applications.gecko.id` set to an empty str
- [x] a manifest with `applications.gecko.id` set to an addon ID
- [x] a manifest with a version and both IDs set (in which case we'll prefer the browser settings ID)

Once we have some logs, I'll try to confirm we can correlate them with signed addons. If not, we might need pass the data out to [the "signing operation succeeded" lines](https://github.com/mozilla-services/autograph/blob/1b01419c138ff4e3ee47d0819fdc44ccac285266/handlers.go#L228-L239).

g-k: > Some more information:
> 
>     1. the `version` should be mandatory

Hmm, `extractAddonIDAndVersionFromWebextManifest` could return an error and we could log it as a warning instead of info level.

If other XPI types don't include a manifest, we won't log warnings. If they do, we might log some extra warnings but that'd be fine.

>     2. the GUID (same value as the Gecko ID) is somehow inserted in the signature of a signed XPI so I suppose there is a way to _always_ get the GUI somewhere?

Does addons-server generate that and send it as [the Options.ID](https://github.com/mozilla-services/autograph/blob/main/signer/xpi/xpi.go#L485)?

willdurand: > Does addons-server generate that and send it as [Options.ID](https://github.com/mozilla-services/autograph/blob/main/signer/xpi/xpi.go#L485)?

I think so, yes: https://github.com/mozilla/addons-server/blob/80f526e84084a776f9f379301e11250202f9a8ab/src/olympia/lib/crypto/signing.py#L77

willdurand (approved):
r+wc

g-k:
Thanks Will!

Follow up items from comments and original issue:

- [ ] return parsed ID and version to `SignFile` in xpi
  - [ ] log parsed ID and version with [the "signing operation succeeded" lines](https://github.com/mozilla-services/autograph/blob/1b01419c138ff4e3ee47d0819fdc44ccac285266/handlers.go#L228-L239) instead of from the repacking code
  - [ ] when signing in mode `ModeAddOn` and `ModeAddOnWithRecommendation` check the lints:
    - [ ] warn for a missing `manifest.json` file 
    - [ ] warn for multiple IDs: `applications.gecko.id` and `browser_specific_settings.gecko.id`
    - [ ] warn for manifest ID not matching request `Options.ID`
    - [ ] warn for a missing `version`

Warnings return errors if we want to do more linting in autograph and other web extension signers pass for legitimate uses.

PR #782, opened 2021-10-04
fix 743 multi file and 744 debsign support

### LINKED ISSUE #743: support signing and returning multiple files in a signing request
Blocks #742 #744

Possible implementation:

1. add a `POST /sign/files` endpoint with request format accepting an array of base64 encoded files (like [/sign/file](https://github.com/mozilla-services/autograph/blob/main/docs/endpoints.rst#21request)):

```
POST /sign/files
Host: autograph.example.net
Content-type: application/json
Authorization: Hawk id="alice", mac="756lSgQEYLoc6V0Uv2wS8pRg/h+4WFUVKWQynCFvY8Y=", ts="1524487134", nonce="MrpGL35q", hash="9m3WhtGQDuHermi5fDYBGJlOqNeK5B3nk0lKreZ+YSw=", ext="933126753"

[
    {
      "inputs": [
           "UEsDBBQACAAIAAAAAAAAAAAAAAAAAAAAAAATAAAAQW5kcm9pZE1hbmlmZXN0LnhtbKSYS2ybx7XHf0PqbVmW4...BwAACigAAAAA",
           "UEsDBBQACAAIAAAAAAAAAAAAAAAAAAAAAAATAAAAQW5kcm9pZE1hbmlmZXN0LnhtbKSYS2ybx7XHf0PqbVmW4...BwAACigAAAAA",
           ...
     ],
      "keyid":"testapp-android-v3",
      "options":null
    }
]
```

2. add a new interface with a new method `.SignFiles()` that accepts a slice of input files and returns a slice of signed files (instead of single values):
 
```golang
type MultipleFileSigner {
  SignFiles(files [][]byte, options interface{}) ([]SignedFile, error)
  GetDefaultOptions() interface{}
} 
```

3. return responses with an array of dicts of signed filename to base64 file contents: 

```json
[
  {
    "ref": "1p21kj11od4no13o1xepn22mkc",
    "type": "debsign",
    "mode": "???",
    "signer_id": "debsign1",
    "public_key": "MHYwEAYHKoZIzj0CAQYFK4EEACIDYgAE7oM/ewOhz6qtHyQhqJvT3SiefGPWqGwEUAZGVkuSIwvteVKrd8jnAjHYyCaYpIg9Vo10WnhXvm96L3KAbOE6Cyu3fMtKhZZIMf+Qqes9+66ae/NTeIWlDiGrjNeD+ClM",
    "signed_files": [
           {"filename": "foo_signed.sig", "base64_content":  "Niffk674SNKzQaq23z2sv7xkU_IEgrPc8_tEFGw0bYXlNJDpAPe7hEaipyg-wY10_XUzkoRphtYVIAa70Hw22EkWfSGAdzosEYyxsDai52PG088KqasP_nd_byiiqIAz"},
           {"filename": "bar_signed.sig", "base64_content":  "Niffk674SNKzQaq23z2sv7xkU_IEgrPc8_tEFGw0bYXlNJDpAPe7hEaipyg-wY10_XUzkoRphtYVIAa70Hw22EkWfSGAdzosEYyxsDai52PG088KqasP_nd_byiiqIAz"},
           ...
     ]
  }
]
```

We'd update logs to include the input and output hashes of each file, and monitoring would ignore the multifile responses (like the apk2 signer; pending #622).

Alternative options:

* require clients to pack the files beforehand (e.g. as a tarball) then unpack them on the server side and repack them
* accept files in a different encoding #276 (this could also make #214 easier too)

see also: https://mozilla-hub.atlassian.net/browse/ITSEC-105 about the possibility of switching to one signing request per http request (simplifies including http response codes in the application layer for #318 too)

### LINKED ISSUE #744: debsign support
Support signing debian packages as discussed in email thread with subject "VPN Linux signing approach".

Signer Behavior:

1. accept the dsc, buildinfo, and changes files
2. acquire the global gpg2 lock
2. Like the apk2 and gpg2 signers, shell out to [debsign](https://manpages.debian.org/stretch/devscripts/debsign.1.en.html) to run `debsign -k <my key name> --re-sign *.dsc *.buildinfo *.changes`
3. release the gpg2 lock
4. return the signed .dsc, .buildinfo, or .changes files

Assumptions:

* there aren't significant differences between the ubuntu and debian buster versions of debsign
* the `*.dsc *.buildinfo *.changes` files aren't huge (the manpage suggests they should at least be smaller than the full package)

Current estimate is ~1 week of uninterrupted time.

Alternatively, we could try to decompose `dpkg-buildpackage` to get the sign hook input or try to extract hashes, but as Ben pointed out that'd require more custom tooling and incur a higher maintenance overhead to keep up with upstream changes.

g-k:
# WIP

fix #743 multiple file signing
fix #744 debsign support


docs/endpoints.md
```diff
@@ -82,6 +82,100 @@ Each signature response contains the following fields:
...
+
+The number of files must be between 1 and 32 inclusive. All characters
+in a file name must be an alphanumeric character, dash, underscore or
+dot. The file name must start with an alphanumeric character and dots
+cannot occur next to each other in the file name.
+
+example:
+
+``` bash
+POST /sign/file
```
ajvb, line 124: Should this be `/sign/files`?

docs/endpoints.md
```diff
@@ -82,6 +82,100 @@ Each signature response contains the following fields:
...
+The response format is the same as `/sign/data` except
+instead of the `signature` field autograph returns the
+field:
+
+-   `signed_files` an array of dicts of file name and base64 encoded
+    signed file content as described in the request section. Each
+    signer uses a different format, so refer to their documentation
+    for more information.
+
+TODO: document restrictions
```
ajvb, line 175: What "restrictions" does this refer to?

handlers.go
```diff
@@ -47,18 +57,20 @@ func hashSHA256AsHex(toHash []byte) string {
...
-		"options":     sigreq.Options,
-		"mode":        sigresp.Mode,
-		"ref":         sigresp.Ref,
-		"type":        sigresp.Type,
-		"signer_id":   sigresp.SignerID,
-		"input_hash":  inputHash,
-		"output_hash": nil,
-		"user_id":     userid,
-		"t":           int32(time.Since(starttime) / time.Millisecond), //  request processing time in ms
+		"rid":           rid,
```
ajvb, line 62: nit: Seeing this set of fields in a number of locations, I wonder if this should be moved to a struct that can output a `log.Fields`.

handlers_test.go
```diff
@@ -50,42 +50,91 @@ func TestBadRequest(t *testing.T) {
...
+		{`/sign/file`, `POST`, `[{"input": ""}]`},
+		{`/sign/files`, `POST`, `[{"input": ""}]`},
 		// input not in base64
 		{`/sign/data`, `POST`, `[{"input": "......."}]`},
 		{`/sign/hash`, `POST`, `[{"input": "......."}]`},
-		// asking for a xpi signature using a hash will fail
+		{`/sign/file`, `POST`, `[{"input": "......."}]`},
+		{`/sign/files`, `POST`, `[{"input": "......."}]`},
+
+		// missing files
```
ajvb, line 71: Tests look great! Is there one for the `unsigned filename must match ^[-_\.a-zA-Z0-9]{1,256}$` check that I'm missing?

docs/endpoints.md
g-k: d'oh yeah, nice catch! Fixed in 449aa5eb

docs/endpoints.md
g-k: Ah, this was a stray stash. Removed in 84ed7180

The filename restrictions are mentioned in [the /sign/files request body section](https://github.com/mozilla-services/autograph/pull/782/files#diff-b52944c3b61aec977284cbe459a436d41c0ee9193582df47cfe8ecad1c29066dR116-R120).

handlers.go
g-k: Yeah, it's not exactly DRY. The [logrus readme](https://github.com/sirupsen/logrus#example) includes this example:

```go
  // A common pattern is to re-use fields between logging statements by re-using
  // the logrus.Entry returned from WithFields()
  contextLogger := log.WithFields(log.Fields{
    "common": "this is a common field",
    "other": "I also should be logged always",
  })
```

so that might be an option too. Unless you think it should be addressed in this PR I'll take a TODO and file an issue to clean that up.

signer/gpg2/gpg2_test.go
```diff
@@ -231,7 +315,153 @@ func TestSignData(t *testing.T) {
 				}
 				t.Logf("GnuPG PGP signature verification output:\n%s\n", out)
 			})
+		})
+	}
+}
+
+func TestGPG2Signer_SignFiles(t *testing.T) {
```
ajvb, line 322: Is there verification of `os.RemoveAll()` running successfully within `SignFiles`?

ajvb (approved):
LGTM! 🎉 

Just have two comments related to the use of `os.RemoveAll()` in the new gpg signing code.

handlers_test.go
ajvb: Not sure if it's needed, but would it be worth including a test for a filename with say `@` or some other non-valid character in it?

g-k: Good call! I added a few more in d99e8754.

Checked the OWASP cheat sheet too: https://cheatsheetseries.owasp.org/cheatsheets/File_Upload_Cheat_Sheet.html#filename-sanitization

signer/gpg2/gpg2.go
```diff
@@ -269,3 +354,86 @@ type Options struct {
...
+		return
+	}
+
+	// create a tmp dir outside the signer GPG home
+	inputsTmpDir, err := ioutil.TempDir("", fmt.Sprintf("autograph_%s_%s_%s_sign_files", s.Type, s.KeyID, s.Mode))
+	if err != nil {
+		err = fmt.Errorf("gpg2: error creating tempdir for debsign: %w", err)
+		return
+	}
+	defer os.RemoveAll(inputsTmpDir)
```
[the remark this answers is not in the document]
g-k: Hmm yeah good call! I thought `defer` would propagate errors somehow.

I think it's OK if we don't clean up the input files, but we definitely want to know if we're failing to delete them.

So:

142025daae888f1f43251e2b43e51bc1436c6769 / 4683927 will warn but not return an error when input files or SignData content file cleanup fails
b8962b7fb579695d79f3ba4ce88c341a8d5ce165 will error if pub and private key cleanup fails, since leaving them on disk is an error. Other errors (e.g. key import failure) will take precedence.

g-k:
> (re-approving after new commits dismissed last approval)

TY! I still need to squash this PR too.

PR #796, opened 2021-11-08
Avoid confusion with DEV server of Remote Settings

leplatrem:
While adding support for our new DEV environment on the remotesettings-devtools addon, we realized this comment could be confusing. Because our DEV server uses the STAGE instance of Autograph :)


g-k (approved):
r+ thanks!

PR #821, opened 2022-03-23
Update AMO code owner

willdurand:
GitHub says this file has errors, though. Also there is #806..


willdurand:
mmm looks like @diox should be added to this repo.

Issue #849, opened 2023-03-21
Allow `~` characters in filenames for debsign


ahal:
In the debian packaging world, `~` is a special character that can be used in version strings to denote a pre-release or test package which should be prioritized below all other versions (or something like that). The important bit is that it has meaning.

Unfortunately, autograph validates all file names we sign against a regex and disallows `~`. It would make sense to support files with `~` in the name for debsign where this has meaning.

It looks like this check happens here:
https://github.com/mozilla-services/autograph/blob/0edcdc80246921740e864a56d17bf2939beb3be0/signer/signer.go#L222

The good news is that it should be simple to add `~` to the regex. The bad news is that isn't debsign specific. However, the previous check is already validating that the filename can't start with `~`. So from a security standpoint, I think this would cover malicious attempts to access the home dir? In that case maybe allowing it for all signers isn't a big deal?

ahal:
@hwine any thoughts? Can we just add `~` to the regex here?

hwine:
N.B. prioritization & scheduling will be handled in bug-1823721

oskirby:
For some context, the Debian policy guide has a formal description of the package version strings [here](https://www.debian.org/doc/debian-policy/ch-controlfields.html#version), and IIRC the `upstream_version` and `debian_version` components of that version are able to appear in the filenames of artifacts to be signed.

The TLDR is that a you can wind up with package filenames containing `[0-9A-Za-z.+-~]` (alphanumerics, full stop, plus, hyphen, and tilde).

hwine:
PRs accepted 🙂 (with tests, of course)

Please make sure it applies cleanly to tag 4.1.6 (not main), and integration tests pass. (Note, you'll need `docker-compose` locally for `make integration-tests` to pass. We're not ready to update the tool chain to modern docker due to CircleCI issues.)

PR #850, opened 2023-03-21
Allow Debian version special chars in unsigned filenames

oskirby:
To full support Debian package signing, there are a few extra special characters that we may need to accept as valid filenames: `.` `-` `~` and `+` (full-stop, hyphen, tilde and plus). This patch should extend the `signer.isValidUnsignedFilename()` function to permit these characters, and adds some unit tests to make sure it works as expected.

Since the tilde character is sometimes used as a suffix by some text editors to denote backup files, I felt it was prudent to add an extra check to ensure that filenames also end with an alphanumeric character. I wouldn't expect this to cause issues, but I figured I should mention it as well in case anyone knows of a usecase that this might break.

I have tested this locally with `make integration-test` and it appears to pass.

Fixes: #849

### DIFF
--- a/signer/signer.go
+++ b/signer/signer.go
@@ -221,8 +221,10 @@ func isValidUnsignedFilename(filename string) error {
 	if !regexp.MustCompile(`[a-zA-Z0-9]$`).MatchString(filename) {
 		return fmt.Errorf("unsigned filename must end with an alphanumeric character")
 	}
+	// support debian version conventions as documented at:
+	//    https://www.debian.org/doc/debian-policy/ch-controlfields.html#version
 	if !regexp.MustCompile(`^[-_\.+~a-zA-Z0-9]{1,256}$`).MatchString(filename) {
-		return fmt.Errorf(`unsigned filename must match ^[-_\.a-zA-Z0-9]{1,256}$`)
+		return fmt.Errorf(`unsigned filename must match ^[-_\.+~a-zA-Z0-9]{1,256}$`)
 	}
 	if regexp.MustCompile(`\.\.`).MatchString(filename) {
 		return fmt.Errorf("unsigned filename must not include ..")
--- a/signer/signer.go
+++ b/signer/signer.go
@@ -215,10 +215,13 @@ type NamedSignedFile namedFile
 
 // isValidUnsignedFilename
 func isValidUnsignedFilename(filename string) error {
-	if !regexp.MustCompile(`^[a-zA-z0-9]`).MatchString(filename) {
+	if !regexp.MustCompile(`^[a-zA-Z0-9]`).MatchString(filename) {
 		return fmt.Errorf("unsigned filename must start with an alphanumeric character")
 	}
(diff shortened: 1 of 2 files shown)


signer/signer.go
```diff
@@ -215,10 +215,13 @@ type NamedSignedFile namedFile
 
 // isValidUnsignedFilename
 func isValidUnsignedFilename(filename string) error {
-	if !regexp.MustCompile(`^[a-zA-z0-9]`).MatchString(filename) {
+	if !regexp.MustCompile(`^[a-zA-Z0-9]`).MatchString(filename) {
```
hwine, line 218: Nice Catch!

signer/signer.go
```diff
@@ -215,10 +215,13 @@ type NamedSignedFile namedFile
...
 		return fmt.Errorf("unsigned filename must start with an alphanumeric character")
 	}
-	if !regexp.MustCompile(`^[-_\.a-zA-Z0-9]{1,256}$`).MatchString(filename) {
+	if !regexp.MustCompile(`[a-zA-Z0-9]$`).MatchString(filename) {
```
hwine, line 221: @ahal - this will affect all gpg2 zip file contents signing. RelEng is, I believe, the only user of that signing mode. Can you sign off that the change won't adversely impact any other signing, please?

signer/signer.go
```diff
@@ -215,10 +215,13 @@ type NamedSignedFile namedFile
...
+		return fmt.Errorf("unsigned filename must end with an alphanumeric character")
+	}
+	if !regexp.MustCompile(`^[-_\.+~a-zA-Z0-9]{1,256}$`).MatchString(filename) {
```
hwine, line 224:
```suggestion
    # support debian versionconventions as documented at:
    #    https://www.debian.org/doc/debian-policy/ch-controlfields.html#version
	if !regexp.MustCompile(`^[-_\.+~a-zA-Z0-9]{1,256}$`).MatchString(filename) {
```

hwine (changes requested):
One minor change suggested.

N.B. I'll have other folks reviewing -- I'll be holding off on my approval until they comment.

signer/signer.go
ahal: Assuming you mean this only impacts signing formats using the `files` endpoint, this should be safe. Afaict, VPN is the only consumer using that endpoint with gpg signing. Further, that format only signs files with specific file extensions (which all end in alphanumeric characters).

Even if this does impact other things other than `files`, the chance of it causing problems seems exceedingly small. I can't think of any files we'd want to sign that would end in non alphanumeric characters.

hwine (approved):

quiiver (approved):
lgtm. This is outside the scope of this PR but generally in Go you use `regexp.MustCompile` for global variable initialization. `regexp.MustCompile` will cause the whole process to panic vs returning an `error` with `regexp.Compile`

hwine:
Ran `unit-test` locally, build passes. Pushing @d72b95ef0e8e78259aa62b659ae01451692ece8d to deploy to stage. This PR remains the source-of-truth.

PR #852, opened 2023-06-16
Bug 1838680 - Do not force --min-sdk-version anymore because of outda…

JohanLorenzo:
…ted values

Reasons are documented in [bug 1838680 comment 1](https://bugzilla.mozilla.org/show_bug.cgi?id=1838680#c1).

r? @hwine cc @rocketsroger


JohanLorenzo:
Lint failure seems unrelated to this patch. I'm sorry, I'm no go expert to take it on 😬

hwine:
> Lint failure seems unrelated to this patch. I'm sorry, I'm no go expert to take it on 😬

okay to fail `lint-vet-fmt` and `unit-test` on CI, but `ci/circleci: build-integrationtest-verify` must pass.

Locally, both the integration & unit tests must pass.

JohanLorenzo:
Thanks for highlighting that, @hwine. I didn't realize `build-integrationtest-verify` ran some kind of linting. I removed the unused dependency in `apk2.go` and the job is back to green.

hwine:
@JohanLorenzo there are failures in the unit-tests - I just updated the `Makefile` to allow you to safely run the unit tests locally via "`make test-in-docker`". After running that, I get:
```bash
❯ grep --line-number FAIL unit-test.log
504:autograph-unit-test                    | --- FAIL: TestSignaturePass (11.57s)
515:autograph-unit-test                    | FAIL
517:autograph-unit-test                    | FAIL       github.com/mozilla-services/autograph   19.391s
2367:autograph-unit-test                    | FAIL
```
(remember to run "`make build`" after any code change before testing)

JohanLorenzo:
I see, it seems the 4 `FAIL`s are all related to this error: 

```
autograph-unit-test                    | === CONT  TestSignaturePass
autograph-unit-test                    |     handlers_racing_test.go:146: failed with 500: signing request 1dnw9avx58cm012n5x730rmch5 failed with error: apk2: failed to sign
autograph-unit-test                    |         Exception in thread "main" com.android.apksig.apk.MinSdkVersionException: Failed to determine APK's minimum supported platform version. Use --min-sdk-version to override
```

It looks like the APK it's trying to sign doesn't contain `minSdkVersion` in its manifest. It looks like a bug in the test.

handlers_racing_test.go
```diff
@@ -82,7 +82,7 @@ func TestSignaturePass(t *testing.T) {
...
...AAAH4BYKTfM1MAAABXAAAAFAAYAAAAAAABAAAApIGyDwAATUVUQS1JTkYvTUFOSUZFU1QuTUZVVAUAA1BBhxR1eAsAAQToAwAABOgDAABQSwECHgMKAAAAAAB1f3VMAAAAAAAAAAAAAAAADAAYAAAAAAAAABAA7UFTEAAAdGhpcmRfcGFydHkvVVQFAAOuubJadXgLAAEE6AMAAAToAwAAUEsBAh4DCgAAAAAAdX91TAAAAAAAAAAAAAAAABUAGAAAAAAAAAAQAO1BmRAAAHRoaXJkX3BhcnR5L2phdmFfc3JjL1VUBQADrrmyWnV4CwABBOgDAAAE6AMAAFBLAQIeAwoAAAAAAHV/dUwAAAAAAAAAAAAAAAAhABgAAAAAAAAAEADtQegQAAB0aGlyZF9wYXJ0eS9qYXZhX3NyYy9lcnJvcl9wcm9uZS9VVAUAA665slp1eAsAAQToAwAABOgDAABQSwECHgMKAAAAAAB1f3VMAAAAAAAAAAAAAAAAKQAYAAAAAAAAABAA7UFDEQAAdGhpcmRfcGFydHkvamF2YV9zcmMvZXJyb3JfcHJvbmUvcHJvamVjdC9VVAUAA665slp1eAsAAQToAwAABOgDAABQSwECHgMKAAAAAAB1f3VMAAAAAAAAAAAAAAAANQAYAAAAAAAAABAA7UGmEQAAdGhpcmRfcGFydHkvamF2YV9zcmMvZXJyb3JfcHJvbmUvcHJvamVjdC9hbm5vdGF0aW9ucy9VVAUAA665slp1eAsAAQToAwAABOgDAABQSwECHgMUAAAACAAAAH4BCmgnsZIAAADVAAAATAAYAAAAAAABAAAApIEVEgAAdGhpcmRfcGFydHkvamF2YV9zcmMvZXJyb3JfcHJvbmUvcHJvamVjdC9hbm5vdGF0aW9ucy9Hb29nbGVfaW50ZXJuYWwuZ3d0LnhtbFVUBQADUEGHFHV4CwABBOgDAAAE6AMAAFBLAQIeAxQAAAAIAAAAfgHAqt49uQAAAHcBAABIABgAAAAAAAEAAACkgS0TAAB0aGlyZF9wYXJ0eS9qYXZhX3NyYy9lcnJvcl9wcm9uZS9wcm9qZWN0L2Fubm90YXRpb25zL0Fubm90YXRpb25zLmd3dC54bWxVVAUAA1BBhxR1eAsAAQToAwAABOgDAABQSwUGAAAAABoAGgCiCgAAaBQAAAAA",
```
hwine, line 85: Rather than replacing the test, I'd prefer to retain the existing test, and reflect the behavior change by explicitly turning the test into an xfail with note.

handlers_racing_test.go
```diff
@@ -82,7 +82,7 @@ func TestSignaturePass(t *testing.T) {
...
+					Input: "UEsDBAAAAAAIACEIIQJ3XNdl0gIAAMwHAAATAAAAQW5kcm9pZE1hbmlmZXN0LnhtbJ2VTU9TQRSGz/QCrZYiFMq3RhMTjQlFkYVxYYJEokllodGFKyoFIbTQ9F5QEmNYu3bpwoV7/QX+AhcuWPhr1GfOndtOr238uM17Z+adc945c+bcaSA5+ZoVMTIvrwdELkjneef1R8AsKIMK2AYn4CP4Ar6DH2DZiNwBa2AbHIG34D04BdMZkSWwBt6Ab+BqwBg8A3XwAZyCQalLVZ7LFq3IgOwzajASYt6SV9KUA2lJRL8GV2BuF5vHjPbkKWxLQpgDOJG8HHUxq7xrqpWeWW+vMop2lZkXjKK+ulWiqzN6KXfpb2JzSGR2JtReEmUoj3jbnYxhdcAaTVTq6PVWnvkLq2QXncwEcoMzuk4vQ++mntxDOZYVjaSO3ya2kfNfYn5Io47a6+ba4yMQ4SuShdtnpRY2u5rt2RRT1txHGkvEyNcsE8GKPCCzIhf/4BdHt0XObdaOYSr4PsF3Ve7LPXIY57zXbmzsaX8baZzHstZMVTNq81mmd9xHyeblyn/52b1WdU8rv2XxmuzQi/C7LYv8Qix2UGzgEWrOujMTr7+oa+3RtrANddx9FoWuHC7wbdqKibRSbE4a6rELb/0jPc+mq9WqVjffpe7G7mRbM9egmg9Vp9an5v7FZ937cg9hbRwLwFY094jJybLYO8iYAIyC4Ywxc2AeNMEJ+BQY0xo0JgSC1bRWuchPnlHaKTuG/+zx9pmhP8Ev6+6yKb1N4vkcn7lx3KDWlo1Rgrzjznh2lxx3Fgw7u2E3b+/GkuNKjptL+dr+vMcVXLwVPZFOvOddvBkv3sDzKzluKKUfuJyktZI18h4/3mONbEevaNupeD9dnNXacP8HidY5pxV4WqJ1Geej4Dib50nHTSbn1UP/MlzR0592+sljbW7Rjns2I87GeDEYvWvj9ca8/KT9Er2Sx0/00Ss6vaKnl/ZL+PQeEj6du4RPn0+QqvGklk2f2v8FUEsBAgAAAAAAAAgAIQghAndc12XSAgAAzAcAABMAAAAAAAAAAAAAAAAAAAAAAEFuZHJvaWRNYW5pZmVzdC54bWxQSwUGAAAAAAEAAQBBAAAAAwMAAAAA",
 					KeyID: "testapp-android",
```
hwine, line 86: Would it make sense to add the notes about creating this test case (in a PR comment) to an actual doc? (Or a comment in this file?)  In general, `autograph` devs won't be APK experts, so that domain knowledge could be very useful. :smile:

JohanLorenzo: Sounds good. Done!

handlers_racing_test.go
JohanLorenzo: Agreed. A new test case was added instead of replacing the existing one.

JohanLorenzo:
> * use the manifest values, if present, to update values actually used (i.e. manifest has higher precedence than command line option)

Good call 👍 I used that approach

hwine:
@quiiver can you give a quick glance, please -- especially for any memory leaks, as this is in a long running process. Particularly [this new function](https://github.com/mozilla-services/autograph/pull/852/files#diff-4ee034ddcad5186c45b707fe39a7a68096d244ccedd6d705d143243cbdc39145R191-R204) which is munging string arrays.

signer/apk2/apk2.go
```diff
@@ -176,6 +189,19 @@ func (s *APK2Signer) SignFile(file []byte, options interface{}) (signer.SignedFi
 	return signer.SignedFile(signedApk), nil
 }
 
+// go 1.16 doesn't support generics.
+// TODO: Replace string by T in function signature once we use go 1.18+
+func InsertIntoSliceAtIndex(destination []string, element string, index int) []string {
```
quiiver, line 194: Not a blocker but no need for this to be a package public function.

quiiver (approved):
lgtm

signer/apk2/apk2.go
JohanLorenzo: TIL the case of the first letter tells whether a function is publicly exposed or not! (I'm a super-n00b at Go 😁). Thanks for pointing that out. Fixed!

hwine: Really? yuck -- I thought we learned not to do that with Fortran! :(

hwine (approved):
passes full unit test locally (ci version has some XFAILS)

PR #861, opened 2023-10-18
Slim down Autograph container

grahamalama:
The `autograph` container is large, which makes pulling it in CI pipelines pretty time consuming. I looked for anything we could do to make it smaller to speed up pulls, and I came up with these "quick wins":

- build binaries in separate stage and copy to final stage
- install deb packages with `--no-install-recommends`
- remove `apt-get` lists as a cleanup step
- add `.git` directory to `.dockerignore`

These changes make for a 44% reduction of the size of the compressed container

```
❯ ls -lah ../ | grep tar
-rw-r--r--@  1 gbeckley  staff   556M Oct 18 18:44 after.tar.gz
-rw-r--r--@  1 gbeckley  staff   998M Oct 18 18:21 before.tar.gz
```


grahamalama:
This passed the integration test, though I noticed some of these in [the output on CircleCI](https://app.circleci.com/pipelines/github/mozilla-services/autograph/1837/workflows/7b59e9fb-0615-4e11-a38e-0d82e784b5fe/jobs/10053/parallel-runs/0/steps/0-105):
> fatal: not a git repository (or any of the parent directories): .git

I'm guessing that comes from here https://github.com/mozilla-services/autograph/blob/7bc7ab2ce78848cd6a69dc4753990a6d89e02eac/tools/autograph-client/build_test_xpis.sh#L14

Is the SHA itself important here? Maybe there's another value we can provide, or maybe it's not important in a CI setting? If we do need it, maybe we can provide it as a build arg or environment variable.

hwine:
**N.B.:** [Remote settings](https://github.com/mozilla/remote-settings/blob/9ed1829862bc6c4ccfcb7361a391e3845f849add/docs/tutorial-local-server.rst#with-multi-signoff-advanced) does some gnarly stuff for testing. This is not a "guaranteed interface", but it'd be nice not to break it.

jmhodges:
@oskirby updated the dockerfile [in similar ways just now](https://github.com/mozilla-services/autograph/pull/897). Is there still stuff left on table for us to do?

jmhodges:
(That comment was directed @grahamalama. Edited to be more clear!)

grahamalama:
@jmhodges did you mean to link to [this PR](https://github.com/mozilla-services/autograph/pull/898)? It seems to get more into the nitty gritty of the Dockerfile than https://github.com/mozilla-services/autograph/pull/897.

I'm not seeing any PR that does quite the same thing as this one, but maybe these changes could be folded into one of the ones from @oskirby. https://github.com/mozilla-services/autograph/pull/898 seems to make the image (just a bit) larger, but I don't see a reason why we couldn't apply similar strategies (multistage build, cleaning after we use `apt`, leverage `.dockerignore`) to a different base image.

oskirby:
I have updated my PR #898 to include these size reduction techniques, and I have added a workaround for the git error in the XPI signer test.

grahamalama:
Excellent, @oskirby! Thanks. I'll close this PR then in favor of #898

PR #864, opened 2023-11-16
Sanitize Configs & export to csv

hwine:
Generates csv files for authorizations and signers. Recommended usage would be:
  { sops -d app.yaml ; echo "---" ; sops -d edge.yaml ; } | autograph-config-sanitizer


mlarsonweber (approved):
lgtm

PR #867, opened 2023-12-11
Add android store support scripts

### LINKED ISSUE #866: Scripts to general Android APK keys and package for play store
To generate and upload keys (signing and upload) for new Android apps requires a specific set of steps in a specific order, with the "baton" of who can perform the next step switching back and forth between release management, developers, and autograph maintainers. Combined with the difficulty of setting up an appropriate environment in which to perform the cryptographic operations, there is too much room for typos. If the actors are many timezones apart, this can produce unacceptable delays.

The approach envisioned here is to:
- Script as many of the steps as practical.
- Keep the scripts in a public repository (this one) for easy of review by Android Store SMEs.
- Ensure the scripts are not tampered with by making them part of the Autograph release process, and storing on the production container image.
- Updating all processes to reference doing the work via tools in a container from the current production image.

hwine:
Work to support #866 being:
- visible for review by SME
- accessible via official containers

Expected to fix #866

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,9 @@ tools/sign-addon/venv/*
 
 vendor/go.mozilla.org/cose/samples/*
 version.json
+
+# Python dev stuff
+**/.venv*
+**/venv*
+**/__pycache__
+**/*.egg-info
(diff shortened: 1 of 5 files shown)


mlarsonweber (approved):
lgtm

PR #871, opened 2023-12-29
Back out unverified code from production branch

hwine:
See https://github.com/mozilla-services/autograph/issues/870 for details. The git commands to create this commit are given below. (The README.md changes were manual.) All tests pass locally per the README.md instructions.

Git Commands:
> git fetch
> git checkout main
> git checkout --branch hwine/issue870
> git restore --source=7.1.11 -- handlers_racing_test.go signer/apk2/apk2.go
> git add
> git commit

PR #872, opened 2023-12-29
Changes for bug 1838680

hwine:
The changes in 7bc7ab2ce78848cd6a69dc4753990a6d89e02eac were backed out in order to ensure the production branch is deployable from the head in #871. (These changes had been landed for testing from #852, and deployed for a while in stage, but never validated. Hence we were tagging the prior revision for production deploys.)

The other files in the original commit were improvements to the documentation and development processes. Since these did not impact the production code, they were left in.


hwine:
@JohanLorenzo FYI

PR #11, opened 2016-01-20
Conditional docker builds in travis

jvehent:
@mostlygeek : r?

PR #873, opened 2024-01-24
vendor the Go dependencies

jmhodges:
This was done by running `go mod vendor` in the top-level directory.

There's no difference between vendoring with a modern Go and 1.16, but
that can verified (as I've done locally) with commands like:

```
go install golang.org/dl/go1.16.5@latest # fetches tooling to install Go 1.16.5
go1.16.5 download
go1.16.5 mod vendor
```

Notably, while this vendors all of the PKCS #11 Go libraries and similar, this doesn't vendor any of the C dependencies (e.g. the CloudHSM library which is pinned and installed outside of Go's tooling)


jmhodges:
For folks less familiar with Go tooling, the `vendor` directory has special meaning.See the [vendoring docs](https://go.dev/ref/mod#vendoring). Specifically,

> If the vendor directory is present in the main module’s root directory, it will be used automatically if the go [version](https://go.dev/ref/mod#go-mod-file-go) in the main module’s go.mod [file](https://go.dev/ref/mod#glos-go-mod-file) is 1.14 or higher.

willdurand:
> For folks less familiar with Go tooling, the `vendor` directory has special meaning. See the [vendoring docs](https://go.dev/ref/mod#vendoring). Specifically,
> 
> > If the vendor directory is present in the main module’s root directory, it will be used automatically if the go [version](https://go.dev/ref/mod#go-mod-file-go) in the main module’s go.mod [file](https://go.dev/ref/mod#glos-go-mod-file) is 1.14 or higher.

FWIW, this is what we're doing in other Mozilla go-based tools, e.g. https://github.com/mozilla-services/stubattribution

hwine (approved):
all tests pass locally

PR #882, opened 2024-03-11
Disable dependabot non-security updates

### LINKED ISSUE #881: Disable dependabot
With commit c890e14, we've frozen everything we care about, until such time as we resume development. Thus, dependabot updates are not actionable, do disable to reduce noise.

All the [settings](https://github.com/mozilla-services/autograph/settings/security_analysis) for dependabot have been disabled. However the `.github/dependabot.yml` file still exists, so issues are still being opened.

We should remove that file to reduce non-actionable tickets.

hwine:
We're in maintenence mode with frozen dependencies. We can reenable once there is active development.

Fixes #881


bbangert (approved):
Yep, agreed.

PR #884, opened 2024-03-21
adding contentsignaturepki leaf cert CN overrides

jmhodges:
This adds the ability to override the common name
set in the Subject of the leaf certificates
generated by the contentsignaturepki type singer
configurations.

This is to allow for new certificates created with
the same key material but different singer
configurations to be validated by Firefox and
other clients as they were previously. This is an
important requirement for the add-on and content
signing root key succession plan as we're asking
the distribution services to use new signers (new
`keyids`) for the new keys, while keeping the old
common names in the end-user clients.

Fixes [bug 1886886](https://bugzilla.mozilla.org/show_bug.cgi?id=1886886)

### DIFF
--- a/signer/contentsignaturepki/contentsignature_test.go
+++ b/signer/contentsignaturepki/contentsignature_test.go
@@ -76,15 +76,21 @@ func TestSign(t *testing.T) {
 		if err != nil {
 			t.Fatalf("testcase %d failed to get X5U %q: %v", i, s.X5U, err)
 		}
-		key := certs[0].PublicKey.(*ecdsa.PublicKey)
+		leaf := certs[0]
+		key := leaf.PublicKey.(*ecdsa.PublicKey)
 		if !sig.(*verifier.ContentSignature).VerifyData([]byte(input), key) {
 			t.Fatalf("testcase %d failed to verify signature", i)
+		}
+
+		if leaf.Subject.CommonName != testcase.expectedCommonName {
+			t.Errorf("testcase %d expected common name %#v, got %#v", i, testcase.expectedCommonName, leaf.Subject.CommonName)
 		}
 	}
 }
 
 var PASSINGTESTCASES = []struct {
-	cfg signer.Configuration
+	cfg                signer.Configuration
+	expectedCommonName string
 }{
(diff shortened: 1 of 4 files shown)


signer/contentsignaturepki/x509.go
```diff
@@ -146,3 +146,11 @@ func (s *ContentSigner) makeChain() (chain string, name string, err error) {
 	name = fmt.Sprintf("%s-%s.chain", cert.Subject.CommonName, cert.NotAfter.Format("2006-01-02-15-04-05"))
 	return
 }
+
+func (s *ContentSigner) domainForLeafCert() string {
+	subdomain := s.ID
+	if s.subdomainOverride != "" {
```
hwine, line 152: What's the value if the config does not contain a `SubdomainOverride` value? Is it `""` or `nil`?

signer/contentsignaturepki/contentsignature_test.go
```diff
@@ -76,15 +76,21 @@ func TestSign(t *testing.T) {
...
+		if leaf.Subject.CommonName != testcase.expectedCommonName {
+			t.Errorf("testcase %d expected common name %#v, got %#v", i, testcase.expectedCommonName, leaf.Subject.CommonName)
+		}
 	}
 }
 
 var PASSINGTESTCASES = []struct {
-	cfg signer.Configuration
+	cfg                signer.Configuration
+	expectedCommonName string
```
hwine, line 93: If I'm reading this correctly, we modified the existing test case for the new code. I'd prefer that we add a new test case to show that the "old config format" (without the `SubdomainOverride` field) still works.

signer/contentsignaturepki/x509.go
jmhodges: It's a `""`. That's the Go zero value for strings.

signer/contentsignaturepki/contentsignature_test.go
jmhodges: Good call. Done

hwine (approved):
lgtm - PR builds clean & tests pass locally

PR #885, opened 2024-03-22
make the subdomain_override test exemplary

jmhodges:
The docs for SubdomainOverride
(subdomain_override) correctly state that if you
add it to a signer config, you should also change
the X5U and ChainUploadLocation configs to avoid
colliding with the other singer. Let's do that in
the tests even though this is not part of the
codebase that is in a position to check for that
kind of behavior.


signer/signer.go
```diff
@@ -117,7 +117,10 @@ type Configuration struct {
 
 	// SubdomainOverride is to override the subdomain of the leaf certificates
 	// created. This is mostly for contentsignaturepki. If this isn't set, the
-	// `KeyID` is used as the subdomain, instead.
+	// `KeyID` is used as the subdomain, instead. When setting this value to
+	// match another extant signer id, also be sure to set the X5U and
+	// ChainUploadLocations of this signer configuration to avoid uploading
```
hwine, line 122:
```suggestion
	// ChainUploadLocations of this signer configuration to a unique value to avoid uploading
```

hwine (approved):
r+ with nit

signer/signer.go
jmhodges: Ah, I accepted this, but then merged over and then double checked and: we can't say `unique` here because we already have multiple signers with the same ChainUploadLocations deployed. Is there another language you'd like that would account for how this is already used?

hwine: Oh, ick - yah, I got nothing to improve the original then.

hwine (approved):
👍🏻

PR #893, opened 2024-04-24
Update Docker image to Debian bookworm and get CI tests to pass

oskirby:
This is trying to see what happens if we update the Docker image on which autograph is built from `golang:1.16.10-buster` to something a bit more recent. In this case I chose `debian:bookworm` since it's the current stable release of Debian.

So far, this PR can be divided up into three epochs:
1. Commits up to 6bc6d6a67cd843365359afeec5fcaeb9b12510ee are about updating the autograph docker image.
2. Commits up to 97c74c0d9bdb1ce718705ebcf120d0e540b900e5 are about getting the CI jobs passing.
3. Commits after that are focusing on upgrading the Circle CI environment to newer images.

Because one of my objectives of this work was to get the CI tests passing again, I have uncovered a bunch of broken tests and linting errors which have accumulated in the absence of working tests. So this has kind of turned into a large pull request. If this is too much to digest in one PR, let me know and I can make an effort to break it up into smaller pieces.


jmhodges:
Oh, this repo is a squash and merge one! Could you split this up into smaller PRs? I'd rather us not have to revert all of this work just because one commit had a problem. Some of these also feel totally separable like the docker-compose changes.

jmhodges:
Also, this patch rules

oskirby:
I am switching this back to a draft, and I will split it up into multiple PR's that are a little more digestable.

oskirby:
I am going to close this as all of the good bits now live on in PR #898

Issue #853, opened 2023-06-16
Fix CI 


hwine:
Currently, only `ci/circleci: build-integrationtest-verify` passes on CI.

The unit tests will pass locally. So should be made to pass on CI.

Per @quiiver this should be straight forward without bumping golang version.

Will - any guidance here appreciated. More folks are getting into the code 😉

Note: if we're fighting any CircleCI issues here, moving to GitHub Actions is a viable path.

hwine:
Temp workaround done in #856 so we can simplify the "does a PR pass" story.

The changes do not "fix" anything, rather they treat the current failures a XFAIL at the most granular level available in the CircleCI config file. When the underlying tests are truly fixed, the CI run will fail.

PR #894, opened 2024-04-26
Fix failures in CI testing

oskirby:
In order to start work on some larger upgrades and refactoring, with the confidence that we aren't breaking things, we should start by getting the unit tests and code quality checks in CI to start passing again. This PR attempts to split just that part of the work out from #893 and get it into a smaller form, suitable for review and merging.

To break down the contents of this PR:
1. Fix the installation of test tooling (`staticcheck`, `golint` and `goveralls`) so that they don't interfere with vendored modules, and pin them to a known-good version while we're at it.
2. Remove a conflicting copy of `github.com/mozilla-services/autograph/verifier/contentsignature` from the vendor dir.
3. Fix the resulting linter errors and unit tests breakage that were previously uncaught.
4. Fix a command substitution bug in the checking of lint results.
5. Expect the tests to pass in CI.

Fixes: #853


hwine (commented):
I :love: the idea, but I'm clueless about some of the changes. Deferring to Jeff.

vendor/github.com/mozilla-services/autograph/verifier
jmhodges, line 1: It looks like you added the `replace` directive to `go.mod` and this symlink both. I think you just need the `replace`? (I tried reproducing behavior by deleting this symlink and running `go mod vendor` and then some `go build` in a few spots. But you mentioned some static analysis work, so I could be missing the right check)

jmhodges (changes requested):

vendor/github.com/mozilla-services/autograph/verifier
jmhodges: Oh, it was `staticcheck` not working right? Strange.

jmhodges (approved):
LGTM!

Thank you so much!

I do wonder if, in a future patch, we wouldn't be better off getting rid of the nested go.mods, but I see [we have clients using the module](https://github.com/search?type=code&q=github.com%2Fmozilla-services%2Fautograph%2Fverifier%2Fcontentsignature) elsewhere.

vendor/github.com/mozilla-services/autograph/verifier
oskirby: The thing I found is that removing the symlink and running `go mod vendor` causes go to rebuild the vendor directory, and it just re-creates a copy of `verifier/contentsignature` in `vendor/github.com/mozilla-services/autograph/verifier/contentsignature`.

So, all we have really done is update the vendored copy without removing it. It produces successful builds, but lays the foundation for more errors to creep in again by having two copies of the code.

The symlink isn't a great solution either though as it will just get blown away the next time someone tries a `go mod vendor` again.

oskirby: I think in this case staticcheck was working as expected, and the project was being built with the copy out of the vendor directory. The problem is, that version was old, and had begun to diverge from the copy in `verifier/contentsignature`

jmhodges:
Sorry for the delay on this merge. I wanted to get you collab rights, but we got distracted by yesterday's events

PR #847, opened 2023-03-08
Update: Docker compose is a sub command now

hwine:
New versions of Docker have 'compose' as a subcommand. The existing scripts used the older `docker-compose` executable, which has been removed.

`make integration-test` runs fine with the change.

Fixes #846

PR #897, opened 2024-05-02
Update: Use cimg/go:1.16 image and docker compose subcommand

oskirby:
Rebasing PR #847 now that we have passing tests.

The `circleci/golang:1.16-buster` docker image has been deprecated, and has not received updates in over two years now and we have developer friction now due to the use of the deprecated `docker-compose` command. We should update these to their modern equivalents.


jmhodges (approved):
Would mildly prefer the comments to have been left alone (if we interpret them as referencing the tool's name) or have backticks around it (if we interpret them as referring to the literal command), but that isn't a blocker and I don't hold that opinion strongly

jmhodges (approved):

jmhodges (approved):

PR #898, opened 2024-05-02
Update: Use Debian bookworm as base docker image

oskirby:
Debian [10/buster](https://www.debian.org/releases/buster/) has been discontinued, and stopped receiving security updates as of June 2022, and LTS support is going to expire later this year in June 2024. We should make an effort to upgrade to a newer version of Debian. With that work in mind, I have selected Debian [12/bookworm](https://www.debian.org/releases/bookworm/) given that it is the current stable release of Debian. 

This PR attempts to upgrade the base docker image from `golang:1.16-buster` to `debian:bookworm`. This also necessitates an upgrade to golang 1.19, and brings with it a handful of test fixes due to changes in the standard go libraries.

This is a subset of PR #893 with only the bits necessary to upgrade the base docker image.


jmhodges:
This one has to wait for us to get off CentOS 7 on our VMs. We'll need to plan out what that looks like next week

Dockerfile
```diff
@@ -1,20 +1,17 @@
-FROM golang:1.16.10-buster
+FROM debian:bookworm
 EXPOSE 8000
 
-ENV GODEBUG=x509ignoreCN=0
```
jmhodges, line 4: Hm, this should still be required to work correctly. Nervy that it does. Might be making leaf certificates incorrectly? Did I miss something thought?

oskirby: To explain this one a little bit, the `x509ignoreCN` environment variable tells golang to weaken its certificate subject name checks and accepts certificates which only contain a `CN` field when trying to match a DNS name. Support for this variable was removed in Go 1.17 (see notes for crypto/x509 [here](https://go.dev/doc/go1.17#minor_library_changes)). And a valid certificate, as far as go is concerned, must now use the X509v3 `subjectAltName` extension to provide a DNS name.

This caused a problem for the integration tests, because the test certificate for the DB only provides `CN=db` and this is no longer sufficient for go 1.17, which is the reason for commit 43fb9573f79a900744e0786b1af00acb52854d35 in this PR to update the test certs.

jmhodges: Ah, I was worried that variable applied at certificate creation, not just verification, time and we'd be generate certs without a CommonName and only a SAN but that's not the case (from looking up the older docs for [VerifyHostname](https://pkg.go.dev/crypto/x509@go1.16.15#Certificate.VerifyHostname) and writing [a test impl](https://gist.github.com/jmhodges/42c8b1cb14451844014b00dc16a45a31)). 

I think I saw the cose_test.go and verifier_test.go changes thought they might be related. And I'm extra paranoid about this because this month we had [to correct a bug](https://github.com/mozilla-services/autograph/pull/884) around us making the CN).

Any idea why the error messages for cose_test.go and verifier_test.go got worse? Not a strong request but maybe you know already

middleware.go
```diff
@@ -48,7 +48,7 @@ func setRequestStartTime() Middleware {
 	}
 }
 
-//  Run the request through all middlewares
```
jmhodges, line 51: You know, we could do these gofmt changes out of this PR since we still have wait on the centos7 change. I can post a quick PR.

jmhodges: https://github.com/mozilla-services/autograph/pull/900

oskirby: I'll rebase my PR after that one lands to remove it - IIRC `lint-vet-fmt` job will fail without the linting changes. Not that it really matters, but I like keeping the CI green.

Dockerfile
oskirby: The changes in the tests are due to some refactoring of the Go certificate parsing routines in `crypto/x509` that changed the error messages that library generates when we feed it some invalid ASN.1. From the Go 1.17 [changelog](https://go.dev/doc/go1.17#minor_library_changes):
> [ParseCertificate](https://go.dev/pkg/crypto/x509/#ParseCertificate) has been rewritten, and now consumes ~70% fewer resources. The observable behavior when processing WebPKI certificates has not otherwise changed, except for error messages.

Since our unit tests are using the error messages as the pass/fail conditions, they need to be updated to accommodate new versions of Go when those messages change.

jmhodges: Hell yeah, thank you

jmhodges:
This is patch is shippable once we do the CentOS7 upgrades! Thank you so much!

middleware.go
oskirby: Rebase has been completed!

jmhodges (approved):

jmhodges (changes requested):
Oh actually missed something important: this doesn't bump the circleci docker image versions from 1.16, so how is the 1.19 change happening? I don't see that written down in here but it's mentioned in the PR summary.

jmhodges:
> Oh actually missed something important: this doesn't bump the circleci docker image versions from 1.16, so how is the 1.19 change happening? I don't see that written down in here but it's mentioned in the PR summary.

Oh, in the `apt-get install -y golang` bit? Hrm, I'm a lil nervy about floating on that. Is there anyway to do this work while pinning the Go version explicitly?

jmhodges:
I should explain my contexting: I'm asking about that because I'm _actually_ thinking about how we upgrade to even more modern Go from that one. It could be your answer is "I have an idea for that but it will happen after this. The idea is $X".

oskirby:
You are correct that this doesn't touch the `cimg/go:xxx` docker images, but the version of those images doesn't matter as they are just used as a launching point to invoke the autograph container where the unit tests are being run (which uses the golang version we get from Debian bookworm). I guess that could be tidied up a bit to make the go version a little more consistent.

oskirby:
As for how to pin versions, we are doing it the "Debian way" right now, which means that Debian has chosen 1.19 as the version that goes with the Bookworm release. They may release security fixes from time to time, but will only ever do so by increasing the patch version so that it will always be Go 1.19.X.

If we want to move forward or backwards to a different major/minor version then we can do so by installing a different package from the `bookworm-backports` repository.

For example:
  - `apt-get install golang` will always give us some 1.19 version of golang, possibly with security fixes.
  - `apt-get install golang-1.21` will give us golang 1.21 but we have to enable a backport repository to get it.
 
Fun fact, `golang` is actually just installs some symlinks to wrap  the `golang-1.19` package. So we could also pin things by specifying `golang-1.19` regardless of whether we're using backports or not.

oskirby:
Okay! A new feature has been added to the `Dockerfile` and it now accepts an optional `GO_VERSION` argument which can be used to change the version of Golang used. For example, adding the following to the docker compose file builds the `autograph-app` container with Golang version 1.22:
```
services:
  app:
    build:
      args:
        GO_VERSION: 1.22
```

oskirby:
FWIW, this is still not full upgraded to go 1.19 as the top level `go.mod` targets 1.16

```
module github.com/mozilla-services/autograph

go 1.16
```

But I think that is an upgrade for another day as we will have to update dependencies and deal with libraries that were deprecated between 1.16 and 1.19.

jmhodges:
Thank you for this work! I was going to totally just take this as future stuff to worry in response to your first response here.

jmhodges (approved):

PR #900, opened 2024-05-03
run a modern gofmt on the codebase

jmhodges:
First seen in PR #898, and maybe worthy of
breaking it out to a separate PR.

Run with:

go1.22.2 fmt ./...
go1.22.2 fmt ./tools/... # to catch the submodules

### DIFF
--- a/middleware.go
+++ b/middleware.go
@@ -48,7 +48,7 @@ func setRequestStartTime() Middleware {
 	}
 }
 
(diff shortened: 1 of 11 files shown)


jmhodges:
This doesn't include the `-s` (simplify) because I can't remember if the struct name elision from slices works on 1.16. Could, though.

oskirby (approved):
Looks good to me!

PR #901, opened 2024-05-03
Update deploy job to use cimg/deploy:2024.03.1

oskirby:
In my recent PR #897 to update the CircleCI environment, we needed to change the docker cache path because `/go/` is no longer writeable as a non-root user. Unfortunately, this seems to have caused the deploy job to fail due to differing project directories between the `docker:18.06.3-ce` image and the ones used by CircleCI.

To fix this, I think we can update the deployment docker image to a CircleCI one and it should all be good.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -120,7 +120,7 @@ jobs:
 
   deploy:
     docker:
-      - image: docker:18.06.3-ce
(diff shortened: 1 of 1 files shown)


jmhodges (approved):
Hell yeah

PR #907, opened 2024-06-05
set Go version in go.mod to 1.19

jmhodges:
This matches the version we're installing in the Dockerfile. Many
security updates have occurred in our deps and dependabot can't update
them but also the GO tooling and language has improved substantially in
the past few years.

Go 1.19 tooling expects the X.Y version format while more modern
versions will expect X.Y.Z. Switch to latter later.

This PR was made by running:

```
$  go mod edit -go=1.19
$  go mod tidy
$  go mod vendor
```


jmhodges:
When we do our next Go version bump, I recommend going straight to the latest version of Go. Go 1.20 has [an RSA performance drop](https://github.com/golang/go/issues/59442) that I know has been fixed in Go 1.22, at least (maybe 1.21 but I can't recall).

oskirby (approved):
Looks good to me, I am happy to accept this change.

However, as a note for future us we now have a handful of deprecation warnings in the staticcheck tests about `io/ioutil` which would be nice to clean up at some point.

jmhodges:
> Looks good to me, I am happy to accept this change.
> 
> However, as a note for future us we now have a handful of deprecation warnings in the staticcheck tests about `io/ioutil` which would be nice to clean up at some point.

[Draft PR](https://github.com/mozilla-services/autograph/pull/911) for that work that I'll wait to see if it passes CI before merging this one.

jmhodges:
> > Looks good to me, I am happy to accept this change.
> > However, as a note for future us we now have a handful of deprecation warnings in the staticcheck tests about `io/ioutil` which would be nice to clean up at some point.
> 
> [Draft PR](https://github.com/mozilla-services/autograph/pull/911) for that work that I'll wait to see if it passes CI before merging this one.

Oh, I don't see those deprecation warnings in the CI output of this Go 1.19 PR. Odd. But it kind of looks like our lints are bugged? There's error exit codes being raised that CircleCI seems to be configured to ignore?

PR #909, opened 2024-06-12
Collect package installation into common base image

oskirby:
In the Debian bookworm update, we split the Docker container build into two phases, the first builds the autograph binaries and the second phase installs config and gets it ready for deployment. Unfortunately, this meant that we needed to install packages and build tooling into both stages, which doubled the amount of time it took for package installation.

This attempts to clean that up by creating a common base image with the Debian package setup, and re-using it for both the build and deployment phases. This saves us about 60-90s of build time and it makes local development via `docker compose run unit-test` significantly faster by caching the package setup.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
 ARG GO_VERSION=1.19
 
 #------------------------------------------------------------------------------
-# Build Stage
+# Base Debian Image
 #------------------------------------------------------------------------------
-FROM debian:bookworm as builder
+FROM debian:bookworm as base
 ARG GO_VERSION
 
 ENV DEBIAN_FRONTEND='noninteractive' \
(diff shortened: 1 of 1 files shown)


jmhodges (approved):
Nice. I was going to worry about us shipping an image with gcc installed into it (for image size reasons, mostly) but we already were! This is a strictly better solution.

oskirby:
On the topic of image size, this currently sits somewhere around 580MB for the Autograph image, and off the top of my head the storage is utilized something like:
 - 200MB: Generic Debian
 - 60MB: Autograph Sources
 - 40MB: Autograph Binaries
 - 40MB Golang Build Tooling
 - 200MB: Dependencies just for `apksigner` - amazingly it depends on the full graphical JRE runtime. It's *a lot* of stuff.
 - 40MB: Signer and build package dependencies.

It's still a lot better than what it used to be, but we have some low-hanging fruit (Autograph Soruces and Golang build tooling) that don't really need to be there in the shipped image. And we could save a ton of storage space if we found a different signer for APKs that didn't pull in the entire JRE runtime.

PR #910, opened 2024-06-14
Fix monitor contentsignature checks

oskirby:
It turns out that the monitor integration test has been failing for a while now because the `contentsignature` signer type is not required to provide an X5U for signature verification, but the monitor check is reusing the verification code from `contentsignaturepki` which does require an X5U and the public key comes from slightly different places in each case.

If you look a little closer at the monitor integration tests, you will find errors like this:
```
Verifying content signature from signer "appkey1"
Response from signer "appkey1" does not pass: content signature response is missing an X5U to fetch
Response was: {Ref:8pa9gsxl0nxk3vmthnxqgjx0r Type:contentsignature Mode:p384ecdsa SignerID:appkey1 PublicKey:MHYwEAYHKoZIzj0CAQYFK4EEACIDYgAE7oM/ewOhz6qtHyQhqJvT3SiefGPWqGwEUAZGVkuSIwvteVKrd8jnAjHYyCaYpIg9Vo10WnhXvm96L3KAbOE6Cyu3fMtKhZZIMf+Qqes9+66ae/NTeIWlDiGrjNeD+ClM Signature:nCTEFViv7rIioqm4fmD38-iyx38fL20ryNlZQGan4T2k83cCOG7TI9tufKzjpEGR7r0DNBZUsr8KrURccQA9qobqjOPBkbnzjztKqsYAZchfLrbLCjlcigW6LxFsIXzM SignedFile: SignedFiles:[] X5U: SignerOpts:<nil>}
Verifying content signature from signer "appkey2"
Response from signer "appkey2" does not pass: content signature response is missing an X5U to fetch
Response was: {Ref:27u8rhwzevdl020v7mr9xyd6ij Type:contentsignature Mode:p256ecdsa SignerID:appkey2 PublicKey:MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEMdzAsqkWQiP8Fo89qTleJcuEjBtp2c6z16sC7BAS5KXvUGghURYq3utZw8En6Ik/4Om8c7EW/+EO+EkHShhgdA== Signature:4ys9GWOTXzoQjWNKObIY3B4PubtYSaVeCj1uW0ZMx8x45kQedtCHEX8N2vyN8DU4hggwYWcA7Gj1rFq7eXaU9Q SignedFile: SignedFiles:[] X5U: SignerOpts:<nil>}
```

To fix this, we need a to add a proper verification algorithm for `contentsignature` and use it in monitor when there's no X5U present. To do this, we add two new methods:
 - `contentsignature.VerifyResponse`: Accepts a `formats.SignatureResponse` from the `/sign/data` endpoint and performs a check of the ECDSA signature, and
 - `contentsignature.VerifyResponseHash`: Does the same thing, but checks the result of the `/sign/hash` endpoint.

Some food for thought: it is awkward that the `contentsignature` algorithm calculates its hashes differently in both the `/sign/hash` and `/sign/data` cases, which means that this is the only signature type where we need the context of the signing call to perform signature validation.

It would also be a good idea to check that when an X5U is present, the public key of the leaf/EE certificate matches the PublicKey returned in the signature.

### DIFF
--- a/handlers_racing_test.go
+++ b/handlers_racing_test.go
@@ -168,7 +168,11 @@ func TestSignaturePass(t *testing.T) {
 					t.Fatalf("in test case %d on endpoint %q, error '%v' in response %d;\nrequest was: %+v\nresponse was: %+v failed to decode input",
 						i, testcase.endpoint, decodeErr, j, testcase.signaturerequests[j], response)
 				}
-				err = contentsignature.VerifyResponseHash(rawInput, response)
+				if req.URL.RequestURI() == "/sign/hash" {
+					err = contentsignature.VerifyResponseHash(rawInput, response)
+				} else {
+					err = contentsignature.VerifyResponse(rawInput, response)
+				}
 			case xpi.Type:
 				err = verifyXPISignature(testcase.signaturerequests[j].Input, response.Signature)
 			case apk2.Type:
--- a/handlers_racing_test.go
+++ b/handlers_racing_test.go
@@ -163,10 +163,10 @@ func TestSignaturePass(t *testing.T) {
 		for j, response := range responses {
 			switch response.Type {
 			case contentsignature.Type:
-				rawInput, err := base64.StdEncoding.DecodeString(testcase.signaturerequests[j].Input)
-				if err != nil {
+				rawInput, decodeErr := base64.StdEncoding.DecodeString(testcase.signaturerequests[j].Input)
+				if decodeErr != nil {
 					t.Fatalf("in test case %d on endpoint %q, error '%v' in response %d;\nrequest was: %+v\nresponse was: %+v failed to decode input",
-						i, testcase.endpoint, err, j, testcase.signaturerequests[j], response)
+						i, testcase.endpoint, decodeErr, j, testcase.signaturerequests[j], response)
 				}
 				err = contentsignature.VerifyResponseHash(rawInput, response)
 			case xpi.Type:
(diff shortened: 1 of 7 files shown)


handlers_racing_test.go
```diff
@@ -163,10 +163,16 @@ func TestSignaturePass(t *testing.T) {
...
-				err = verifyContentSignatureResponse(
-					testcase.signaturerequests[j].Input,
-					response,
-					testcase.endpoint)
+				rawInput, decodeErr := base64.StdEncoding.DecodeString(testcase.signaturerequests[j].Input)
+				if decodeErr != nil {
+					t.Fatalf("in test case %d on endpoint %q, error '%v' in response %d;\nrequest was: %+v\nresponse was: %+v failed to decode input",
+						i, testcase.endpoint, decodeErr, j, testcase.signaturerequests[j], response)
+				}
+				if req.URL.RequestURI() == "/sign/hash" {
```
jmhodges, line 171: ugh, okay, we def got to delete this type after this

jmhodges (approved):

handlers_racing_test.go
oskirby: If you are referring to the fact that the signature is generated differently in `SignHash()` and `SignData()` - it looks like the same situation exists for `contentsignaturepki`, but it's just that we have no tests covering that scenario.

PR #911, opened 2024-06-20
migrate ioutil calls to io,os calls

jmhodges:
The Go 1.19 and nearby versions marked these calls as deprecated. Clean
them up with their documented replacements.

Ran these commands to fix it up:

```
$ gofmt -w -r 'ioutil.TempFile -> os.CreateTemp' $(find . -type f -name '*.go' | grep -v "/vendor/")
$ gofmt -w -r 'ioutil.ReadAll -> io.ReadAll' $(find . -type f -name '*.go' | grep -v "/vendor/")
$ gofmt -w -r 'ioutil.WriteFile -> os.WriteFile' $(find . -type f -name '*.go' | grep -v "/vendor/")
$ gofmt -w -r 'ioutil.TempDir -> os.MkdirTemp' $(find . -type f -name '*.go' | grep -v "/vendor/")
```

Then removing the `ioutil` import from the files that no longer use it.

Oh, also ha to revert some gofmt changes in `verifier/contentsignature`
that I'll ship separately.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -9,7 +9,6 @@ package main
 import (
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 
 	log "github.com/sirupsen/logrus"
@@ -30,7 +29,7 @@ func httpError(w http.ResponseWriter, r *http.Request, errorCode int, errorMessa
 	// request body is read before writing a response.
 	// https://github.com/golang/go/issues/15789
 	if r.Body != nil {
-		io.Copy(ioutil.Discard, r.Body)
+		io.Copy(io.Discard, r.Body)
 		r.Body.Close()
 	}
 	http.Error(w, msg, errorCode)
--- a/handlers.go
+++ b/handlers.go
@@ -12,7 +12,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"net/http"
 	"os"
 	"path"
@@ -90,7 +90,7 @@ func (a *autographer) handleSignature(w http.ResponseWriter, r *http.Request) {
 		httpError(w, r, http.StatusUnauthorized, "authorization verification failed: %v", err)
 		return
 	}
-	body, err := ioutil.ReadAll(r.Body)
+	body, err := io.ReadAll(r.Body)
 	if err != nil {
 		httpError(w, r, http.StatusBadRequest, "failed to read request body: %s", err)
 		return
@@ -453,7 +453,7 @@ func (a *autographer) handleGetAuthKeyIDs(w http.ResponseWriter, r *http.Request
 		return
 	}
 	if r.Body != nil {
-		body, err := ioutil.ReadAll(r.Body)
+		body, err := io.ReadAll(r.Body)
 		if err != nil {
 			httpError(w, r, http.StatusBadRequest, "failed to read request body: %s", err)
 			return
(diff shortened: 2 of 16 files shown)


oskirby (commented):
I think you missed a few in the tests...
```
handlers_test.go:20:2: "io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details.  (SA1019)
main_test.go:11:2: "io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details.  (SA1019)
signer/apk2/apk2_test.go:5:2: "io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details.  (SA1019)
signer/gpg2/gpg2_test.go:7:2: "io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details.  (SA1019)
signer/xpi/jar_test.go:7:2: "io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details.  (SA1019)
signer/xpi/xpi_test.go:9:2: "io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details.  (SA1019)
make: [Makefile:49: show-lints] Error 1 (ignored)
```

You will also need to update the branch now that #907 has been merged

jmhodges:
Already merged in, actually! Do need to update the `base` for this PR, though. 

Missed those because I'd had an exclusion on "*_test.go" in my editor during the search. 

I'd looked through the CI output and didn't see those errors during draft nor now, so I suppose I should fix the CircleCI config now.

jmhodges:
Ugh, the `staticcheck` one had rolled off the screen and I'd missed it. I'd looked through the other 3. I suspect we should just make the `staticcheck` error and drop `golint` as its been [deprecated in favor of staticcheck](https://github.com/golang/lint?tab=readme-ov-file#readme) and go vet.

jmhodges:
Sigh, `make staticcheck` doesn't even work correctly on macOS.

```
test 0 -eq $(grep -c -Pv '^/go/pkg/mod/|SA1019' /tmp/autograph-staticcheck.txt)
grep: invalid option -- P
usage: grep [-abcdDEFGHhIiJLlMmnOopqRSsUVvwXxZz] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]
/bin/sh: line 0: test: 0: unary operator expected
make: *** [staticcheck] Error 2
```

jmhodges:
Yeah, the problem with the exit codes is that `tee` is eating them in the Makefile targets, but getting it to do the right thing in a Makefile requires a bunch of named pipe nonsense or telling `make` to use bash inside the files. (see https://stackoverflow.com/questions/25306774/best-way-to-handle-pipes-and-their-exit-status-in-a-makefile)

I think the better option is to just drop the `show-lints` stuff. It's not clear what the benefit is. It's output is confusing in the per-lint runs (because it looks like things are erroring) and creates an implicit cache without expiration mechanisms on top of eating the error codes.

jmhodges:
I believe this change is sufficient now

oskirby (approved):
Looks good!

PR #913, opened 2024-06-21
remove golint from build

jmhodges:
This removes golint from the circleci config and the Makefile.

golint has been deprecated in favor of the vet and staticcheck tools.
See https://github.com/golang/lint?tab=readme-ov-file#readme and
https://github.com/golang/go/issues/38968

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,16 +6,13 @@ PACKAGE_NAMES := github.com/mozilla-services/autograph github.com/mozilla-servic
 # The GOPATH isn't always on the path.
 GOPATH := $(shell go env GOPATH)
 
-all: generate test vet lint staticcheck install
+all: generate test vet staticcheck install
 
 # update the vendored version of the wait-for-it.sh script
 install-wait-for-it:
 	curl -o bin/wait-for-it.sh https://raw.githubusercontent.com/vishnubob/wait-for-it/master/wait-for-it.sh
 	sha256sum -c bin/wait-for-it.sh.sha256
 	chmod +x bin/wait-for-it.sh
-
-install-golint:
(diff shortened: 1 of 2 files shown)


jmhodges:
(Output of digging into #911)

Makefile
```diff
@@ -37,17 +34,13 @@ vendor:
...
-lint:
-	$(GOPATH)/bin/golint $(PACKAGE_NAMES) | tee /tmp/autograph-golint.txt
-	test 0 -eq $$(grep -c -Pv 'stutters|suggestions' /tmp/autograph-golint.txt)
-
 # refs: https://github.com/mozilla-services/autograph/issues/247
 check-no-crypto11-in-signers:
 	test 0 -eq $(shell grep -Ri crypto11 signer/*/ | tee /tmp/autograph-crypto11-check.txt | wc -l)
 
 show-lints:
-	-cat /tmp/autograph-golint.txt /tmp/autograph-crypto11-check.txt /tmp/autograph-staticcheck.txt
```
oskirby, line 49: I would rather we simply get rid of the `show-lints` target and output the status of the linters directly to console as they are running.

For example, I think the `check-no-crypto11-in-signers` target could just use the exit code from `grep` and be rewrittten as:

```
# refs: https://github.com/mozilla-services/autograph/issues/247
check-no-crypto11-in-signers:
	! shell grep -Ri crypto11 signer/*/
```

jmhodges: I can do both! I was removing golint because it's unsupported and its docs say to replace it with go vet and staticcheck (like we've almost done). I'll pop up another PR to do that work

oskirby (approved):

jmhodges:
Mooted by #951

PR #914, opened 2024-06-21
remove unneeded grep of staticcheck output

jmhodges:
This grep was a hack to avoid fixing a deprecated API for DSA. The grep
command fails on macOS because `-P` doesn't exist on the default macOS
grep.

Fortunately, the use of the deprecated API was corrected in
https://github.com/mozilla-services/autograph/pull/770 (original ticket:
https://github.com/mozilla-services/autograph/issues/667), so we can
just drop it whole cloth.

In the future, I believe we'll want to use staticcheck's built-in ways
of ignoring certain lints, anyway:
https://staticcheck.io/docs/configuration/#ignoring-problems

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,9 +70,6 @@ race:
 
 staticcheck:
 	$(GOPATH)/bin/staticcheck $(PACKAGE_NAMES) | tee /tmp/autograph-staticcheck.txt
-	# ignore errors in pkgs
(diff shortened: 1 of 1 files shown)


oskirby (approved):
Looks good!

jmhodges:
Mooted by https://github.com/mozilla-services/autograph/pull/951

PR #915, opened 2024-06-21
run gofmt on verifier/contentsignature

jmhodges:
This is the output of running `gofmt -s -w ./verifier/contentsignature`.

Found these changes lurking around when fixing up the ioutil
deprecations in #911


oskirby (approved):
Looks good!

Issue #350, opened 2019-10-02
X5U urls are sent as inaccessible `file://` URLs in development


mythmon:
## STR

1. Run Autograph from docker: `docker run mozilla/autograph`
2. Request a signature using the Normandy development credentials: `curl -XPOST -d '{...}' https://localhost:8765/sign/data`
3. Query the returned x5u to validate the key.

## Expected results

The x5u URL is accessible without additional, undocumented configuration.

## Actual results

The x5u URL is a `file://` URL that is inaccessible in the documented configuration.

This makes it impossible to use Normandy and Autograph in development, making Normandy development significantly harder to get right.

g-k:
@mythmon where does the normandy dev autograph config live?

g-k:
OK so the planned fix is:

- [ ] always return `http://` or `https://` X5Us
- [ ] in dev mode, have autograph serve the files referenced by X5Us from a local static directory `localhost/static/`. This might require changes to https://github.com/mozilla-services/autograph/blob/master/tools/genpki/genpki.go too.

mythmon:
> @mythmon where does the normandy dev autograph config live?

The [default autograph config](https://github.com/mozilla-services/autograph/blob/647e6042d6c33e3322c9a8157abb9a8859efe83c/autograph.yaml#L69-L119) includes the Normandy development config.

leplatrem:
I missed the conversation on Slack, but I share this need :)

This is what devs have to do when setting up a local Remote Settings server: https://remote-settings.readthedocs.io/en/latest/tutorial-local-server.html#configure-multi-signoff

jvehent:
Where are we on this? I remember the conversation but not the resolution. Is that still blocking normandy and kinto?

leplatrem:
It makes local development setup quite tedious, it would be nice to fix. But it's not blocking per-se

PR #917, opened 2024-07-06
AUT-54: Add endpoint to serve local x5u chains

oskirby:
## Description
It has slightly bothered me that our integration tests, and local development experience frequently relies on `/tmp` being mounted into the container to serve certificate chains which are generated and stored locally using the `file://` scheme. It is very janky, inadequately documented, and can lead to a brittle development experience. This caught my attention when my `/tmp` volume got corrupted with some stale state and it wound up breaking everything until I manually deleted the volume.

It has also caused friction for downstream consumers of the Autograph container during development, such as [remote-settings](https://remote-settings.readthedocs.io/en/latest/tutorial-local-server.html#with-multi-signoff-advanced)

I propose the following solution, we create a new endpoint: `/x5u/:keyid/:chainfile` which acts as a mirror to serve the contents of the configured `chainuploadlocation` for a signer, if, and only if, that signer has been configured with a local file upload location (eg: uses the `file://` scheme).

Then, when generating signatures, we check if the signature contains an x5u URL with the file scheme, and replace it with the URL to the new `x5u` endpoint.

This should mean that we no longer need to share `/tmp` between autograph containers, and we can eliminate the use of this volume altogether.

I have added some tests for the endpoint, but I am not completely happy with the coverage as we would need a way to mock out the signers and mess around with their chain upload locations to really get full coverage.

## Reference
See issue: #350 ([AUT-54](https://mozilla-hub.atlassian.net/browse/AUT-54))


bhearsum (dismissed):
This probably ought to be reviewed by @jmhodges or someone else that knows the codebase better, but the idea is certainly solid (and obviously desired by other stakeholders), and this implementation seems fine.

If I understand correctly, these files are written to `ChainUploadLocation` over at https://github.com/mozilla-services/autograph/blob/5f97ba38de22e7859802d285153fdba0a144a20d/signer/contentsignaturepki/upload.go#L30-L31, and we're just serving those indirectly through autograph.

The _most_ robust and production-like way to do this would probably be to add a mock s3 container (such as https://github.com/adobe/S3Mock), but that is both a) more work, and b) may not satisfy the needs of Remote Settings, so I suspect it's probably never going to be worthwhile to do.

handlers.go
```diff
@@ -495,3 +515,76 @@ func (a *autographer) handleGetAuthKeyIDs(w http.ResponseWriter, r *http.Request
 	w.WriteHeader(http.StatusOK)
 	w.Write(signerIDsJSON)
 }
+
+func (a *autographer) handleGetX5U(w http.ResponseWriter, r *http.Request) {
+	if r.Method != "GET" {
+		httpError(w, r, http.StatusMethodNotAllowed, "%s method not allowed; endpoint accepts GET only", r.Method)
+		return
+	}
+	if r.Body != nil {
```
jmhodges, line 524: All of this code can be deleted. There's no `Body` ever sent on `GET`s and we shouldn't do anything with them, anyway. Go will handle the reading of bogus requests when we return the 405.

handlers.go
```diff
@@ -495,3 +515,76 @@ func (a *autographer) handleGetAuthKeyIDs(w http.ResponseWriter, r *http.Request
 	w.WriteHeader(http.StatusOK)
 	w.Write(signerIDsJSON)
 }
+
+func (a *autographer) handleGetX5U(w http.ResponseWriter, r *http.Request) {
+	if r.Method != "GET" {
+		httpError(w, r, http.StatusMethodNotAllowed, "%s method not allowed; endpoint accepts GET only", r.Method)
+		return
```
jmhodges, line 522: This is already being handled in the routing later, no? That `Methods("GET")` call?

(I know we have some other APIs doing this repetitive work, but I think that was a bit of a mistake. I could be convinced the other way, but I know this might all be mooted with the `http.FileServer` work)

handlers_test.go
```diff
@@ -720,60 +790,194 @@ func TestHandleGetAuthKeyIDs(t *testing.T) {
...
+		},
+		{
+			name:            "invalid method HEAD returns 405",
+			method:          "HEAD",
+			url:             "http://foo.bar/x5u/normandy/example.pem",
+			nilBody:         true,
+			expectedStatus:  http.StatusMethodNotAllowed,
+			expectedBody:    "HEAD method not allowed; endpoint accepts GET only\r\nrequest-id: -\n",
+			expectedHeaders: http.Header{"Content-Type": []string{"text/plain; charset=utf-8"}},
+		},
```
jmhodges, line 862: I totally understand the desire here, but I think we want this HTTP method rejection work to be done in the routing layer. It feels a lil like "testing the framework". If you want, you could refactor the routing set up to make it easier to test?

handlers_test.go
```diff
@@ -720,60 +790,194 @@ func TestHandleGetAuthKeyIDs(t *testing.T) {
...
+			method:          "GET",
+			url:             "http://foo.bar/x5u/normandy/example.pem",
+			urlRouteVars:    map[string]string{"keyid": "normandy"}, // missing chainfile
+			nilBody:         true,
+			expectedStatus:  http.StatusInternalServerError,
+			expectedBody:    "route is improperly configured\r\nrequest-id: -\n",
+			expectedHeaders: http.Header{"Content-Type": []string{"text/plain; charset=utf-8"}},
+		},
+		{
+			name:            "GET with directory parent escapes returns 400",
```
jmhodges, line 884: I love these tests!

handlers_test.go
```diff
@@ -720,60 +790,194 @@ func TestHandleGetAuthKeyIDs(t *testing.T) {
...
+			method:          "GET",
+			url:             "http://foo.bar/x5u/normandy/../example.pem",
+			urlRouteVars:    map[string]string{"keyid": "normandy", "chainfile": "../example.pem"},
+			nilBody:         true,
+			expectedStatus:  http.StatusBadRequest,
+			expectedBody:    "Invalid X5U file name '../example.pem'\r\nrequest-id: -\n",
+			expectedHeaders: http.Header{"Content-Type": []string{"text/plain; charset=utf-8"}},
+		},
+		{
+			name:            "GET with directory root escapes returns 400",
```
jmhodges, line 894: nit (and so ignorable if you disagree):Since we make so many of these might be an opportunity to reduce the size of the test with a tiny helper func to generate these "GET for this particular API" structs

handlers.go
```diff
@@ -495,3 +515,76 @@ func (a *autographer) handleGetAuthKeyIDs(w http.ResponseWriter, r *http.Request
...
+			break
+		}
+		chainFilePath := path.Join(parsedURL.Path, pathChainFile)
+		stat, err := os.Stat(chainFilePath)
+		if err != nil || !stat.Mode().IsRegular() {
+			break
+		}
+
+		// Serve files from the chain upload location
+		http.ServeFile(w, r, chainFilePath)
```
jmhodges, line 584: I wrote all the rest of this review and then got here and thought:

Stepping back, could some of this work be done with [http.FileServer](https://pkg.go.dev/net/http#FileServer) instead? Like, in the routing layer, we iterate over all of the `getSigners` just once and create a `http.FileServer(http.Dir(chainUploadLocationDir))` per keyid with a `file://` chainUploadLocation? It would mean not having to do basically any of these checks since [http.Dir](https://pkg.go.dev/net/http#Dir) and FileServer do them for us. The small cost is it would serve the dir under it, but I think we're okay with that for dev and test.

jmhodges (changes requested):

jmhodges:
This is a great idea! I think the `http.FileServer(http.Dir())` change is worth investigating before we ship

handlers.go
oskirby: IIUC, technically this is incorrect. The HTTP protocol doesn't prohibit a `GET` from providing a body, it is just that by convention no one does so.

handlers.go
oskirby: I assumed that there was a reason for having that code, so I simply copied what the other handlers were doing. However, if the routing layer will do this for us I agree it would be a lot cleaner and I can spend another cycle updating the rest of the handlers to drop the unsupported method cases and cleaning up the tests.

handlers_test.go
oskirby: I don't think a method would really clean this up any significant amount and we would still be typing out all the same things, just as method arguments instead of struct initialization values.

handlers.go
jmhodges: Sure, but I forgot to mention that the Go http library will close out the Body for us, anyhow, so it's still redundant.

handlers_test.go
jmhodges: Oh, I was just seeing that most of the fields on the structs were all the same except for like 3. But, again, a nit, and to your taste.

handlers_test.go
oskirby: Sadly, all of these tests are no longer relevant :)

oskirby:
I have refactored this work so that it now uses registers an `http.FileServer` for each signer, and we can rely on the mux router to map requests to the directories for each signer.

signer/signer.go
```diff
@@ -34,6 +34,9 @@ import (
 // IDFormat is a regex for the format IDs must follow
 const IDFormat = `^[a-zA-Z0-9-_]{1,64}$`
 
+// IDFormatRegexp is the compiled regex from IDFormat
+var IDFormatRegexp = regexp.MustCompile(IDFormat)
+
```
jmhodges, line 39: I believe this can be removed now as it doesn't seem to be used

handlers_test.go
```diff
@@ -37,6 +37,96 @@ import (
 	margo "go.mozilla.org/mar"
 )
 
+type HandlerTestCase struct {
```
jmhodges, line 40: I'm having difficulty seeing what's changed functionally in this test code. Could you point me to the actual differences from the original that isn't just refactoring?

handlers.go
```diff
@@ -74,6 +75,22 @@ func logSigningRequestFailure(sigreq formats.SignatureRequest, sigresp formats.S
 	}).Info(fmt.Sprintf("signing operation failed with error: %v", err))
 }
 
+// If the x5u is a local file, mirror it via the `x5u` endpoint instead.
+func mirrorLocalX5U(r *http.Request, response formats.SignatureResponse) string {
```
jmhodges, line 79: I'm getting this confused with other similarly named things. Perhaps `rewriteX5UIfLocalFile`? Something about rewriting, maybe. The `mirror` sounds like it's copying the object from one place to another.

Also, nit: [godoc format](https://go.dev/blog/godoc) is to start the first sentence with the function's name

handlers.go
```diff
@@ -74,6 +75,22 @@ func logSigningRequestFailure(sigreq formats.SignatureRequest, sigresp formats.S
...
+	parsedX5U, err := url.Parse(response.X5U)
```
jmhodges, line 80: Now that there's less to look at I'm wondering: could we do this URL rewriting earlier in the process? That is to say, is it possible to rewrite the X5U URL on the signer at signer creation time and not need all this happening at request time? 

A couple follow-up questions I have after that: Would doing that rewrite earlier (if we can) allow us to simplify `contentsignaturepki.GetX5U`? And would doing the url rewrite earlier also mean the `getMirroredX5U` helper could go away since it doesn't need to know about files anymore? (I know I have a comment on `getMirroredX5U` wondering about whether it needs to exist given the current state of the code, but I'm asking here in case that other question gets a "yes it needs to exist" answer)

docs/endpoints.md
```diff
@@ -271,6 +271,12 @@ See `/sign/data`, the response format is identical.
 A successful request return a `201 Created` with a response
 body containing an S/MIME detached signature encoded with Base 64.
 
+## /x5u/:keyid/:chain
+
+This is an endpoint used to fetch certificate chains which are generated and
+stored locally. If the signer is configured with a `chainuploadlocation` with
+the `file://` scheme, then the contents of that file location are mirrored here.
```
jmhodges, line 278: nit: a brief mention of only wanting to use this for local dev and testing might be nice

jmhodges (changes requested):
Love how small this is getting! And thanks for investigating the Dir stuff! Have some questions about what's left and some tweaks (like the godoc and the extra var kicking around)

handlers_test.go
jmhodges: Whoops, sorry, I thought I deleted this one! I figured out what was going on!

handlers.go
oskirby: In theory it can be done statically by changing the X5U in the signer config to match what gets served by `http.FileServer` but to figure that out what the X5U should be, autograph would need to know its hostname to fully form the URL. And the hostname might not even be knowable at startup if it's running behind proxies or across container boundaries.

On the other hand, handling this at the time of the request is easy because we can just copy the hostname from the HTTP request headers.

handlers.go
```diff
@@ -203,6 +225,8 @@ func (a *autographer) handleSignature(w http.ResponseWriter, r *http.Request) {
 			X5U:        requestedSignerConfig.X5U,
 			SignerOpts: requestedSignerConfig.SignerOpts,
 		}
+		sigresps[i].X5U = rewriteLocalX5U(r, sigresps[i])
```
jmhodges, line 228: nit: this could be done in the struct set up but no more review cycles required to do that

jmhodges (dismissed):

jmhodges (approved):

PR #918, opened 2024-07-12
Experiments for faster testing

oskirby:
The goal here is to make the developer experience faster by eliminating the need to rebuild everything in order to get to a test.


jmhodges:
Another idea: Is there a way to create a `docker compose` entry that allows us to `docker compose run /bin/bash -l` it with the source code as a volume and run our `go build`, `go test`, and such there?

I got close with something like 
[foo.patch](https://github.com/user-attachments/files/16282013/foo.patch) but I'm hitting `.cache` writablility errors and I have to set `export PATH="${PATH}:/usr/lib/go-1.22/bin"` manually. 

```
app@689383b0ec43:~/src/autograph$ go test
failed to initialize build cache at /app/.cache/go-build: mkdir /app/.cache: permission denied
```

oskirby:
My go-to command line invocation to get a shell for testing stuff is `docker compose run --rm app /bin/bash` and it just drops me into a shell inside the container.

The PATH stuff stuff might be some kind of inconsistency in the user environment. IIRC the container is set up so that you get to run with UID 10001 but the PATH variables set at the top of the Dockerfile might only apply to root. We might also need to run as root inside the container to get write access to anything.

jmhodges:
oo, compelling

PR #921, opened 2024-07-15
fix: speed up integration test times & cleanup

bhearsum:
We could probably eke more out of this by rearchitecting how they run, but this covers the low hanging fruit that stuck out to me. On my local machine, it takes integration tests from ~5m to ~1m30s.

In CI it looks like it took the CircleCI integration test from ~12min to ~7min, but only made small gains in the already parallelized integration tests for various signers.

This is mostly a quick hack. Feel free to reject it if it's undesirable for any reason.

### DIFF
--- a/bin/run_integration_tests.sh
+++ b/bin/run_integration_tests.sh
@@ -52,14 +52,16 @@ docker compose run \
 	       -e AUTOGRAPH_URL=http://app:8000 \
 	       --workdir /app/src/autograph/tools/autograph-client \
 	       --entrypoint ./integration_test_api.sh \
-	       app
+	       app &
 docker compose run \
 	       --rm \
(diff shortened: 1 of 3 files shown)


oskirby:
I am little concerned that forking the tests will make it harder to diagnose when a test goes wrong, as it's not clear which log messages came from which test anymore. I prefer the approach we have going on in Github actions where the tests are parallelized across different jobs.

I would actually prefer that we got rid of the `run_integration_tests.sh` script altogether and instead try to make development easier by running a specific integration test if desired (eg: `docker compose run -f tools/autograph-client/integration-tests.yaml <test-name>`

bhearsum:
That's fair! I'm happy to toss this PR out.

PR #922, opened 2024-07-16
create working CODEOWNERS file

jmhodges:
We would like to have code reviewers automatically alerted and assigned
when PRs are posted to autograph codebases. This is the next step
towards that for the autograph repo. One this merge, we'll enable the
automatic assignation in the GitHub repo settings (one day, we'll have
terraform for those settings, I hope).

Autograph is going to use the @mozilla-services/autograph-devs GitHub
team to provide automatic assignation of code reviewers in autograph.
These folks are expected to provide timely reviews of code coming into
autograph. See the autograph code review Confluence page for more.

As a back up and in order to provide some escape valves, we also have a
@mozilla-services/autograph-mergers. This team is folks who are trusted
to merge code but can't provide timely reviews like our SREs. This may
not be a team that we keep around in the future, but given our current
team size and volatility, it's a good back up.

Finally, the `addons.mozilla.org` teams has previously requested alerts
when code is changed in `signer/xpi`. We provide that here using the
same GitHub team as the one used in the mozilla/addons-server
CODEOWNERS.

Updates AUT-18

PR #923, opened 2024-07-17
adjust AMO team in CODEOWNERS

bhearsum:
I'm pretty sure this is what we want. The team that [the original PR referenced](https://github.com/mozilla-services/autograph/pull/922) is in https://github.com/mozilla/addons-server/blob/master/.github/CODEOWNERS which is 5 years out of date.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,3 +1,3 @@
 * @mozilla-services/autograph-devs @mozilla-services/autograph-mergers
 
-signer/xpi/ @mozilla-services/autograph-devs @mozilla-services/autograph-mergers @mozilla/addons-engineering
+signer/xpi/ @mozilla-services/autograph-devs @mozilla-services/autograph-mergers @mozilla-services/addons-engineering


.github/CODEOWNERS
```diff
@@ -1,3 +1,3 @@
 * @mozilla-services/autograph-devs @mozilla-services/autograph-mergers
 
-signer/xpi/ @mozilla-services/autograph-devs @mozilla-services/autograph-mergers @mozilla/addons-engineering
\ No newline at end of file
+signer/xpi/ @mozilla-services/autograph-devs @mozilla-services/autograph-mergers @mozilla/addons-amo-engineering
```
jmhodges, line 3: Ugh, this is telling us that (like my last PR) this CODEOWNERS has an error around addons-amo-engineering. The last time, we still needed to give `@mozilla-services/addons-engineering` the ability to `Write` to this repo.

This time we have to give this team `Write` access, but I can't seem to find a team that matches `@mozilla/addons-amo-engineering` in the search functionality to give them access. I suspect either that team is not a public team, or GitHub isn't allowing me to add anyone from there because I'm not a member of any team in the `@mozilla` org (contrast that with `@mozilla-services`). I'm not able to confirm that `@mozilla/addons-amo-engineering` even exists because I'm not on any team in `@mozilla`.

We may need to engage with the addons team directly to see what exact team they're using and if they can make that team public.

jmhodges (changes requested):

.github/CODEOWNERS
willdurand: Either use the team from the mozilla-services org OR from the mozilla org:

1. https://github.com/orgs/mozilla-services/teams/addons-engineering
2. https://github.com/orgs/mozilla/teams/addons-amo-engineering

jmhodges: Okay, I guess we'll use the one in the same org, then since the other one seems to not be marked as `Public`?

jmhodges:
(I've added their team as `Write` to this repo to clean up the CODEOWNERS lints)

PR #924, opened 2024-07-17
AUT-124: upgrade to go 1.22

bhearsum:
Looks like this is trivial, and @oskirby did all the heavy lifting here already?


oskirby (approved):
I am pleased that it was this easy.

It wasn't just me. @jmhodges also deserves some of the credit for cleaning up a bunch of deprecated packages in the 1.16 -> 1.19 upgrade.

PR #926, opened 2024-07-18
finish off Go 1.22 upgrade

jmhodges:
This patch cleans up some stragglers in the CircleCI config that were
still running on 1.19.

We also change `go.mod` to include the patch portion of the Go version.
Including `x.y.z`, not just `x.y` is what the Go 1.22 tooling (and, I
believe 1.21) does and expects. If we don't include the patch version,
various third party tools like GitHub's CodeQL can break (see
https://github.com/github/codeql/issues/15647 for the last time I ran
into this). This go.mod version change is actually what led me to the
other missing CircleCI version bumps as changing go.mod broke those
CircleCI jobs.

Along the way, we remove some dependency installations from the linting
CircleCI job. While future linting maybe need these dependencies, the
current ones do not.

I've made AUT-158 to consolidate our go linting stuff and avoid some of
these problems. Some (not all) of the rationale for AUT-158 will be
mooted by a move to GitHub Actions.

Along the way, I also ran `go mod tidy` which correctly noted that the
the google/uuid dep is now a direct dep. Hilariously, this is the change
I did first.

do less work in the lint-vet-fmt job

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,13 +8,9 @@ version: 2.1
 jobs:
   lint-vet-fmt:
     docker:
-      - image: debian:bookworm
+      - image: cimg/go:1.22
     steps:
-      - run:
-          name: install dependencies
-          command: |
-            apt-get update
-            apt-get install -y golang build-essential git ca-certificates libltdl-dev
+      # TODO(AUT-158): this linting always succeeds even if there are errors, and future correct linting may need correct cgo compliation
       - checkout
       - run:
           name: check crypto11 not used in signers
@@ -38,7 +34,7 @@ jobs:
           command: |
             make vet
             make -C tools/autograph-monitor vet
-            make -C verifier/contentsignature vet 
+            make -C verifier/contentsignature vet
       - run:
           name: run staticcheck
           command: |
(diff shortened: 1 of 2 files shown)


bhearsum (dismissed):
Looks reasonable. Sorry I missed this stuff :(.

I'm happy to help overall/improve the CI situation as well.

jmhodges:
Sorry to request the re-review, but I realized we could just not install deps at all in the linting CircleCI job. Have simplified the code and corrected the PR summary and commit message.

bhearsum:
> Sorry to request the re-review, but I realized we could just not install deps at all in the linting CircleCI job. Have simplified the code and corrected the PR summary and commit message.

Can you clarify why this is?

jmhodges:
> > Sorry to request the re-review, but I realized we could just not install deps at all in the linting CircleCI job. Have simplified the code and corrected the PR summary and commit message.
> 
> Can you clarify why this is?

The apt dependencies that were being installed aren't actually being used in any of the linting processes, so they're irrelevant. Once you remove those, the failures of `apt-get update` in the cimg/go:1.22 images stopped mattering because we weren't needing to run it.

bhearsum (approved):
Thanks for the additional info!

PR #929, opened 2024-07-19
Add statsd request success function on handleGetAuthKeyIDs

hneiva:


main_test.go
```diff
@@ -458,3 +460,23 @@ func TestPortOverride(t *testing.T) {
...
 	}
 }
+
+func TestStatsWriteSuccess(t *testing.T) {
+	t.Parallel()
+	req, err := http.NewRequest("GET", "https://foo.bar", nil)
+	if err != nil {
+		t.Errorf("Error setting up request.")
+	}
+	ag.statsWriteSuccess(req, "test")
```
bhearsum, line 470: Ideally, the `stats` object on `ag` would be some sort of mock or other thing that we can inspect afterwards. Or at the very least, we need to ensure that it was called appropriately.

jmhodges (commented):
Yeah, agreed with Ben's point about the null object.  

We'll need a baseline `attempts` here. 

And this would possibly be nice as capturing the status codes returned. Sometimes these metric libraries provide that out of the box (perhaps in a more recent version than we have). Does it? Or do we need to write the Write/WriteHeader wrappers ourselves?

Finally, I don't see an explanation of this change but I'm assuming you only did one HTTP endpoint to see if this is what we wanted, yeah? Def want the rest.

hneiva:
I was not able to find a good way of mocking Stats for these tests.

Updated the `statsWriteSuccess` function to return any errors, and added a test case that captures an error.

>  I'm assuming you only did one HTTP endpoint to see if this is what we wanted, yeah? Def want the rest.

Exactly. Wanted to keep the PR minimal for adding just the functionality, will add other methods later.

bhearsum:
> And this would possibly be nice as capturing the status codes returned. Sometimes these metric libraries provide that out of the box (perhaps in a more recent version than we have). Does it? Or do we need to write the Write/WriteHeader wrappers ourselves?

@hneiva - We can forego any of the test mocking if you'd like, but can we get this part addressed?

jmhodges:
For my memory: there is a `NoOpClient` we can use instead of doing `nil` checks everywhere. https://pkg.go.dev/github.com/datadog/datadog-go/statsd#NoOpClient

bhearsum:
@hneiva  - Is review still needed on this iteration? I think it was decided out of band that some changes would be made?

hneiva:
No need for this anymore since the work Jeff and I did on other PRs

PR #933, opened 2024-07-30
record statsd metrics for availability

jmhodges:
This records the HTTP response status codes returned by autograph to our
existing statsd infrastructure. It will be helpful for monitoring uptime
and other errors.

This was done in conjunction with @hneiva 

Metrics will look like "<api name>.response.status.<status code>" (the
HTTP status code autograph is trying to return) and and "<api
name>.request.attempts" (the number of requests that have come into
autograph from its POV). We also include some
`api.response.status.<status code>` and `api.request.attempts` metrics
to more easily calculate the percentage of successful requests across
all of autograph's public API.

We do so by writing a small middleware that wraps the given
http.ResponseWriter, writes the status code to statsd when `Write` and
`WriteHeader` are called, handles the interaction between `Write` and
`WriteHeader`, and avoids writing the headers and the status code twice.

We also add a mock implemenation of `statsd.ClientInterface` for
the rare cases we want to test what metrics are being sent (which we
have in this patch). It was generated by `mockgen` (sibling to the
uber/mock library added). How to reproduce that mock is included in its
comment at the top of the file (per usual).

Updates AUT-150


jmhodges:
Depends on #932 to ship

stats.go
```diff
@@ -18,8 +20,64 @@
...
+	stats        statsd.ClientInterface
+
+	headerWritten *atomic.Bool
+}
+
+func (w *statsdWriter) Write(b []byte) (int, error) {
+	if !w.headerWritten.Load() {
+		w.WriteHeader(http.StatusOK)
+	}
+	return w.ResponseWriter.Write(b)
```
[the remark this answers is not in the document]
jmhodges: This was a false positive and I've responded to with details of it being a false positive on the original PR #932

stats.go
```diff
@@ -34,3 +36,60 @@ func (a *autographer) addStats(conf configuration) error {
 	log.Infof("Statsd enabled at %s with namespace %s", conf.Statsd.Addr, conf.Statsd.Namespace)
 	return nil
 }
+
+// newStatsdWriter returns a new http.ResponseWriter that sends HTTP response
+// statuses as metrics to statsd. The metric emitted is prefixed with and
+// suffixed with ".response.status.<status code>". The whole metric for
+// "myhandler" will be something like "myhandler.response.status.200". The
+// returned http.ResponseWriter doesn't support the http.Flusher or
+// http.Hijacker type.
```
bhearsum, line 45: This is compact and informative comment, thank you for writing it!

main.go
```diff
@@ -202,16 +202,18 @@ func run(conf configuration, listen string, debug bool) {
...
-	router.HandleFunc("/__monitor__", monitor.handleMonitor).Methods("GET")
-	router.HandleFunc("/sign/files", ag.handleSignature).Methods("POST")
-	router.HandleFunc("/sign/file", ag.handleSignature).Methods("POST")
-	router.HandleFunc("/sign/data", ag.handleSignature).Methods("POST")
-	router.HandleFunc("/sign/hash", ag.handleSignature).Methods("POST")
-	router.HandleFunc("/auths/{auth_id:[a-zA-Z0-9-_]{1,255}}/keyids", ag.handleGetAuthKeyIDs).Methods("GET")
+	router.HandleFunc("/__heartbeat__", statsMiddleware(ag.handleHeartbeat, "http.heartbeat", stats)).Methods("GET")
+	router.HandleFunc("/__lbheartbeat__", statsMiddleware(handleLBHeartbeat, "http.lbhearbeat", stats)).Methods("GET")
+	router.HandleFunc("/__version__", statsMiddleware(handleVersion, "http.__version", stats)).Methods("GET")
+	router.HandleFunc("/__monitor__", statsMiddleware(monitor.handleMonitor, "http.monitor", stats)).Methods("GET")
```
bhearsum, line 211: Do we really need to send stats for _all_ of these endpoints? `/__heartbeat__` seems like all we would need to measure uptime. (I guess `/__lbheartbeat__` as well if we care about measuring full uptime vs. app up but db down uptime.)

stats.go
```diff
@@ -34,3 +36,60 @@ func (a *autographer) addStats(conf configuration) error {
...
+		w = newStatsdWriter(w, handlerName, stats)
+		h(w, r)
+	}
+}
+
+// apiStatsMiddleware is an HTTP handler that emits the metrics as described in
+// statsMiddleware, but also increments equivalent metrics as if their
+// handlerName was "api". This is a shortcut to make it easier to get our uptime
+// across the whole over our actual public API.
+func apiStatsMiddleware(h http.HandlerFunc, handlerName string, stats statsd.ClientInterface) http.HandlerFunc {
```
bhearsum, line 89: It seems unfortunate that we need to duplicate data like this. Could we avoid it if we had a common prefix for all of the things that use this? Eg: `http.api`, to catch both the `sign` and `auths` endpoints?

internal/mockstatsd/clientinterface.go
```diff
@@ -0,0 +1,265 @@
+// Code generated by MockGen. DO NOT EDIT.
+// Source: vendor/github.com/DataDog/datadog-go/statsd/statsd.go
+//
+// Generated by this command:
+//
+//	mockgen -package mockstatsd -source vendor/github.com/DataDog/datadog-go/statsd/statsd.go -destination internal/mockstatsd/clientinterface.go -imports github.com/mozilla-services/autograph/vendor/github.com/DataDog/datadog-go/statsd=github.com/DataDog/datadog-go/statsd ClientInterface
```
bhearsum, line 6: It looks like mockgen is long deprecated/archived - is there something maintained we ought to be using for this instead?

bhearsum (commented):
I don't have anything that _necessarily_ blocks, but there's a few questions below, so I'm leaving this unapproved for the moment.

stats.go
jmhodges: Yeah, that's fair. I'm used to having to do duplicative rollups due to metric renames, etc., but we have a greenfield here. I'll make the change!

main.go
jmhodges: Oh, "uptime" in my prior experience has changed to mean more like "availability", or "success rate", and isn't really about whether the service is "running".

These are heartbeats so that we know the service has come up correctly during deploy and so that the service has traffic routed to it by the loadbalancers. But they don't tell us much about how successfully we're handling user requests, so we need the actual counts from the API calls.

internal/mockstatsd/clientinterface.go
jmhodges: Oh, you might be looking at the https://github.com/golang/mock one, but I've installed https://github.com/uber-go/mock here in the go.mod and as the mockgen binary locally. That one is more supported than the other.

Though, I do dream of just taking patreon money and just maintaining this thing better than both.

stats.go
jmhodges: Done!

main.go
jmhodges: But I think I'm just sloppy with it. I def meant "availability" or "success rate"

jmhodges:
lol docker hub is 503'ing

main.go
bhearsum: If I understand correctly, you're saying that what we really care about is the counts from the actual API that distribution services use, which makes sense to me. I guess I'm still not clear on why it's worth instrumenting these endpoints though? And especially why we're instrumenting `__version__`. It's not a major objection, just curious.

internal/mockstatsd/clientinterface.go
bhearsum: Ah; my bad. Thanks for the pointer!

bhearsum (approved):

main.go
jmhodges: The heartbeat's to help discover problems with them (so we can figure out why an lb is dropping the machine: did it reach the process and it was erroring or was it pointing at a process that was already down?) and "do all the same things" for the `__version__` one. Just in case we, I dunno, kept crashing the process whenever __version__ got hit or something.

PR #934, opened 2024-07-30
remove nil checks around uses of autographer.stats

jmhodges:
This removes the now unneeded checks to see if `autographer.stats` is
nil wherever we gather stats.

We also remove some nil checks in signer.StatsClient that are no longer
workable as it accepts the statds.ClientInterface type.

Some of the older code was checking for error returns that they probably
shouldn't (logging about metrics problems tends to clutter up code and
the alerting should already tell you you have a problem), but we leave
those alone for now.

Updates AUT-159


bhearsum (dismissed):

jmhodges:
Ugh, when you change the base branch, it dismisses reviews. Sorry.

bhearsum (approved):

PR #935, opened 2024-07-30
bump postgres to 14 in our tests

jmhodges:
We've moved to postgres 14 in staging and production for a while.


oskirby (approved):
Looks good to me!

PR #938, opened 2024-08-01
Fetch autograph logs in Github actions integration tests

oskirby:
This should add a step after an integration test runs that outputs the logs from the autograph container to the github workflow console. This should better allow us to investigate the cause of a failure. Note that the use of `if: ${{ always() }}` should cause this step to be run even if the preceeding steps result in a failure.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -123,7 +123,7 @@ jobs:
         shell: bash
         if: ${{ always() }}
         run: |
-          if echo "${{ matrix.testcase }}" | grep -q hsm;
+          if echo "${{ matrix.testcase }}" | grep -q hsm; then
             docker compose logs app-hsm
           else
             docker compose logs app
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -118,3 +118,13 @@ jobs:
       - name: Running ${{ matrix.testcase }}
         shell: bash
         run: docker compose -f tools/autograph-client/integration-tests.yml run ${{ matrix.testcase }}
+
+      - name: Fetching autograph logs
+        shell: bash
+        if: ${{ always() }}
```
bhearsum, line 124: Will this cause issues if we don't reach the previous step? Maybe we don't care if this fails in weird ways if something early (eg: checking out) fails?

oskirby: If the container never starts, this just outputs nothing so the failure mode is a no-op in this case.

I also don't think it matters if the job fails in weird ways at this point as we are already dealing with an error. What's one more?

bhearsum (approved):

PR #793, opened 2021-11-03
signer: xpi: remove RSA key cache

g-k:
refs: ITSEC-184

PR #939, opened 2024-08-01
AUT-164: remove rsa key caching feature

bhearsum:
This has been unused for ~3 years. This is essentially an unbitrotted version of https://github.com/mozilla-services/autograph/pull/793.

### DIFF
--- a/signer/xpi/README.md
+++ b/signer/xpi/README.md
@@ -57,8 +57,7 @@ signers:
         -----END PRIVATE KEY-----
 ```
 
-The signer can also include optional config params for an RSA key cache
-and recommendations file:
+The signer can also include optional config params for a recommendations file:
 
 ``` yaml
 signers:
@@ -73,14 +72,6 @@ signers:
         partner: true
       relative_start: 0h
       duration: 26298h
-    # RSA key gen is slow and CPU intensive, so we can optionally
-    # pregenerate and cache keys with a worker pool
-    rsacacheconfig:
-      numkeys: 25
-      numgenerators: 2
-      generatorsleepduration: 1m
-      fetchtimeout: 100ms
-      statssamplerate: 1m
     certificate: |
         -----BEGIN CERTIFICATE-----
         MIIH0zCCBbugAwIBAgIBATANBgkqhkiG9w0BAQsFADCBvDELMAkGA1UEBhMCVVMx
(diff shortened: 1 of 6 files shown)


signer/xpi/x509.go
```diff
@@ -12,76 +12,17 @@
...
-			// because the end entity has the same key size has
-			// the signer, but we're paranoid so handling it
-			log.Warnf("WARNING: xpi rsa cache returned a key of size %d when %d was requested", key.N.BitLen(), size)
-			key, err = rsa.GenerateKey(s.rand, size)
-		}
-	case <-time.After(s.rsaCacheFetchTimeout):
-		// generate a key if none available
-		key, err = rsa.GenerateKey(s.rand, size)
-	}
+	key, err = rsa.GenerateKey(s.rand, size)
```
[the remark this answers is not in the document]
bhearsum: This might be technically valid? But we have [a min size check](https://github.com/mozilla-services/autograph/blob/5f97ba38de22e7859802d285153fdba0a144a20d/signer/xpi/xpi.go#L213-L215), and [2048 is the set as the min size](https://github.com/mozilla-services/autograph/blob/5f97ba38de22e7859802d285153fdba0a144a20d/signer/xpi/xpi.go#L62).

jmhodges (approved):

signer/xpi/README.md
```diff
@@ -73,14 +72,6 @@ signers:
...
       relative_start: 0h
       duration: 26298h
-    # RSA key gen is slow and CPU intensive, so we can optionally
-    # pregenerate and cache keys with a worker pool
-    rsacacheconfig:
-      numkeys: 25
-      numgenerators: 2
-      generatorsleepduration: 1m
-      fetchtimeout: 100ms
-      statssamplerate: 1m
```
willdurand, line 83: AMO has the same config for localdev, see: https://github.com/mozilla/addons-server/blob/78a854871370bb2d53940a70b4048dc203be4a30/scripts/autograph_localdev_config.yaml#L142-L149 -  I am guessing that change will either ignore or break the localdev Autograph service? In any case, I'll file an issue to update the config.

willdurand (approved):
r+wc

signer/xpi/README.md
bhearsum: Yeah, I'm not sure if it'll break or not, but there's obviously no reason to remove it soon!

jmhodges: For clarity: the options will, yep, definitely be ignored in the yaml file once you update to this version of autograph and you won't see a break.

PR #941, opened 2024-08-05
run unit tests once when enabling race detector

jmhodges:
Previously, running

```
docker compose build unit-test && docker compose \
run -e RACE_TEST=1 unit-test
```

would run the tests once without the race
detector and then again with the race detector.

Found this while discussing how to handle AUT-162 with Ben Hearsum.

Along the way, remove the duplicate running of
`verifier/contentsignature` (it's already in PACKAGE_NAMES in the
top-level Makefile).

I'm not sure we want all this infrastructure for running unit tests, but
I'm not going to be the one to turn off coverage calculations. (Assuming
they still are working correctly, which I'm realizing I've not checked
in on since joining the project.)

Updates AUT-188
Updates AUT-189

### DIFF
--- a/bin/run_unit_tests.sh
+++ b/bin/run_unit_tests.sh
@@ -13,15 +13,13 @@ REPORT_COVERAGE=${CI:-""}
 if [ "$REPORT_COVERAGE" = "true" ]; then
     make install-goveralls
 fi
-# run app unit tests
-make test
-# run monitor unit tests
-make -C tools/autograph-monitor test
 
 if [ "$RACE_TEST" = "1" ]; then
     make race
     make -C tools/autograph-monitor race
-    make -C verifier/contentsignature race
+else
+    make test
+    make -C tools/autograph-monitor test
 fi
 
 if [ "$REPORT_COVERAGE" = "true" ]; then


bin/run_unit_tests.sh
```diff
@@ -13,15 +13,13 @@ REPORT_COVERAGE=${CI:-""}
...
     make install-goveralls
 fi
-# run app unit tests
-make test
-# run monitor unit tests
-make -C tools/autograph-monitor test
 
 if [ "$RACE_TEST" = "1" ]; then
     make race
     make -C tools/autograph-monitor race
```
bhearsum, line 19: Do we know why `autograph-monitor` is treated specially rather than being listed in `PACKAGE_NAMES`?

jmhodges: Looked into this, and I'm not sure. The tooling may have been worse back in the day and not allowed you to run `go test` on two different `package main`s, or the original developers expected to have to do more in the tools/autograph-monitor/Makefile to make the package testable.

So, to prove that out, I now have a patch that adds it to `PACKAGE_NAMES` and removes the `make` calls and the tests work correctly (added a failing to test to see that it was run, too). Tangentially, I kinda hate PACKAGE_NAMES and would love to get the `./...` alias working for all of this kind of stuff but I don't have a perfect solution for all the work it's doing, right now. 

I can post that patch as a follow-up. Would rather keep this one focused on the deduping.

bhearsum: Yeah, separate patch is fine.

bhearsum (approved):

bin/run_unit_tests.sh
jmhodges: Oh, just dug deeper, and it looks like maybe it was for the separated coverage calculation? https://github.com/mozilla-services/autograph/blob/6bfbe916169f2f79c19f72cc8f3a7b62155720d8/bin/run_unit_tests.sh#L25

I think it'll be fine if we combine it in, but I'll admit I've not use the goveralls stuff.

PR #942, opened 2024-08-05
correct signer/apk2 data races

jmhodges:
This change fixes the obvious data races in the signer/apki2 tests. It
also adds the `EMPTY_TESTCACHE` environment variable to
run-unit-tests.sh to allow easy resetting of the Go test cache to
reproduce data races easily.

When run with `docker run unit-test -e RACE_TEST=1`, the following data race was found:

```
==================
WARNING: DATA RACE
Read at 0x000000988860 by goroutine 21:
  github.com/mozilla-services/autograph/signer/apk2.TestNewSigner.func1()
      /app/src/autograph/signer/apk2/apk2_test.go:39 +0x40
  testing.tRunner()
      /usr/lib/go-1.22/src/testing/testing.go:1689 +0x180
  testing.(*T).Run.gowrap1()
      /usr/lib/go-1.22/src/testing/testing.go:1742 +0x40

Previous write at 0x000000988860 by goroutine 22:
  github.com/mozilla-services/autograph/signer/apk2.TestNewSigner.func2()
      /app/src/autograph/signer/apk2/apk2_test.go:48 +0x3c
  testing.tRunner()
      /usr/lib/go-1.22/src/testing/testing.go:1689 +0x180
  testing.(*T).Run.gowrap1()
      /usr/lib/go-1.22/src/testing/testing.go:1742 +0x40

Goroutine 21 (running) created at:
  testing.(*T).Run()
      /usr/lib/go-1.22/src/testing/testing.go:1742 +0x5e4
  github.com/mozilla-services/autograph/signer/apk2.TestNewSigner()
      /app/src/autograph/signer/apk2/apk2_test.go:36 +0x44
  testing.tRunner()
      /usr/lib/go-1.22/src/testing/testing.go:1689 +0x180
  testing.(*T).Run.gowrap1()
      /usr/lib/go-1.22/src/testing/testing.go:1742 +0x40

Goroutine 22 (running) created at:
  testing.(*T).Run()
      /usr/lib/go-1.22/src/testing/testing.go:1742 +0x5e4
  github.com/mozilla-services/autograph/signer/apk2.TestNewSigner()
      /app/src/autograph/signer/apk2/apk2_test.go:45 +0x60
  testing.tRunner()
      /usr/lib/go-1.22/src/testing/testing.go:1689 +0x180
  testing.(*T).Run.gowrap1()
      /usr/lib/go-1.22/src/testing/testing.go:1742 +0x40
==================
```

This was caused by tests modifying two global variables `apk2signerconf`
and `apk2signerconfModeV3enabled`. This was worsened by the tests being
marked as `t.Parallel` allowing them to be run concurrently.

This change removes those globals and provides their data from two new
functions in the tests, instead. The names are still bad in some cases,
but we're fixing problems, not trying to clean up the world.

Updates AUT-188

### DIFF
--- a/bin/run_unit_tests.sh
+++ b/bin/run_unit_tests.sh
@@ -18,6 +18,11 @@ make test
 # run monitor unit tests
 make -C tools/autograph-monitor test
 
+EMPTY_TESTCACHE=${EMPTY_TESTCACHE:-""}
+if [ "$EMPTY_TESTCACHE" = "1" ]; then
+    go clean -testcache
+fi
+
 if [ "$RACE_TEST" = "1" ]; then
     make race
     make -C tools/autograph-monitor race
--- a/signer/apk2/apk2_test.go
+++ b/signer/apk2/apk2_test.go
@@ -36,7 +36,7 @@ func TestNewSigner(t *testing.T) {
 	t.Run("valid no mode", func(t *testing.T) {
 		t.Parallel()
 
-		s := assertNewSignerWithConfOK(t, apk2signerconf)
+		s := assertNewSignerWithConfOK(t, newApk2SignerConf())
 		if s.v3Enabled {
 			t.Fatalf("%s: unexpectedly enabled v3 signing", t.Name())
 		}
@@ -45,8 +45,9 @@ func TestNewSigner(t *testing.T) {
 	t.Run("valid empty mode", func(t *testing.T) {
 		t.Parallel()
 
-		apk2signerconf.Mode = ""
-		s := assertNewSignerWithConfOK(t, apk2signerconf)
+		missingMode := newApk2SignerConf()
+		missingMode.Mode = ""
+		s := assertNewSignerWithConfOK(t, missingMode)
 		if s.v3Enabled {
 			t.Fatalf("%s: unexpectedly enabled v3 signing", t.Name())
 		}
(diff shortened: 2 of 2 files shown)


bhearsum (approved):
Just to be clear: the problem is the parallel accesses of, eg: `apk2signerconf` in multiple goroutines in the test code? (Not any usage of the configuration in `apk2.go` itself?)

I ask because the only thing changing here AFAICT is that we have multiple instances of each config (but they all have exactly the same data).

jmhodges:
> Just to be clear: the problem is the parallel accesses of, eg: `apk2signerconf` in multiple goroutines in the test code? (Not any usage of the configuration in `apk2.go` itself?)
> 
> I ask because the only thing changing here AFAICT is that we have multiple instances of each config (but they all have exactly the same data).

It was the writes to the fields of the global variables that were the problem, made even worse by the fact that the writes were concurrent. Like [this one](https://github.com/mozilla-services/autograph/blob/6e525a6f89512fd60859f1f5ab38ca51f356eccf/signer/apk2/apk2_test.go#L92) or [this one](https://github.com/mozilla-services/autograph/blob/6e525a6f89512fd60859f1f5ab38ca51f356eccf/signer/apk2/apk2_test.go#L76)

The globals were having their fields modified in different tests. That's bad enough because, as usual with globals in tests, their values weren't getting reset carefully, but on top of that the `t.Parallel` calls were allowing the `testing` library to run those tests concurrently in different goroutines.

PR #943, opened 2024-08-06
move tools/autograph-monitor into top-level make

jmhodges:
This is review feedback from a previous patch.
https://github.com/mozilla-services/autograph/pull/941#discussion_r1705636141

This patches moves tools/autograph-monitor is `PACKAGE_NAMES` in the
top-level Makefile and removes the separated make calls.

This might mess up the coverage calculations we were previously doing,
but I'm not sure we've been getting much value out of those. We can fix
them up later if we do.

This also tries to clean up some unusual ways the autograph-monitor is
referenced in the rest of the Makefile. I'm not sure what that was
about. Perhaps, at some point, the original authors had code files in
that directory with different `package`s in them?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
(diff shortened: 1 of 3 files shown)


jmhodges:
Adding ben to the reviewers since he was the original requestor for this

PR #944, opened 2024-08-06
AUT-162: fix flakey gpg tests

bhearsum:
As expected, these are easily fixed by using distinct keychains for each test.

This PR will initially be opened with an additional commit that makes these tests fail more or less 100% of the time. I will then follow-up with another commit with the fix showing that it works. Finally, I will update the PR with just the fix, and provide links to the first two test runs for comparison.

Test run prior to fix
=======
Test run - failed: https://app.circleci.com/pipelines/github/mozilla-services/autograph/2507/workflows/4b597895-de8c-4262-893c-83b94d75b5ca/jobs/12143
Log excerpt:
```
panic: Fail in goroutine after TestSignData has completed [recovered]
	panic: Fail in goroutine after TestSignData has completed
	panic: Fail in goroutine after TestSignData has completed

goroutine 1662 [running]:
testing.tRunner.func1.1()
	/usr/lib/go-1.22/src/testing/testing.go:1608 +0x92
panic({0x605f00?, 0xc0009365d0?})
	/usr/lib/go-1.22/src/runtime/panic.go:770 +0x132
testing.(*common).Fail(0xc0000cad00)
	/usr/lib/go-1.22/src/testing/testing.go:952 +0xd4
testing.(*common).Fail(0xc0001876c0)
	/usr/lib/go-1.22/src/testing/testing.go:946 +0x3c
testing.(*common).Fail(0xc0004edba0)
	/usr/lib/go-1.22/src/testing/testing.go:946 +0x3c
testing.tRunner.func1.2({0x605f00, 0xc0009365c0})
	/usr/lib/go-1.22/src/testing/testing.go:1615 +0x49
testing.tRunner.func1()
	/usr/lib/go-1.22/src/testing/testing.go:1634 +0x377
panic({0x605f00?, 0xc0009365c0?})
	/usr/lib/go-1.22/src/runtime/panic.go:770 +0x132
testing.(*common).Fail(0xc0000cad00)
	/usr/lib/go-1.22/src/testing/testing.go:952 +0xd4
testing.(*common).Fail(0xc0001876c0)
	/usr/lib/go-1.22/src/testing/testing.go:946 +0x3c
testing.(*common).Fail(0xc0004edba0)
	/usr/lib/go-1.22/src/testing/testing.go:946 +0x3c
testing.(*common).FailNow(0xc0004edba0)
	/usr/lib/go-1.22/src/testing/testing.go:981 +0x26
testing.(*common).Fatalf(0xc0004edba0, {0x65ed6b?, 0x400?}, {0xc00028ef00?, 0x700?, 0x0?})
	/usr/lib/go-1.22/src/testing/testing.go:1089 +0x5e
github.com/mozilla-services/autograph/signer/gpg2.TestSignData.func1.1.2(0xc0004edba0)
	/app/src/autograph/signer/gpg2/gpg2_test.go:381 +0x8db
testing.tRunner(0xc0004edba0, 0xc00010a880)
	/usr/lib/go-1.22/src/testing/testing.go:1689 +0xfb
created by testing.(*T).Run in goroutine 466
	/usr/lib/go-1.22/src/testing/testing.go:1742 +0x390
FAIL	github.com/mozilla-services/autograph/signer/gpg2	8.808s
```

Test run with fix
=======
Test run - passed: https://app.circleci.com/pipelines/github/mozilla-services/autograph/2508/workflows/157fa0db-fda1-49f2-85ed-a5778197388c/jobs/12145

### DIFF
--- a/signer/gpg2/gpg2_test.go
+++ b/signer/gpg2/gpg2_test.go
@@ -2,6 +2,7 @@ package gpg2
 
 import (
 	"bytes"
+	"crypto/md5"
 	_ "embed"
 	"fmt"
 	"os"
@@ -14,7 +15,14 @@ import (
 )
 
 func assertNewSignerWithConfOK(t *testing.T, conf signer.Configuration) *GPG2Signer {
-	s, err := New(conf)
+	// These test names end up quite long, and some versions of GPG don't work well
+	// with very long paths. See https://bugzilla.redhat.com/show_bug.cgi?id=1813705
+	// To work around this, we use a hashed version of the test name. This keeps
+	// the prefix consistent across runs while still being short enough.
+	hashedName := md5.Sum([]byte(t.Name()))
+	tmpDirPrefix := fmt.Sprintf("%x", hashedName)
+
+	s, err := New(conf, tmpDirPrefix)
 	if s == nil {
 		t.Fatal("expected non-nil signer for valid conf, but got nil signer and err %w", err)
 	}
@@ -67,7 +75,7 @@ func assertNewSignerWithConfOK(t *testing.T, conf signer.Configuration) *GPG2Sig
 }
 
 func assertNewSignerWithConfErrs(t *testing.T, invalidConf signer.Configuration) {
-	s, err := New(invalidConf)
+	s, err := New(invalidConf, "")
 	if s != nil {
 		t.Fatalf("expected nil signer for invalid conf, but got non-nil signer %v", s)
 	}
@@ -530,15 +538,14 @@ func TestGPG2Signer_SignFiles(t *testing.T) {
 			}
 
 			assertClearSignedFilesVerify(t, s, "verify-debsigned-files", gotSignedFiles)
+			matches, err := filepath.Glob(filepath.Join(s.tmpDir, "sign_files*", "*"))
+			if err != nil {
+				t.Fatal(err)
+			}
+			if len(matches) != 0 {
+				t.Fatalf("sign files did not clean up its temp input directories: %s", matches)
+			}
 		})
-	}
-
-	matches, err := filepath.Glob("/tmp/autograph_*sign_files*")
-	if err != nil {
-		t.Fatal(err)
-	}
-	if len(matches) != 0 {
-		t.Fatalf("sign files did not clean up its temp input directories: %s", matches)
 	}
 }
 
--- a/signer/gpg2/gpg2_test.go
+++ b/signer/gpg2/gpg2_test.go
@@ -356,13 +356,17 @@ func TestSignData(t *testing.T) {
 					t.Fatal(err)
 				}
 
-				defer os.Remove("/tmp/autograph_test_gpg2_keyring.gpg")
-				defer os.Remove("/tmp/autograph_test_gpg2_secring.gpg")
-				defer os.Remove("/tmp/autograph_test_gpg2_keyring.gpg~")
+				tmpKeyringDir, err := os.MkdirTemp("", "keyring")
+				if err != nil {
+					t.Fatal(err)
+				}
+				defer os.RemoveAll(tmpKeyringDir)
+
+				tmpKeyring := filepath.Join(tmpKeyringDir, "keyring.gpg")
 
 				// call gnupg to create a new keyring, load the key in it
 				gnupgCreateKeyring := exec.Command("gpg", "--no-default-keyring",
-					"--keyring", "/tmp/autograph_test_gpg2_keyring.gpg",
+					"--keyring", tmpKeyring,
 					"--import", tmpPublicKeyFile.Name())
 				out, err := gnupgCreateKeyring.CombinedOutput()
 				if err != nil {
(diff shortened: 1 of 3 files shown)


signer/gpg2/gpg2_test.go
```diff
@@ -529,6 +537,8 @@ func TestGPG2Signer_SignFiles(t *testing.T) {
 		})
 	}
 
+	// This is not an ideal check, and can cause failures when other code
+	// or tests fail to clean up properly - but it's probably good enough.
 	matches, err := filepath.Glob("/tmp/autograph_*sign_files*")
```
jmhodges, line 542: Ugh, I, unfortunately, think this is something that will definitely cause flakes. Would it be possible to pass in a `prefix string` into `gpg2.New` and use it everywhere we're creating temp files? The place its called in `addSigners` could just us a static prefix (instead of being configurable) for now, but it would clear up all the other problems that this could happen might be straight-forward if we do that. 

(I could be wrong about it being straight-forward and I'm happy to hear I've missed something important. I know it would mean having to look at all the other `os.Remove*` deals.)

jmhodges (changes requested):

signer/gpg2/gpg2_test.go
bhearsum: Hm, I'm not sure that will guarantee that we're free of flakes still, but it wouldn't certtainly reduce the possibility. Maybe we ought to pass a directory to create the keyring in instead, as I'm already doing in the other test in this patch? That way there's no globbing or guesswork that has potential interference.

jmhodges: So sorry, yes, I def meant what you're describing and didn't think through the mechanics of needing to pass in a directory vs a prefix passed in when creating that directory (and then deleteling stuff that only matched the temp dir with that prefix; that's where I failed to think it through: the prefix would only be in the middle of the path!)

bhearsum: I think this is all fixed now. Thanks for the push on it - I ended up catching a longstanding issue with this assertion because of it!

signer/gpg2/gpg2_test.go
```diff
@@ -14,7 +15,14 @@ import (
 )
 
 func assertNewSignerWithConfOK(t *testing.T, conf signer.Configuration) *GPG2Signer {
-	s, err := New(conf)
+	// These test names end up quite long, and some versions of GPG don't work well
+	// with very long paths. See https://bugzilla.redhat.com/show_bug.cgi?id=1813705
+	// To work around this, we use a hashed version of the test name. This keeps
+	// the prefix consistent across runs while still being short enough.
+	hashedName := md5.Sum([]byte(t.Name()))
```
jmhodges, line 22: Great catch!

signer/gpg2/gpg2_test.go
```diff
@@ -529,7 +541,7 @@ func TestGPG2Signer_SignFiles(t *testing.T) {
 		})
 	}
 
-	matches, err := filepath.Glob("/tmp/autograph_*sign_files*")
+	matches, err := filepath.Glob("/tmp/autograph_*sign_files*/*")
```
jmhodges, line 544: Since we're doing things in parallel, I think we could still accidentally hit this if another test is running that's writing this same pattern, no? Unless we pass in the full dir path

signer/gpg2/gpg2_test.go
```diff
@@ -14,7 +15,14 @@ import (
...
 func assertNewSignerWithConfOK(t *testing.T, conf signer.Configuration) *GPG2Signer {
-	s, err := New(conf)
+	// These test names end up quite long, and some versions of GPG don't work well
+	// with very long paths. See https://bugzilla.redhat.com/show_bug.cgi?id=1813705
+	// To work around this, we use a hashed version of the test name. This keeps
+	// the prefix consistent across runs while still being short enough.
+	hashedName := md5.Sum([]byte(t.Name()))
+	tmpDirPrefix := fmt.Sprintf("%x", hashedName)
+
+	s, err := New(conf, tmpDirPrefix)
```
jmhodges, line 25: What if we passed in the full temp dir into New, instead? That would me we wouldn't have to glob for anything, we'd handle the edge cases where the base temp dir isn't `/tmp`, we'd know exactly what directory we can delete after the test is over and there wouldn't ever be a collision with a different test. (This is what I thought you were going to do, and may have misspoke and made you think I didn't want that)

jmhodges (changes requested):

signer/gpg2/gpg2_test.go
bhearsum: That it was I was originally thinking of doing, yeah. I backed off of it after I realized we had access to `tmpDir` after the fact, so it didn't seem necessary to do.

jmhodges: Oh, oh, so we actually can use `s.tmpDir` in the `os.RemoveAll`?

bhearsum: Which one are you talking about?

jmhodges: The one at line 371?

jmhodges: Nevermind, I clearly misread!

jmhodges (approved):
Thank you! I misread some uses of `s.tmpDir` as `tmpDir` and got confused.

PR #945, opened 2024-08-09
AUT-186: upgrade to aws-sdk-go-v2

bhearsum:
This turns out to be fairly straightforward? I reworked `uploadToS3` a tiny bit to accommodate basic testing. The easiest way to do this was to move some of the boilerplate one function up. It would be ideal to keep _all_ of it where it currently is, but doing so means we can't dependency inject anything.

Seeing as there's literally no logic other than whether or not errors are forwarded, I'm tempted to drop the tests and move the code back. I could be convinced either way.


signer/contentsignaturepki/upload.go
```diff
@@ -26,21 +34,28 @@ func (s *ContentSigner) upload(data, name string) error {
 	}
 	switch parsedURL.Scheme {
 	case "s3":
-		return uploadToS3(data, name, parsedURL)
+		// Context is a required argument, but in our uses,
+		// LoadDefaultConfig pulls the necessary configuration
+		// from the environment.
+		cfg, err := config.LoadDefaultConfig(context.Background())
```
jmhodges, line 40: Not for this PR and not asking for this to change because we're inheriting prior decisions here, but noting only so that future folks can see our thinking: these `LoadDefaultConfig` and `NewFromConfig`s are typically meant to happen at boot time, and that we do them over and over is a lil suspect. Not a big deal because we only do dozens of these a month, but not quite what's expected

signer/contentsignaturepki/upload.go
```diff
@@ -26,21 +34,28 @@ func (s *ContentSigner) upload(data, name string) error {
...
-func uploadToS3(data, name string, target *url.URL) error {
-	sess := session.Must(session.NewSession())
-	uploader := s3manager.NewUploader(sess)
-	_, err := uploader.Upload(&s3manager.UploadInput{
+func uploadToS3(client S3UploadAPI, data, name string, target *url.URL) error {
+	_, err := client.Upload(context.Background(), &s3.PutObjectInput{
 		Bucket:             aws.String(target.Host),
 		Key:                aws.String(target.Path + name),
-		ACL:                aws.String("public-read"),
+		ACL:                "public-read",
```
jmhodges, line 58: There's a constant we can use for this now: `types.ObjectCannedACLPublicRead` from `"github.com/aws/aws-sdk-go-v2/service/s3/types"`

signer/contentsignaturepki/upload_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"errors"
+	"net/url"
+	"testing"
+
+	"github.com/aws/aws-sdk-go-v2/feature/s3/manager"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+)
+
+// Mocks adapted from https://aws.github.io/aws-sdk-go-v2/docs/unit-testing/
+type mockUploadAPI func(ctx context.Context, input *s3.PutObjectInput, opts ...func(*manager.Uploader)) (*manager.UploadOutput, error)
```
jmhodges, line 14: dope

signer/contentsignaturepki/upload_test.go
```diff
@@ -0,0 +1,76 @@
...
+			data:      "foo",
+			name:      "failed_upload",
+			target:    "https://foo.bar",
+			expectErr: true,
+		},
+	}
+
+	for _, tt := range cases {
+		t.Run(tt.name, func(t *testing.T) {
+			tt := tt
```
jmhodges, line 56: Oh, [Go 1.22 changed how for-loops work](https://go.dev/blog/loopvar-preview) so we no longer need to worry about capturing like this. 

But if we wanted to continue doing that for now as practice for the Go code we rely on that's not yet on Go 1.22, I think you want this redefinition outside of the closure to avoid racing on it (otherwise, the capture has already happened). So move it above the `t.Run` call.

jmhodges (changes requested):

jmhodges:
Stoked at how little had to change here! Thank you

signer/contentsignaturepki/upload.go
bhearsum: Yeah, I'm going to forego doing it as part of this, but seeing as https://mozilla-hub.atlassian.net/browse/AUT-182 will necessitate some changes to configuration, I may end up doing it in advance or as part of that work.

bhearsum:
@jbuck - can you confirm that `AWS_REGION` is set for Autograph? https://github.com/mozilla-services/cloudops-deployment/blob/12c6092e310608a298e3ccae3dc3005e7a5a9083/projects/autograph/puppet/modules/autograph/manifests/app.pp#L41-L43 suggests it is, but I've misread puppet code in the past...

(The new sdk will not function without it set to a concrete region, so this is important before we merge this.)

jmhodges (approved):
If you run `go mod

jmhodges:
@bhearsum I know you didn't ask me, you are correct! (And I've recently had to dig up and use my puppet-knower license)

bhearsum:
> @bhearsum I know you didn't ask me, you are correct! (And I've recently had to dig up and use my puppet-knower license)

Thanks - I think that's confirmation enough to merge this!

PR #951, opened 2024-08-19
consolidate linting and add it to GitHub Actions

jmhodges:
This consolidates our linting tools into the golangci-lint tool. It also
removes the unsupported, archived golint tool.

Docs about golangci-lint can be found here: https://golangci-lint.run/

Specifics about how to configure specific linters and what they are can
be found at https://golangci-lint.run/usage/linters/

The new GitHub Action integration also will annotate the PR its run on
with the violations it fails instead of just printing them out to the
console. This means we'll get nice lil UI things directly next to the
code in a PR that's causing the linter violation. This is the Action
maintained by the golangci-lint maintainers.

The default configuration of `golangci-lint` includes `errcheck` which
is very handy but was going to cause a significant amount of churn here.
I've made AUT-202 for us to turn it on later.

We use `depguard` to replace the `check-no-crypto11-in-signers` target.
It's a lil funky (I don't like the "**/" prefix that's required and
could cause false positives if we ever gain another `signer` package
name) but it's nice to have it in place for future checks like it. Plus,
it won't have the previous kind of false positives of the old grep.

`staticcheck` is still run in golangci-lint by default and we make sure
it remains enabled. No change there.

Along the way, pick up some new lint errors that the previous linting
missed.

One is that the `main.httpError` func is Sprintf-like, and `go vet`'s
`printf` checks were being violated in a few places. Same deal with some
uses of `fmt.Errorf` without any additional arguments. Some `log.Printf`
calls had the wrong number of arguments for the format string they were
given. One of the `main.httpError` calls is false positive-y and gets a
useless `"%s"` because we have no "zero arguments" function equivalent,
but it's fine.

Our immediate update of the `debug` variable in `main_racing_test.go`
meant the new `ineffassign` linter was unhappy. It's slightly false
positive-y because I suspect it was folks just trying to be clear about
what the zero value of a bool is, but the new code for that is shorter
and more typical Go.

`go vet` ("`govet`" in golangci-lint parlance) also had an "unreachable
code" trigger in the autograph-client codebase. A nice simplification
made it go away.

`golangci-lint` got a 5 minute timeout (`--timeout`) in the GitHub
Actions. That time was chosen arbitrarily, and 5 minutes would be a long
time to wait for a linting check to finish. We could definitely shorten
it and maybe should.

Also, the Makefile gains the ability to install `golangci-lint`, because
I don't have a better pattern for how folks should install it right now.

Updates AUT-158
Updates AUT-197

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,40 +8,12 @@ version: 2.1
 jobs:
   lint-vet-fmt:
     docker:
-      - image: cimg/go:1.22
+      # If you adjust this version, also adjust the GOLANGCI_LINT_VERSION in the
+      # Makefile
+      - image: golangci/golangci-lint:v1.60.1-alpine
     steps:
-      # TODO(AUT-158): this linting always succeeds even if there are errors, and future correct linting may need correct cgo compliation
       - checkout
-      - run:
-          name: check crypto11 not used in signers
-          command: |
-            make check-no-crypto11-in-signers
-            make show-lints
-      - run:
-          name: run golint
-          command: |
-            make install-golint lint
-            make show-lints
-            make -C tools/autograph-monitor lint
-            make -C verifier/contentsignature lint
-      - run:
-          name: run gofmt
-          command: |
-            make -s fmt-diff | tee fmt.diff
-            test -z "$(cat fmt.diff)"
-      - run:
-          name: run go vet
-          command: |
-            make vet
-            make -C tools/autograph-monitor vet
-            make -C verifier/contentsignature vet
-      - run:
-          name: run staticcheck
-          command: |
-            make install-staticcheck staticcheck
-            make show-lints
-            make -C tools/autograph-monitor staticcheck
-            make -C verifier/contentsignature staticcheck
+      - run: golangci-lint run
 
   unit-test:
     # based on the official golang image with more docker stuff
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -58,7 +58,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version: stable
+          go-version-file: './go.mod'
       - name: golangci-lint
         # If you bump this version, double check that we shouldn't also bump
         # GOLANGCI_LINT_VERSION in the Makefile. (The version here is for the
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -44,6 +44,30 @@ jobs:
           name: autograph-images-${{ github.sha }}
           path: docker-cache/
 
+  lint:
+    name: Linting
+    runs-on: ubuntu-22.04
+    permissions:
+      # Required: allow read access to the content for analysis.
+      contents: read
+      # Optional: allow read access to pull request. Use with `only-new-issues` option.
+      pull-requests: read
+      # Optional: allow write access to checks to allow the action to annotate code in the PR.
+      checks: write
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version: stable
+      - name: golangci-lint
+        # If you bump this version, double check that we shouldn't also bump
+        # GOLANGCI_LINT_VERSION in the Makefile. (The version here is for the
+        # GitHub Action and the one in the Makefile is for the actual version of
+        # golangci-lint.)
+        uses: golangci/golangci-lint-action@v6.1.0
+        with:
+          args: --timeout 5m
+
   unit-tests:
     name: Run Unit Tests
     runs-on: ubuntu-22.04
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,33 @@
+linters:
+  disable-all: true
+  enable:
+    - depguard
+    # TODO(AUT-202): errcheck is too handy to leave out but requires more churn
+    # at time of writing
+    # - errcheck
+    - gosimple
+    - govet
+    - ineffassign
+    - staticcheck
+    - unused
+
+linters-settings:
+  depguard:
+    rules:
+      # The first reason for this is to minimize how often we're having to deal
+      # with many different kinds of key types when the generic forms of the
+      # keys work fine. The second is that it makes sure our code can handle
+      # keys sourced from differing places.
+      #
+      # refs: https://github.com/mozilla-services/autograph/issues/247
+      no-crypto11-in-signer-submodules:
+        list-mode: lax
+
+        files:
+          - "**/signer/**/*.go"
+
+        deny:
+          - pkg: "github.com/ThalesIgnite/crypto11"
+            desc: "Only indirect PKCS#11 work allowed in signer submodules (so, crypto11 is banned there)"
+          - pkg: "github.com/ThalesGroup/crypto11" # ThalesGroup is the more modern version
+            desc: "Only indirect PKCS#11 work allowed in signer submodules (so, crypto11 is banned there)"
--- a/monitor_handler.go
+++ b/monitor_handler.go
@@ -34,7 +34,6 @@ func (m *monitor) handleMonitor(w http.ResponseWriter, r *http.Request) {
 
 	for _, errstr := range m.sigerrstrs {
 		if errstr != "" {
-			// "%s" is a go vet false positive correction
 			httpError(w, r, http.StatusInternalServerError, "%s", errstr)
 			return
 		}
--- a/monitor_handler.go
+++ b/monitor_handler.go
@@ -34,7 +34,8 @@ func (m *monitor) handleMonitor(w http.ResponseWriter, r *http.Request) {
 
 	for _, errstr := range m.sigerrstrs {
 		if errstr != "" {
-			httpError(w, r, http.StatusInternalServerError, errstr)
+			// "%s" is a go vet false positive correction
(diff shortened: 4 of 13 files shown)


.github/workflows/tests.yaml
```diff
@@ -44,6 +44,30 @@ jobs:
...
+        with:
+          go-version: stable
+      - name: golangci-lint
+        # If you bump this version, double check that we shouldn't also bump
+        # GOLANGCI_LINT_VERSION in the Makefile. (The version here is for the
+        # GitHub Action and the one in the Makefile is for the actual version of
+        # golangci-lint.)
+        uses: golangci/golangci-lint-action@v6.1.0
+        with:
+          args: --timeout 5m
```
hneiva, line 69: nit: do we care if this takes > 5m?

.golangci.yaml
```diff
@@ -0,0 +1,33 @@
+linters:
+  disable-all: true
+  enable:
+    - depguard
+    # TODO(AUT-202): errcheck is too handy to leave out but requires more churn
+    # at time of writing
+    # - errcheck
```
hneiva, line 7: Awesome! thanks for leaving a hint :)

hneiva (approved):
Looks great! Thank you!

.github/workflows/tests.yaml
jmhodges: It likely means something is very wrong with our dev environment or our lint tooling. Either would mean we need to address it. The alternative is this taking a long time and folks just going "guess our build is slow" and not noticing til it's too late.

hneiva: heh fair enough!

monitor_handler.go
```diff
@@ -34,7 +34,8 @@ func (m *monitor) handleMonitor(w http.ResponseWriter, r *http.Request) {
 
 	for _, errstr := range m.sigerrstrs {
 		if errstr != "" {
-			httpError(w, r, http.StatusInternalServerError, errstr)
+			// "%s" is a go vet false positive correction
```
oskirby, line 37: Uhh, I think `go vet` is right in this case. It seems like `httpError()` seems to accept format strings, so if `errstr` contains any percent characters it will expect more args on the call stack.

.github/workflows/tests.yaml
```diff
@@ -44,6 +44,30 @@ jobs:
...
+      contents: read
+      # Optional: allow read access to pull request. Use with `only-new-issues` option.
+      pull-requests: read
+      # Optional: allow write access to checks to allow the action to annotate code in the PR.
+      checks: write
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version: stable
```
oskirby, line 61: Rather than `stable` should we instead use the same version specified in the `go.mod`?

.circleci/config.yml
oskirby, line 1: I have a radical suggestion - let's remove the CircleCI linting job entirely if this also introduces a Github version that does the same job. There is limited value in having the same test run in two places and producing the same warning messages.

oskirby (approved):
I left a couple suggestions - but overall I think this is good.

.circleci/config.yml
hneiva: oh, I like this one!

monitor_handler.go
jmhodges: I mean, technically I suppose? But the errors are known to not contain format strings (see, for instance, [these lines](https://github.com/mozilla-services/autograph/blob/5172bfb5ebfddc160172926a3d52e1ea4f39bae4/monitor.go#L78-L88)), and if they do, this isn't C: the `fmt.Sprintf` fills them in with `EXTRA` strings. See https://pkg.go.dev/fmt#hdr-Format_errors for how Go handles format string errors

The alternative would be to write another httpError that just didn't take extra args at all and that didn't seem worth it.

.github/workflows/tests.yaml
jmhodges: I was hoping to not have another spot where we're updating the Go version across our codebase. The Go that runs the lint tooling being different than what we build the code seemed fine, given the [Go backwards compatibility](https://go.dev/blog/compat) guarantees.

If you're interested in this, I'm down, but it seemed like future noise and something we could lock down later if we truly needed to.

.circleci/config.yml
jmhodges: I was going to post a follow-up PR to remove the CircleCI job after this was merged and after I'd removed the CircleCI linter job from the branch protection configuration.

So, if I add this new Linting check to the branch protection requirements, that means everyone has to wait for this PR to merge before they can get their PRs will pass. Similarly, if I remove the CircleCI lint checks before this is merged in, we lose the linting til then. 

I'll admit I prefer to practice some amount of "do the backwards compatible thing, then do the change after".

.github/workflows/tests.yaml
jmhodges: Like, the violation pops up is actually because of the `Sprintf` call, but it doesn't go away when I do this work conditionally (because Sprintf doesn't actually care much). They're just pattern matching on "string + args". See this patch:

```
diff --git a/errors.go b/errors.go
index 9818b5b1..ba636d0b 100644
--- a/errors.go
+++ b/errors.go
@@ -23,7 +23,13 @@ func httpError(w http.ResponseWriter, r *http.Request, errorCode int, errorMessa
 		"code": errorCode,
 		"rid":  rid,
 	}).Errorf(errorMessage, args...)
-	msg := fmt.Sprintf(errorMessage, args...)
+	var msg string
+	if len(args) == 0 {
+		msg = errorMessage
+	} else {
+		msg = fmt.Sprintf(errorMessage, args...)
+	}
+
 	msg += "\r\nrequest-id: " + rid
 	// when nginx is in front of go, nginx requires that the entire
 	// request body is read before writing a response.
diff --git a/monitor_handler.go b/monitor_handler.go
index da97be8c..745ce17f 100644
--- a/monitor_handler.go
+++ b/monitor_handler.go
@@ -34,8 +34,7 @@ func (m *monitor) handleMonitor(w http.ResponseWriter, r *http.Request) {
 
 	for _, errstr := range m.sigerrstrs {
 		if errstr != "" {
-			// "%s" is a go vet false positive correction
-			httpError(w, r, http.StatusInternalServerError, "%s", errstr)
+			httpError(w, r, http.StatusInternalServerError, errstr)
 			return
 		}
 	}
```

monitor_handler.go
jmhodges: Like, the violation that pops up is actually because of the `Sprintf` call, but it doesn't go away when I do this work conditionally (because Sprintf doesn't actually care much). The linter is just pattern matching on "string + args in function signature". See this patch and how it doesn't make the lint violation go away:

```patch
diff --git a/errors.go b/errors.go
index 9818b5b1..ba636d0b 100644
--- a/errors.go
+++ b/errors.go
@@ -23,7 +23,13 @@ func httpError(w http.ResponseWriter, r *http.Request, errorCode int, errorMessa
 		"code": errorCode,
 		"rid":  rid,
 	}).Errorf(errorMessage, args...)
-	msg := fmt.Sprintf(errorMessage, args...)
+	var msg string
+	if len(args) == 0 {
+		msg = errorMessage
+	} else {
+		msg = fmt.Sprintf(errorMessage, args...)
+	}
+
 	msg += "\r\nrequest-id: " + rid
 	// when nginx is in front of go, nginx requires that the entire
 	// request body is read before writing a response.
diff --git a/monitor_handler.go b/monitor_handler.go
index da97be8c..745ce17f 100644
--- a/monitor_handler.go
+++ b/monitor_handler.go
@@ -34,8 +34,7 @@ func (m *monitor) handleMonitor(w http.ResponseWriter, r *http.Request) {
 
 	for _, errstr := range m.sigerrstrs {
 		if errstr != "" {
-			// "%s" is a go vet false positive correction
-			httpError(w, r, http.StatusInternalServerError, "%s", errstr)
+			httpError(w, r, http.StatusInternalServerError, errstr)
 			return
 		}
 	}
```

jmhodges: I guess maybe you're saying you just want the comment deleted?

.circleci/config.yml
oskirby: I am okay with doing the work in two PRs.

monitor_handler.go
oskirby: Yeah, that's all I was trying to say - just remove the comment and it's fine.

jmhodges: *embarrassed face* can do

.github/workflows/tests.yaml
oskirby: I think the core of my concern is the hypothetical scenario:
1. We develop our code against Go version X and the linter is very happy with our code.
2. Go releases a new version, X+1 as the new stable, and introduces a whole bunch of lints.
3. The golangci linter uses the stable version rather than our `go.mod` when deciding which lints apply.
4. Our CI now goes red even though we changed nothing in our code.

If the linter workflow examines our `go.mod` to determine the applicable lints, then I am fine leaving this PR as-is.

jmhodges: Understood. I believe go.mod is looked at for some of the lints in `go vet` (e.g. the loop varialbe related ones). I'm less worried about breaking the CI cuz it's easy enough to plot down a specific Go version if it turns out we need it and like picking up new, better lints without having to work at it.

But I'm happy to adjust to my reviewer here.

jmhodges: oh jeez, I had forgotten `go-version-file: 'path/to/go.mod'` existed. Just using that is clearly better. This is perhaps what you meant?

PR #955, opened 2024-08-27
aggregate http.api metrics as agg.http.api

jmhodges:
Our observability infrastructure in AWS is bad. We're running a
combination of Grafana and InfluxDB versions that don't allow us to
easily sum over globbed metrics. This makes it difficult to write
availability metrics (a.k.a. succcess rates).

Until we've dropped support for the AWS deployment, we need to add
aggregate metrics across the API request paths to make it trivial to
calculate our availability.

This change allows us to emit "agg.http.api" namespaced metrics that are
roll-ups of all of the other "http.api" request and response metrics.

See #933 for prior work

Updates AUT-150


oskirby (approved):
Looks good to me!

PR #958, opened 2024-08-29
roll up HTTP success and status code group metrics

jmhodges:
Following up on #955, I missed that our observability system would also
make it too hard to roll up the various status codes into sums.

So, we add some per-hundred status code metrics, and a
`response.success` metric that includes both 2xx and 4xx responses. The
4xx count as success because (as the [HTTP docs
say](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status)) 4xx
status codes are client errors. Counting someone's system accidentally
using the wrong credentials shouldn't against our success rate.


hneiva (approved):
LGTM

PR #959, opened 2024-08-29
quick self-scheduling impl

jmhodges:


jmhodges:
Mooted by later work!

PR #961, opened 2024-08-29
bump golang.org/x/sys to 0.24

jmhodges:
Every time I try to use a handy new tool, I end up having to also
upgrade golang.org/x/sys. And that dep is always so much that it swamps
the rest of the change.

So, let's do this explicitly now and reduce the churn on later ones.
x/sys is very stable and reliable, so this is a safe operation.

Doing this also gets me more comfortable with turning dependabot back
on.


hneiva (approved):
LGTM

PR #962, opened 2024-08-29
monitor: use explicit errors and http.DefaultClient

jmhodges:
This changes autograph-monitor's `monitor` func to use explicit return
errors instead of named return values. Since they were introduced, folks
have learned that they most just cause confusing scope questions.

And along the way, we use the `http.DefaultClient` instead of creating a
new one every time we make a request.

This is preliminaries for adding some resiliency and metrics to the
monitor codebase.

Updates AUT-216
Updates AUT-201


hneiva (approved):
LGTM

PR #964, opened 2024-08-29
go back to original status check for matrix jobs

jmhodges:


jmhodges:
Mooted by figuring out we were using the wrong, similarly named check in the branch protection settings

PR #968, opened 2024-08-30
AUT-161 - de-globalizing unit tests. Also removed parallelism

alexcottner:
[AUT-161](https://mozilla-hub.atlassian.net/jira/software/c/projects/AUT/boards/1915?selectedIssue=AUT-161) - de-globalizing unit tests. 
Also removed parallelism to prevent parallel database calls trying to do the same thing.
Still on the fence on adding a test framework that would give us a little more structure.

### DIFF
--- a/authorize_test.go
+++ b/authorize_test.go
@@ -17,7 +17,7 @@ import (
 )
 
 func TestMissingAuthorization(t *testing.T) {
-	ag, _ := MockAutographer(t)
+	ag, _ := newTestAutographer(t)
 
 	body := []byte("aaaaaaaaaaaaaaaaaaaaaaaaaaaa")
 	bodyrdr := bytes.NewReader(body)
@@ -35,7 +35,7 @@ func TestMissingAuthorization(t *testing.T) {
 }
 
 func TestBogusAuthorization(t *testing.T) {
-	ag, _ := MockAutographer(t)
+	ag, _ := newTestAutographer(t)
 
 	body := []byte("aaaaaaaaaaaaaaaaaaaaaaaaaaaa")
 	bodyrdr := bytes.NewReader(body)
@@ -54,7 +54,7 @@ func TestBogusAuthorization(t *testing.T) {
 }
 
 func TestBadPayload(t *testing.T) {
-	ag, conf := MockAutographer(t)
+	ag, conf := newTestAutographer(t)
 
 	body := []byte("aaaaaaaaaaaaaaaaaaaaaaaaaaaa")
 	bodyrdr := bytes.NewReader(body)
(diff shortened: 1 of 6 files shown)


main_test.go
```diff
@@ -15,19 +15,15 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-var (
-	ag   *autographer
-	mo   *monitor
-	conf configuration
-)
+func MockAutographer(t *testing.T) (*autographer, configuration) {
```
alexcottner, line 18: Spent a significant time arguing with myself over:
1. The name of this function
2. Bringing in a test framework that would give us more of a concept of `fixtures` or `beforeEach`

Open to changing course on either honestly.

main_test.go
```diff
@@ -52,26 +48,15 @@ func TestMain(m *testing.M) {
 	} else {
 		ag.hawkMaxTimestampSkew = time.Minute
 	}
-	// ok this is lame but the contentsignaturepki signer will upload a file dated
-	// at the given second so we don't want anything else to run at the same second
-	// otherwise that file may get rewritten. Easiest way to solve it? Sleep.
-	time.Sleep(time.Second)
```
alexcottner, line 58: This seemed to have no effect on tests so removed

authorize_test.go
```diff
@@ -213,11 +217,11 @@ func TestAutographerAddAuthorizationsFails(t *testing.T) {
 			name: "authorization without a signer ID fails",
 			auths: []authorization{
 				{
-					ID:      "alice",
+					ID:      "bernie",
```
alexcottner, line 220: Changed ID so we could use the same `autograper` for each test case rather than recreating it to prevent collisions.

main_test.go
jmhodges: Hm, yeah, seems to not really be a fake nor a mock. Unfortunate that they made this restricted constructor instead of making the real constructor safe for testing.

`testableAutographer`? `newTestAurographer`? Maybe the latter

main_test.go
```diff
@@ -52,26 +48,15 @@ func TestMain(m *testing.M) {
...
-
-	// Initialize a monitor.
-	mo = newMonitor(ag, conf.MonitorInterval)
 
-	// run the tests and exit
-	r := m.Run()
+	t.Cleanup(func() {
+		close(ag.exit)
```
jmhodges, line 53: Do we also need to remove the rows from the database table here, too? I worry we're just filling up everyone's test db with data (even before this change).

jmhodges (changes requested):
Thank you! This is a good patch that is giving us an accurate representation of what we should be testing. 

One question about deleting data in the database and I would like the name change, too. (if you hate my suggestions I'm happy to let go of it).

main_test.go
alexcottner: That's a good idea, adding.

main_test.go
alexcottner: Yeah that sounds better.

jmhodges (approved):

PR #973, opened 2024-09-04
upgrade cloudhsm-cleanup tools to AWS CloudHSM v5

jmhodges:
The CloudHSM v5 tools operate differently, and a bit better.

We chose to lean on the `--filter "attr.label=..."` option during our
firefight to clean up staging.

These tools still won't find HSM slots that were created but not
assigned to a signer in the database. But an artificially constructed
list of those can be given to `delete_keys_by_label.sh`.

We may need to resurrect something like `delete_keys_by_handle.sh` but
we didn't do that in the firefight and are trying to match the tooling
we did end up using.


alexcottner (approved):
Looks reasonable

PR #982, opened 2024-09-06
chore: add google kms library to dependencies

bhearsum:
https://mozilla-hub.atlassian.net/browse/AUT-184 suggests that either using the C library or golang one is acceptable; I can't think of any good reason to prefer the former over the latter.


jmhodges (changes requested):
Couple things: if you fetch and vendor a Go package, but don't reference it in the actual code, it'll be removed on the next run of `go mod tidy` (like in a dependabot run).

Second, we were going to try the PKCS#11 route because folks had wanted to continue using the PKCS#11 library as PKCS#11 is a not-cloud-specific standard and we had the config settings ready to be used. I was inclined to support those efforts, buuuut

These two points synthesize into one point: I would like to see the code actually being used before adding the package. With the PKCS#11/C library version that was already basically in place (we just had to update the autogaph config files to look for the .so files in the right place but the crypto11 library was already in use in autograph), and with the Go change, I'd want to see this new Go code being used.

bhearsum:
I'm fine to simply use the PKCS#11 library if preferred. The way it was phrased in JIRA made it sound like both options were equally good, but perhaps I misread, or didn't have the full context!

jmhodges:
I'm down to take a look at what the go version would look like if you've got an idea!

I def wrote that when I was less convinced that the PKCS#11 library would be worthy. It was originally pitched to me as doing something that wasn't cloud-specific. I got a bit more convinced when I looked at how weird we do the HSM set up inside of autograph proper (the weird inversions of config and signer make refactoring a lil difficult) and then had to think about what it would mean for integration testing.

But I really do hate this C native stuff when the docs for Google's HSM work all talk about the languate-native KMS libraries. So, I'm actually interested if you got the time to pull out what changes we'd need.

jmhodges:
(Because I only thought more about it enough to scare myself not to actually think through how small of a change I could make)

bhearsum:
After some backchanneling with @jmhodges, I'm closing this in favour of the C library approach. Using the Go library will require refactoring that I don't want to put in the critical path of the GCP migration, which I hadn't realized initially.

PR #983, opened 2024-09-11
AUT-184: install Google's libkmsp11 into autograph docker image

bhearsum:
Note that the AWS one that we currently use is mounted into the container at runtime. That will naturally stop happening in GCP, as our infra code is brand new, so no down the road cleanup will be needed.

I've started a companion branch in the config repo to go along with this. It's not complete, but it's a start.


Dockerfile
```diff
@@ -44,6 +44,10 @@ RUN cd /app/src/autograph && go install .
 RUN cd /app/src/autograph/tools/autograph-monitor && go build -o /go/bin/autograph-monitor .
 RUN cd /app/src/autograph/tools/autograph-client && go build -o /go/bin/autograph-client .
 
+# Extract and verify the Google KMS library.
+RUN cd /tmp && curl -L https://github.com/GoogleCloudPlatform/kms-integrations/releases/download/pkcs11-v1.6/libkmsp11-1.6-linux-amd64.tar.gz | tar -zx --strip-components=1
```
jbuck, line 48: Could the version of libkmsp11 be a parameter like the `GO_VERSION`?

Dockerfile
```diff
@@ -62,6 +66,9 @@ ADD autograph.yaml /app
 ADD version.json /app
 COPY --from=builder /go/bin /go/bin/
 
+# Copy Google KMS library from the builder.
+COPY --from=builder /tmp/libkmsp11.so /app
```
jbuck, line 70: What directory does the .so need to be in to be automatically picked up. `/usr/local/lib` maybe?

Dockerfile
bhearsum: Yeah, that's a good call. Let's do that.

Dockerfile
bhearsum: As I understand things, it will never be automatically picked up. We need to explicitly specify the path to it in `hsm.path` in the app config. Via [a](https://github.com/mozilla-services/autograph/blob/057f7ee6bb490bc6e5d5fbdd7536c43d943957e3/main.go#L380) [bit](https://github.com/ThalesGroup/crypto11/blob/c73933259cb60509d00f32306eea53d10f8e8f10/crypto11.go#L239) [of](https://github.com/miekg/pkcs11/blob/9078ad6b9d4b17a548e56959b10e8949afab10ec/pkcs11.go#L789-L791) [indirection](https://github.com/miekg/pkcs11/blob/9078ad6b9d4b17a548e56959b10e8949afab10ec/pkcs11.go#L74-L91) we end up calling `dlopen` and have an indirect handle on the loaded library.

jbuck (approved):

jmhodges (approved):

PR #984, opened 2024-09-17
AUT-252: fork crypto11 into the autograph repository

bhearsum:
This is part 1 of the work for AUT-252, and simply moves `crypto11` from the `vendor` directory into a replacement fork in the root of the Autograph repository. No actual code changes have been made as part of this, but I did:
* Remove `Gopkg.lock` and `Gopkg.yaml`, which are simply unnecessary and possibly confusing.
* Rename the package to `github.com/mozilla-services/autograph/crypto11` to avoid `go mod vendor` copying it into the `vendor` directory, which is annoying and confusing. (Maybe this is the wrong thing to do, in which case I'm happy to revert it and let `go mod vendor` do its thing.)


crypto11/ecdsa.go
```diff
@@ -151,7 +151,7 @@ func marshalEcParams(c elliptic.Curve) ([]byte, error) {
 func unmarshalEcParams(b []byte) (elliptic.Curve, error) {
 	// See if it's a well-known curve
 	for _, ci := range wellKnownCurves {
-		if bytes.Compare(b, ci.oid) == 0 {
+		if bytes.Equal(b, ci.oid) {
```
jmhodges, line 154: This implies to me we're adding a different version of crypto11 than what's in production.

Could we get the same version incorporated? Or should we upgrade the vendored version first?

jmhodges (changes requested):

go.mod
```diff
@@ -88,3 +87,5 @@ require (
 )
 
 replace github.com/mozilla-services/autograph/verifier/contentsignature => ./verifier/contentsignature/
+
+replace github.com/mozilla-services/autograph/crypto11 => ./crypto11/
```
jmhodges, line 91: I don't think we need this line if we're changing the paths everywhere?

crypto11/ecdsa.go
bhearsum: The reason this is getting picked up now is because it's no longer in the `vendor` directory, AIUI. I was trying to avoid any added risk by upgrading, but I can give it a shot if you like.

go.mod
bhearsum: Huh, I guess not? (It looks like we also don't need it for `github.com/mozilla-services/autograph/verifier/contentsignature` in that case, which is what I modeled this after.)

crypto11/ecdsa.go
bhearsum: (Or we can just ignore all the new linting failures and use exactly the same code as we are now.)

jmhodges: I'm sorry? No, I'm saying the code that's in vendor isn't the same that's being added here in the new directory. I'm trying to say you're adding a different version than what's in our vendor directory right now and I would like to either not do that. 

I suspected this was a newer version but it could be this crypto11 is of an older version. In either case, I'd like for us to be exact in what we're doing

go.mod
jmhodges: (I've got a draft up removing that `replace` that I'll mark as ready to review when the tests pass: https://github.com/mozilla-services/autograph/pull/986)

crypto11/ecdsa.go
bhearsum: I'm pretty sure I'm doing this? https://github.com/mozilla-services/autograph/pull/984/commits/5432a9051bd49487dbd54e8372646cc4c63ab1bd moves all of the crypto11 stuff from `vendor` up to the new `crypto11` directory - note the newly failing linting tests on that, despite no actual changes to the code.

https://github.com/mozilla-services/autograph/pull/984/commits/23dfeb96b4cdbf91c8a718b55ddcf6a33c1dd9a1 then makes the minor changes you're seeing to fix the linting issues.

go.mod
bhearsum: Noted - I'll remove this change here.

crypto11/ecdsa.go
jmhodges: Oh, I see. I missed that there were changes because of our linting.

jmhodges (approved):
Toss the `replace` out but otherwise its good

PR #985, opened 2024-09-17
AUT-250 - Getting deploy pipeline for autograph-monitor fixed

alexcottner:
[AUT-250](https://mozilla-hub.atlassian.net/browse/AUT-250)

Getting deploy pipeline for autograph monitor fixed. For our current AWS lambda implementation this means:
1. Build a zip file for the lambda in github actions
2. Switch to amazon linux 2023 runtime for the lambda
3. Adjust jenkins deploy pipeline


.github/workflows/deploy.yaml
```diff
@@ -65,3 +65,36 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
           context: .
+
+  monitor:
```
alexcottner, line 69: This is based on the last time I did this in a node project. Unsure if this is 100% correct for a go project.

.github/workflows/deploy.yaml
```diff
@@ -65,3 +65,36 @@ jobs:
...
+      - name: Build monitor
+        run: |
+          cd tools/autograph-monitor
+          make build
+      
+      - name: Create release draft if one doesn't exist yet
+        run: |
+          EXISTING_RELEASE=$(gh release view $GITHUB_REF_NAME --json="id" || echo '')
+          if [[ -z $EXISTING_RELEASE ]]; then
+              gh release create $GITHUB_REF_NAME --generate-notes -d
```
jmhodges, line 95: If someone does make the git tag by going through GitHub's create release flow, wouldn't this never add the artifact to the release (or perhaps, depending on GH's implementation, race with the release)? 

And, if we fix that, isn't there a possibility of this creating a loop? This would create a draft release of our entire repo which would create a loop if run from someone creating a release with a new tag since these tasks are being run when they're `git tag`'ed, triggering this? (There's currently just extra workflow runs happening, I believe)

I've been trying to think through possible solutions and one of the additional problems is that we're also going to do a monitor and autograph-app staging deploy at the same time with this code and that each have different desires for how tags and GH releases work. What if we had separate tags for each of them?

I'm not sure that the release flow is the right solution for storage after we solve that problem, though. If we used the upload-artifact action, could we get consistent URLs based off our workflow runs that jenkins can use when told what to look for?

jmhodges (changes requested):

.github/workflows/deploy.yaml
```diff
@@ -65,3 +65,36 @@ jobs:
...
+      GH_TOKEN: ${{ github.token }}
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version: '>=1.22'
```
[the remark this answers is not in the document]
alexcottner: Ah yeah, the action can take a `go-version-file` parameter and then we don't need to duplicate it.

.github/workflows/deploy.yaml
alexcottner: > If someone does make the git tag by going through GitHub's create release flow, wouldn't this never add the artifact to the release (or perhaps, depending on GH's implementation, race with the release)?

If somebody makes the git tag by going through GitHub's release flow, this will just append the zip file to that existing release. If an existing release is not found (somebody created a tag without already creating a release), then it will create a draft release so it has something it can attach the zip file to. We can edit those release notes or delete them later before publishing.

I've updated the name of that step to be a little more clear.

> And, if we fix that, isn't there a possibility of this creating a loop?

This should not be able to create a loop in that way. At least I haven't been able to create that behavior in other projects with a similar workflow (ex: [kinto-admin](https://github.com/Kinto/kinto-admin/blob/master/.github/workflows/release.yml)).

> What if we had separate tags for each of them?

We could do that. But with shared code, creating release notes may not be intuitive to know which patches apply to which service. That may take some more thinking. Should we try to diagram this out?

> If we used the upload-artifact action, could we get consistent URLs based off our workflow runs that jenkins can use when told what to look for?

Workflow artifacts are a bit trickier than release files. There can only be one release with that name to pull data from. But for a given tag, you could have N workflows with Y artifacts to pull. We could pass that info over to jenkins in a webhook from the github action, but that doesn't work if we're running the job manually in Jenkins. We'd need to create some logic to identify which job we cared about more than one matches our criteria, and allow it to be overridden by passing in a job id.

Is this something we want to explore?

tools/autograph-monitor/constants.go
```diff
@@ -140,6 +140,49 @@ RIpKO53Cd7aV5PdB0pNSPNjc6V+5IPrepALNQhKIpzoHA4oG+LlVVy4R3csPcj4e
 zQQ9gt3NC2OXF4hveHfKZdCnb+BBl4S71QMYYCCTe+EDCsIGuyXWD/K2hfLD8TPW
 thPX5WNsS8bwno2ccqncVLQ4PZxOIB83DFBFmAvTuBiAYWq874rneTXqInHyeCq+
 819l9s72pDsFaGevmm0Us9bYuufTS5U=
+-----END CERTIFICATE-----`,
+	// previous content signing root cert
```
alexcottner, line 144: after reviewing the logs, looks like we were missing this root for stage checks

.github/workflows/deploy.yaml
jmhodges: I def misread these two steps.

I'm happy to ship this as-is and we can deal with the outcomes if they suck when they suck.

jmhodges (approved):

.github/workflows/deploy.yaml
jmhodges: I def misread these two steps.

I'm happy to ship this as-is and we can deal with the outcomes if they suck when they suck.

PR #987, opened 2024-09-17
AUT-252: add new public functions to crypto11 that allow us to generate keys with libkmsp11

bhearsum:
libkmsp11 does not accept any attributes on public keys, and very limited attributes on private keys. To accommodate this, I'm adding new public functions to crypto11 that allow us to fully specify the attributes for a new key creation request. The default attributes have been kept for the current functions to allow us to preserve backwards compatible for AWS. This means that we'll end up making different calls in MakeKey depending on whether we're in GCP or AWS, but I think it's preferable to removing the default attributes altogether, which we already know work fine for AWS.

The new functions have been tested in https://github.com/bhearsum/crypto11-test-gcp/commit/1094dddba6cc347eecd9c90ff82a149b39110b4c (technically we could do some testing directly here or in Autograph if we injected the right things to avoid actually calling into libkmsp11...but I haven't been able to convince myself that it's worth spending the time doing it).

A possible alternative here would be to add a separate method in crypto11 that uses the correct values for GCP, but I'm not sure that level of hardcoding is a great idea.


jmhodges (changes requested):
Tests? Maybe we use these immediately so that the integration tests catch them?

bhearsum:
> Tests? Maybe we use these immediately so that the integration tests catch them?

I'm not opposed to it, but can I'd like to clarify what you're looking for coverage on. AFAIK, the only coverage we have of the key generation parts of crypto11 is through [the calls we make to it in MakeKey](https://github.com/mozilla-services/autograph/blob/6953e65ab9312b5eaddf7ab81985d02f14ef3a47/signer/signer.go#L433) that happen at startup during integration tests. I believe these all run through SoftHSM. That doesn't work if we pass it the attributes that GCP wants:
```autograph-app-hsm  | {"Timestamp":1726669108250740929,"Time":"2024-09-18T14:18:28Z","Type":"app.log","Logger":"autograph","Hostname":"9622338d258a","EnvVersion":"2.0","Pid":1,"Severity":2,"Fields":{"msg":"failed to add signer \"normandy\": contentsignaturepki \"normandy\": failed to initialize end-entity: contentsignaturepki \"normandy\": failed to generate end entity: failed to generate ecdsa key in hsm: pkcs11: 0xD0: CKR_TEMPLATE_INCOMPLETE"}}```

...which means we'd end up with exactly the same coverage as we have now in integration tests AFAICT.

The other type of testing that I can think to do is write some actual tests for crypto11, and dependency inject the `session` away to avoid calling into libkmsp11. Now that we've forked it there's certainly some sense in doing that.

Of course I may also be overlooking some other type of testing that you're getting at here.

File: main.go [hunk 1]
```diff
@@ -385,6 +385,12 @@ func (a *autographer) initHSM(conf configuration) error {
 		return fmt.Errorf("error in initHSM from crypto11.Configure: %w", err)
 	}
 	if tmpCtx != nil {
+		var hsm signer.HSM
+		if os.Getenv("KMS_PKCS11_CONFIG") != "" {
+			hsm = signer.NewGCPHSM(tmpCtx)
```
jmhodges, line 390: How does this interact with our integration testing? Could we get GCP versions of them?

bhearsum: I'll look into this a bit. At the very least we should make sure we have the same coverage we currently have for AWS, but it looks to me like all of the integration tests run through [app-hsm](https://github.com/mozilla-services/autograph/blob/30d626b459b0bb407c71aa32decf2ebd15f6a055/docker-compose.yml#L36-L59), which [uses the softhsm](https://github.com/mozilla-services/autograph/blob/30d626b459b0bb407c71aa32decf2ebd15f6a055/tools/softhsm/autograph.softhsm.yaml#L10-L15). So after #993 we might have a little bit of coverage that runs through `AWSHSM`, but obviously it doesn't end up using the AWS pkcs11 library.

bhearsum:
Moving back to draft state while I iterate on integration tests.

jmhodges:
> > Tests? Maybe we use these immediately so that the integration tests catch them?
> 
> 
> 
> I'm not opposed to it, but can I'd like to clarify what you're looking for coverage on. AFAIK, the only coverage we have of the key generation parts of crypto11 is through [the calls we make to it in MakeKey](https://github.com/mozilla-services/autograph/blob/6953e65ab9312b5eaddf7ab81985d02f14ef3a47/signer/signer.go#L433) that happen at startup during integration tests. I believe these all run through SoftHSM. That doesn't work if we pass it the attributes that GCP wants:
> 
> ```autograph-app-hsm  | {"Timestamp":1726669108250740929,"Time":"2024-09-18T14:18:28Z","Type":"app.log","Logger":"autograph","Hostname":"9622338d258a","EnvVersion":"2.0","Pid":1,"Severity":2,"Fields":{"msg":"failed to add signer \"normandy\": contentsignaturepki \"normandy\": failed to initialize end-entity: contentsignaturepki \"normandy\": failed to generate end entity: failed to generate ecdsa key in hsm: pkcs11: 0xD0: CKR_TEMPLATE_INCOMPLETE"}}```
> 
> 
> 
> ...which means we'd end up with exactly the same coverage as we have now in integration tests AFAICT.
> 
> 
> 
> The other type of testing that I can think to do is write some actual tests for crypto11, and dependency inject the `session` away to avoid calling into libkmsp11. Now that we've forked it there's certainly some sense in doing that.
> 
> 
> 
> Of course I may also be overlooking some other type of testing that you're getting at here.

I was not aware that this wouldn't work with softhsm and had naively expected some way for that to work! That's frustrating!

I'm def willing to hear we have to do something hacky here. We can also put this on the board for a future rework with the Go native libs since we're apparently losing the softhsm based integration testing anyway!

jmhodges:
> > > Tests? Maybe we use these immediately so that the integration tests catch them?
> > 
> > 
> > I'm not opposed to it, but can I'd like to clarify what you're looking for coverage on. AFAIK, the only coverage we have of the key generation parts of crypto11 is through [the calls we make to it in MakeKey](https://github.com/mozilla-services/autograph/blob/6953e65ab9312b5eaddf7ab81985d02f14ef3a47/signer/signer.go#L433) that happen at startup during integration tests. I believe these all run through SoftHSM. That doesn't work if we pass it the attributes that GCP wants:
> > `autograph-app-hsm  | {"Timestamp":1726669108250740929,"Time":"2024-09-18T14:18:28Z","Type":"app.log","Logger":"autograph","Hostname":"9622338d258a","EnvVersion":"2.0","Pid":1,"Severity":2,"Fields":{"msg":"failed to add signer \"normandy\": contentsignaturepki \"normandy\": failed to initialize end-entity: contentsignaturepki \"normandy\": failed to generate end entity: failed to generate ecdsa key in hsm: pkcs11: 0xD0: CKR_TEMPLATE_INCOMPLETE"}}`
> > ...which means we'd end up with exactly the same coverage as we have now in integration tests AFAICT.
> > The other type of testing that I can think to do is write some actual tests for crypto11, and dependency inject the `session` away to avoid calling into libkmsp11. Now that we've forked it there's certainly some sense in doing that.
> > Of course I may also be overlooking some other type of testing that you're getting at here.
> 
> I was not aware that this wouldn't work with softhsm and had naively expected some way for that to work! That's frustrating!
> 
> I'm def willing to hear we have to do something hacky here. We can also put this on the board for a future rework with the Go native libs since we're apparently losing the softhsm based integration testing anyway!

In an offline Slack discussion, we've since realized there is not way to make our softhsm based integration testing work the GCP HSM library libkmsp11. 

We've invested a lot of effort into this lib, but a future change that uses the Go native library for GCP KMS will probably be a good idea. But we'd like to start testing the use of the GCP HSM-backed keys and folks are out of town, so we're going to try this out as-is.

jmhodges (approved):
In an offline Slack discussion, we've since realized there is no way to make our softhsm based integration testing work the GCP HSM library libkmsp11.

We've invested a lot of effort into this lib, and a future change that uses the Go native library for GCP KMS will probably be a good idea. But we'd like to start testing the use of the GCP HSM-backed keys and folks are out of town, so we're going to try this out as-is.

PR #988, opened 2024-09-18
AUT-250 - Missed that the deploy step also ran on merges to main. Moving monitor zip build to seperate workflow

alexcottner:
On the previous PR, I overlooked that the deploy workflow also runs on merges to main. 
We only want the monitor lambda build to happen on tags.
Moving to a separate workflow.

### DIFF
--- a/.github/workflows/build-monitor.yaml
+++ b/.github/workflows/build-monitor.yaml
@@ -32,7 +32,7 @@ jobs:
         run: |
           EXISTING_RELEASE=$(gh release view $GITHUB_REF_NAME --json="id" || echo '')
           if [[ -z $EXISTING_RELEASE ]]; then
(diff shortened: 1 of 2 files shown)


jmhodges (commented):
Real quick while we're here: the patch to cloudops looked at `latest`, but we only create draft releases here. Do draft releases show up as `latest`?

alexcottner:
> Real quick while we're here: the patch to cloudops looked at `latest`, but we only create draft releases here. Do draft releases show up as `latest`?

Draft releases will not count as the `latest` release. The most recent release may not count as the `latest` release either. Whenever a release is published (or a published release is updated) with "Set as the latest release" checked, then that link will be setup.

jmhodges:
So, then, why do we have this making a draft release? I'm not sure what the point of it is.

jmhodges (approved):

PR #989, opened 2024-09-18
AUT-254: install git to fix coverage report generation

bhearsum:
Coveralls needs git installed to properly report coverage information.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -91,6 +91,7 @@ jobs:
       - name: Run Tests
         shell: bash
         run: |
+          /usr/bin/git show -s --format=%cN HEAD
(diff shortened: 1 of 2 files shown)


bhearsum:
OK, so even with git installed we end up with:
```
2024-09-18T18:38:11.4079617Z 18:38:11 gitinfo.go:85: fail to run "/usr/bin/git show -s --format=%aE HEAD": exit status 128
2024-09-18T18:38:11.4081134Z 18:38:11 gitinfo.go:86: git information is omitted
```

After some poking, it looks like this is because the tests run in Docker, where there is no git repo. I'm giving up on this for now, because having full git metadata doesn't block having coverage reported (just having the most possible information).

PR #990, opened 2024-09-19
Adjustments to android-store scripts

alexcottner:
Work in progress...
Adjustments/fixes for the android-store key generation scripts.

### DIFF
--- a/tools/android-store/Makefile
+++ b/tools/android-store/Makefile
@@ -29,9 +29,9 @@ PRIVATE_KEY ?= ../$(APP_NAME)-private-key.pem
 PUBLIC_CERT ?= ../$(APP_NAME)-signing-cert.pem
 WRAP_KEY ?= ../encryption_public_key.pem
 
-# Get the current user's name and group name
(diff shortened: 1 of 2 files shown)


tools/android-store/Makefile
```diff
@@ -30,6 +29,10 @@ PRIVATE_KEY ?= ../$(APP_NAME)-private-key.pem
 PUBLIC_CERT ?= ../$(APP_NAME)-signing-cert.pem
 WRAP_KEY ?= ../encryption_public_key.pem
 
+# The user and group id for the app user
+DOCKER_USER_ID = 10001
```
alexcottner, line 33: Magic numbers from our [Dockerfile](https://github.com/mozilla-services/autograph/blob/main/Dockerfile#L75)

jmhodges: Oh, I meant write this comment in the file itself.

jmhodges (changes requested):
Drop a comment in the file itself about where the id numbers came from

jmhodges (approved):

PR #991, opened 2024-09-19
Improving autograph monitor log messages

alexcottner:
Fixing our log handling for autograph monitor.

- Upgrading AWS lambda package to fix our error status reporting
- Making some monitor logs and messages more explicit

This is definitely a "light touch" PR, just trying to just get some better error reporting. 
If deeper cuts are desired now, we can make them. But I figured this is a good first step to let us know where things are failing.


jmhodges (approved):

tools/autograph-monitor/monitor.go
```diff
@@ -94,8 +94,7 @@ func main() {
 	} else {
```
jmhodges, line 94: Oh, this isn't the code that's being run, right? Should we add a wrapping around `Handler` that logs the errors it returns to help us along?

jmhodges: Oh, I see your comment in Slack saying that the error logging is fixed by the upgrade!

jmhodges (approved):

PR #992, opened 2024-09-20
feat: remove GetPrivKeyHandle

bhearsum:
This function is used, but the only use in production code is in `contentsignaturepki/contentsignature.go`. The only thing we do with it _there_ is insert it into the database. The only time it is consumed from the database is to be printed out as general information as part of cleanup script. That is to say: it is unused for all practical purposes.

This change removes GetPrivKeyHandle and its uses, but does not go as far as removing the database column itself. This is in part because a schema change is less trivial than code removal, but also to guard against the remote chance that I'm misinterpreting something, and we need to roll this back. We can remove the column itself at some future date when we're absolutely certain it is unused.

### DIFF
--- a/database/queries.go
+++ b/database/queries.go
@@ -66,9 +66,11 @@ func (db *Handler) GetLabelOfLatestEE(signerID string, youngerThan time.Duration
 }
 
 // InsertEE uses an existing transaction to insert an end-entity in database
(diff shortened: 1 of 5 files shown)


jmhodges (approved):
Stealing this review from naomi but I've got context on this one

PR #993, opened 2024-09-20
refactor: isolate HSM interactions and crypto11 use in a more generic interface

bhearsum:
The main driver behind this is to enable us to add a `GCPHSM` with its own MakeKey implementation to avoid cluttering up `Configuration.MakeKey` further. While I was doing this, I moved the other HSM and `crypto11`` related things elsewhere as well. Notably, two things are still present in `signer.Configuration` that would be part of a `signer.HSM` implementation in an ideal world:
* The pub key casting in `Configuration.GetKeys`
* `Configuration.CheckHSMConnection` entirely
    
Both of these things depend on `Configuration.GetPrivateKey` which needs a `label` of a key to fetch. I don't think it's appropriate to move that label to the `HSM` interface (because it would tie an instance to a single key). I couldn't find a way to adjust the usage of these methods that didn't end up making things messier, or involved refactoring other things, so I'm punting on it for now.

### DIFF
--- a/signer/hsm.go
+++ b/signer/hsm.go
@@ -1,0 +1,75 @@
+package signer
+
+import (
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/rsa"
+	"fmt"
+	"io"
+
+	"github.com/miekg/pkcs11"
+	"github.com/mozilla-services/autograph/crypto11"
+)
+
+type HSM interface {
+	GetPrivateKey(label []byte) (crypto.PrivateKey, error)
+	MakeKey(keyTpl interface{}, keyName string) (crypto.PrivateKey, crypto.PublicKey, error)
+	GetRand() io.Reader
+}
+
+type GenericHSM struct {
+	ctx *pkcs11.Ctx
+}
+
+// GetPrivateKey locates the keypair given by `label` in the HSM.
+func (hsm *GenericHSM) GetPrivateKey(label []byte) (crypto.PrivateKey, error) {
+	return crypto11.FindKeyPair(nil, label)
+}
+
+func (hsm *GenericHSM) GetRand() io.Reader {
+	return new(crypto11.PKCS11RandReader)
+}
+
+type AWSHSM struct {
+	GenericHSM
+}
+
+// MakeKey generates a new keypair of type `keyTpl` and returns the new key structs.
+func (hsm *AWSHSM) MakeKey(keyTpl interface{}, keyName string) (crypto.PrivateKey, crypto.PublicKey, error) {
+	var slots []uint
+	slots, err := hsm.ctx.GetSlotList(true)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to list PKCS#11 Slots: %w", err)
+	}
+	if len(slots) < 1 {
+		return nil, nil, fmt.Errorf("failed to find a usable slot in hsm context")
+	}
+	keyNameBytes := []byte(keyName)
+	switch keyTplType := keyTpl.(type) {
+	case *ecdsa.PublicKey:
+		priv, err := crypto11.GenerateECDSAKeyPairOnSlot(slots[0], keyNameBytes, keyNameBytes, keyTplType)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to generate ecdsa key in hsm: %w", err)
+		}
+		pub := priv.PubKey.(*ecdsa.PublicKey)
+		return priv, pub, nil
+	case *rsa.PublicKey:
+		keySize := keyTplType.Size()
+		priv, err := crypto11.GenerateRSAKeyPairOnSlot(slots[0], keyNameBytes, keyNameBytes, keySize)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to generate rsa key in hsm: %w", err)
+		}
+		pub := priv.PubKey.(*rsa.PublicKey)
+		return priv, pub, nil
+	default:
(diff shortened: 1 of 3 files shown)


signer/hsm.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/mozilla-services/autograph/crypto11"
+)
+
+type HSM interface {
+	GetPrivateKey(label []byte) (crypto.PrivateKey, error)
+	MakeKey(keyTpl interface{}, keyName string) (crypto.PrivateKey, crypto.PublicKey, error)
+	GetRand() io.Reader
+}
+
+type GenericHSM struct {
```
bhearsum, line 20: I will note that there's an argument to be made that this shouldn't exist, and the trivial implementations of `GetPrivateKey` and `GetRand` should just stay in `signer.go`. My opinion is that if we're striving to _fully_ isolate (ie: move all `crypto11` interactions) here that we should keep this. If we don't care about that, we can drop this and simply have `AWSHSM` (and later `GCPHSM`) implement `MakeKey`.

signer/signer.go
```diff
@@ -130,13 +130,17 @@ type Configuration struct {
...
-	hsmCtx         *pkcs11.Ctx
+	Hsm HSM
 }
 
 // InitHSM indicates that an HSM has been initialized
 func (cfg *Configuration) InitHSM(ctx *pkcs11.Ctx) {
 	cfg.isHsmAvailable = true
-	cfg.hsmCtx = ctx
+	cfg.Hsm = &AWSHSM{
+		GenericHSM{
```
bhearsum, line 140: Needing an explicit `GenericHSM` looks weird to me, and feels like it ought to be unnecessary. My go-fu wasn't strong enough to figure out how to simply set `ctx` directly in `AWSHSM`, however.

bhearsum:
@jmhodges (or anyone else) - if you have some ideas about how we could test this usefully I'd _love_ to hear them. The only way I could come up with is to dependency inject `crypto11` into `MakeKey` such that we can provide a fake version of, eg: `GenerateECDSAKeyPairOnSlot`. This would end up trickling up the call stack to `Configuration.MakeKey` as well, which is a bit annoying. (And also we'd have to make a fake version of `pkcs11.Ctx`, but I think that is quite trivial to create and inject.) My instincts for testing all come from Python, so I may be missing some other options available to us here.

signer/signer.go
jmhodges: Ah, yeah, this is def Go preferring explicit over implicit. The recommendation would be to make a newAWSHSM(ctx) func to hide the details. Honestly, I think my feedback would be to do so, anyway.

signer/signer.go
```diff
@@ -130,13 +130,17 @@ type Configuration struct {
...
 	isHsmAvailable bool
-	hsmCtx         *pkcs11.Ctx
+	Hsm HSM
 }
 
 // InitHSM indicates that an HSM has been initialized
 func (cfg *Configuration) InitHSM(ctx *pkcs11.Ctx) {
 	cfg.isHsmAvailable = true
-	cfg.hsmCtx = ctx
+	cfg.Hsm = &AWSHSM{
```
jmhodges, line 139: I like how you've avoided the siren song I could not: moving these methods out of Configuration entirely. This is probably wiser.

This is instantiation is pretty deep in here. What would this and its calling code look like when we added the GCP version?

Could we maybe have `InitHSM` take the interface type instead of `*pkcs11.Ctx` and do whatever env var or autograph config file checking we need to in `main.go`?

I have two strands of feedback here in one line of code: We should probably have a version that works in GCP in this patch so that we can know we've got the right joints exposed, and we probably want the switching between the versions of these types to happen at boot time in `main.go` somewhere.

jmhodges (changes requested):

signer/signer.go
jmhodges: Also, thanks for your patience on the feedback. I was mulling this over hours ago and got distracted helping with some other problems.

signer/hsm.go
jmhodges: Oh, in another part of the code, my feedback included wanting to use this interface type as an argument! I think it's worthy.

signer/signer.go
jmhodges: (Also, that "do the choosing between AWSHSM and GCPHSM in main.go" will def push me into the camp of "make constructor NewAWSHSM and NewGCPHSM funcs to hide their GenericHSM wrapping details"

signer/signer.go
bhearsum: Ah; yes, I've seen that pattern before come to think of it. Will do.

signer/signer.go
bhearsum: > Could we maybe have InitHSM take the HSM interface type instead of *pkcs11.Ctx and do whatever env var or autograph config file checking we need to in main.go?

Yeah, that sounds like a great idea.

> I have two strands of feedback here in one line of code: We should probably have a version that works in GCP in this patch so that we can know we've got the right joints exposed

We're getting to the point where we have recursive dependencies between #987 and this :stuck_out_tongue:. Part of the reason I did this refactoring now was so that I could add the GCP MakeKey implementation as part of that PR. It sounds like I should just roll that, plus this, plus the `GCPHSM` implementation into one PR for the (good) testing reasons you mentioned?

> we probably want the switching between the versions of these types to happen at boot time in main.go somewhere.

Will do.

jmhodges: Nah, you're right. I'm being a bother. We can leave these separate

jmhodges (approved):

PR #1000, opened 2024-09-20
correct CheckHSMConnection error handling

jmhodges:
With GetPrivKeyHandle removed in #992, we forgot to update the error
handling in CheckHSMConnection causing it to always return an error. Now
we rely soely on the earlier PrivateKeyHasPEMPrefix call for whether the
key is in the HSM.

Along the way, we cause main to error if the HSM connection check fails.
This allows autograph and the integration tests to fail immediately if
the HSM is believed to be unavailable.

We found this problem at staging deploy time because we weren't checking
the HSM connection on boot. The deploy was failing because the HSM
healthcheck failed, but those aren't checked in integration tests.

### DIFF
--- a/main.go
+++ b/main.go
@@ -164,7 +164,10 @@ func run(conf configuration, listen string, debug bool) {
 
 	// initialize the hsm if a configuration is defined
 	if conf.HSM.Path != "" {
-		ag.initHSM(conf)
+		err = ag.initHSM(conf)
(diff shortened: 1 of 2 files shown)


alexcottner (approved):

jmhodges:
FYI, we tested that the new error handling fails the integration tests in PR #1001

PR #1001, opened 2024-09-20
no correct getprivhandle

jmhodges:
- **correct CheckHSMConnection error handling**
- **put the bad change back to test failures**


jmhodges:
Thank you for doing your job, draft PR. You've told us the error handling we added did successfully fail the integration tests with the old error.

PR #1006, opened 2024-09-25
AUT-201 - make autograph monitor work in GCP

alexcottner:
[AUT-201](https://mozilla-hub.atlassian.net/browse/AUT-201)
Doing work to get Monitor working as expected in GCP.

So far this means:

1. Adjusting docker compose to work with a non-lambda monitor
2. Adjusting integration tests
3. Removing the word `lamda` from some places


tools/autograph-monitor/selftest-entrypoint.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -o pipefail
+
+docker compose run monitor
```
jmhodges, line 5: I don't follow how this works. This seems to be running docker compose from within a dockerfile?

... are the CI integration tests actually running this code?

alexcottner: Hah! Copy paste issue. I meant to grab `/go/bin/autograph-monitor`. Although I suppose that could work with a `dind` image.

jmhodges: Why didn't the integration tests fail on this PR? I don't think this code is being run in CI and we would like it to, no?

jmhodges (changes requested):
Just to make sure we've covered I'm adding this as a request for changes:

Why did the integration tests not fail with the typo in place? It feels like maybe this code isn't being run. (I'm happy to be wrong and just not understanding a subtlety in the entrypoint fix but I'd like to know what that is)

tools/autograph-monitor/selftest-entrypoint.sh
alexcottner: It used to override the command to do `/go/bin/autograph-monitor`. I've adjusted the docker-compose and workflow to use the default command, which is now `selftest-entrypoint.sh`.

alexcottner: Ah, because the docker-compose.yaml used to have `command: ["/go/bin/autograph-monitor"]` and it was bypassing that script. But I wanted that script to make things a little more explicit on an error exit. I've removed that and now we're using the default command from the dockerfile (which runs this script).

jmhodges: I don't follow. I don't see any changes in the "Fixing entrypoint bug" commit that looks like that and that was the one that was fixing the typo.

jmhodges (approved):

PR #1007, opened 2024-09-25
add tests for the new hsm interface

bhearsum:
This is sortof a follow up to #993 in that it adds some rudimentary tests for the new `AWSHSM` struct (as well as a bunch of scaffolding to make it happen). This will allow me to add similar tests in #987 where I'll be adding a `GCPHSM` struct.


crypto11/crypto11.go
```diff
@@ -217,7 +253,7 @@ type PKCS11Config struct {
 // If config is nil, and the library has already been configured, the
 // context from the first configuration is returned (and
 // the error will be nil in this case).
-func Configure(config *PKCS11Config) (*pkcs11.Ctx, error) {
+func Configure(config *PKCS11Config, ctx PKCS11Context) (PKCS11Context, error) {
```
bhearsum, line 256: @jmhodges - you mentioned that it's not ideal that we need to pass `nil` here in a few places. I don't think there's a good way to avoid adding this argument altogether (unless we want all the callers to need to create their own `pkcs11.Ctx` objects, which seems a bit onerous). I think the alternative you proposed was a wrapper that would only require the `PKCS11Config` that we would use from existing callers instead?

signer/signer_test.go
```diff
@@ -303,3 +310,123 @@ var rejectedFileNames = []string{
...
+	mockCtx.EXPECT().GetSlotList(gomock.Any()).Return([]uint{0}, nil).AnyTimes()
+	mockCtx.EXPECT().GetTokenInfo(gomock.Any()).Return(pkcs11.TokenInfo{}, nil).AnyTimes()
+	mockCtx.EXPECT().OpenSession(gomock.Any(), gomock.Any()).Return(pkcs11.SessionHandle(0), nil).AnyTimes()
+	mockCtx.EXPECT().GenerateKeyPair(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
+	mockCtx.EXPECT().GetAttributeValue(gomock.Any(), gomock.Any(), gomock.Any()).Return(pubKeyAttrs, nil).AnyTimes()
+	// these ones are called as part of Close(), not as part of our actual testing
+	mockCtx.EXPECT().CloseSession(gomock.Any()).AnyTimes()
+	mockCtx.EXPECT().CloseAllSessions(gomock.Any()).AnyTimes()
+	mockCtx.EXPECT().Finalize().AnyTimes()
+	mockCtx.EXPECT().Destroy().AnyTimes()
```
bhearsum, line 373: These expectations, and all of the above that goes into them are not great. I'm open to alternatives.

crypto11/crypto11.go
jmhodges: Nil interface checks don't work exactly how you'd like them to work and so we avoid doing them: https://trstringer.com/go-nil-interface-and-interface-with-nil-concrete-value/ .

It seems like we're just trying to allow for a fake PKCS11Context to be injected. I would either:

Add a `ConfigureWithPKCS11Context(config, ctx)` function where the ctx is guaranteed by the caller to be non-nil. This is def the "always create their own Ctx objects" solution.

Or add a factory func type to the signature `Configure(config *PKCS11Config, factory func(*PKCS11Config) PKCS11Context)` with a known good Default factory func defined in the package that we can pass it that does the `pkcs11.New(config.Path)`.

It would be easier to see what was better if I knew what the GCPHSM code would look like. I suspect the factory one

signer/signer_test.go
jmhodges: Why are all of these AnyTimes instead of exact? Should the GenerateKeyPair return the key we're creating in this test? A lot of these methods seem to take `gomock.Any()` that we have control over (e.g. OpenSession returns a specific SessionHandle that we would expect it to be sent to CloseSession, no?)

crypto11/crypto11.go
jmhodges: Ah, let me take back that "would be easier" line. I understand you're actually tyring to get the existing MakeKey under test. Sorry for the noise

jmhodges: (I still think the factory one is probably preferable)

signer/signer_test.go
bhearsum: Honestly - I made these quite liberal simply because `pkcs11` isn't under test. I can update them, it just seemed unnecessary to verify that `pkcs11` and/or `crypto11` is behaving well internally when `MakeKey` is what's under test.

jmhodges: Well, you need them to verify that crypto11 is calling the right parts of pkcs11 in the right order, right?

jmhodges: Otherwise, why are we mocking out pkcs11 isntead of crypto11?

bhearsum: I would prefer to mock out crypto11! But AIUI that would require a deeper refactoring to wrap all of the `crypto11` uses in some interface we can then dependency inject.

I don't think there's a any harm in improving this test as things stand now though.

crypto11/crypto11.go
bhearsum: Let me know what you think about the update I pushed - I think it's what you were aiming at?

signer/signer_test.go
bhearsum: I updated the mock expectations.

signer/signer_test.go
```diff
@@ -303,3 +310,190 @@ var rejectedFileNames = []string{
 	"control*chars",
 	"_non_alpha_start",
 }
+
+func MockedPKCS11ContextFactory(ctx crypto11.PKCS11Context) crypto11.PKCS11ContextFactory {
```
jmhodges, line 314: this could be lower case. When I saw it elsewhere, I went looking to see where this was being defined and then used outside of this package and these tests.

jmhodges (approved):
I think we've got a bad habit of making public methods that aren't actually used outside of their package, so it'd be nice to get that one fixed.

jmhodges:
Also, this is great! Thanks for doing this work

PR #1012, opened 2024-09-30
monitor: be more lenient and more strict about AUTOGRAPH_URL

jmhodges:
Before this patch, AUTOGRAPH_URL was required to include a trailing
slash in order to be considered valid but wasn't required to be an
actual URL.

We use url.JoinPath so that if we forget the trailing slash in the
`AUTOGRAPH_URL` env var for the autograph-monitor, things still work
correctly.

Before this, the autograph-monitor was trying to resolve
`"${AUTOGRAPH_URL}__monitor__"` as a domain name when you forgot the
trailing slash.

Along the way, we refactor and test our URL parsing of `AUTOGRAPH_URL`.

We also handle some `err` shadowing in `main` that was getting
complicated by renaming one error variable to `rootErr` and relying on
`errors.Join`'s behavior of returning `nil` if all its error arguments
are `nil`.

### DIFF
--- a/tools/autograph-monitor/monitor.go
+++ b/tools/autograph-monitor/monitor.go
@@ -116,7 +116,7 @@ func rawAutographURLToMonitorEndpoint(autographURLEnvVar string) (string, error)
 	if err != nil {
 		return "", fmt.Errorf("failed to parse AUTOGRAPH_URL as url: %s", err)
 	}
-	if baseURL.Scheme != "https" {
+	if baseURL.Scheme != "https" && baseURL.Scheme != "http" {
 		return "", fmt.Errorf("AUTOGRAPH_URL %#v must be an https:// (or http:// url in testing)", autographURLEnvVar)
 	}
 	if baseURL.Host == "" {
--- a/tools/autograph-monitor/monitor.go
+++ b/tools/autograph-monitor/monitor.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"log"
 	"net/http"
+	"net/url"
 	"os"
 	"runtime"
 	"strings"
@@ -28,11 +29,12 @@ import (
 )
 
 type configuration struct {
-	url           string
-	monitoringKey string
-	env           string
-	rootHashes    []string
-	truststore    *x509.CertPool
+	origAutographURL string
+	requestURL       string
+	monitoringKey    string
+	env              string
+	rootHashes       []string
+	truststore       *x509.CertPool
 
 	// hashes and keystore for verifying XPI dep signers
 	depRootHashes []string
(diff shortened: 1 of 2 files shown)


bhearsum (approved):
Is this worth adding tests for? (I see we have some minimal amount of testing for some parts of `autograph-client` already.)

jmhodges:
> Is this worth adding tests for? (I see we have some minimal amount of testing for some parts of `autograph-client` already.)

Yeah, will do.

jmhodges:
> > Is this worth adding tests for? (I see we have some minimal amount of testing for some parts of `autograph-client` already.)
> 
> Yeah, will do.

I'm pushing up my refactoring here, but I think I'll have to split this into two PRs. Getting this testable with the code as-is required too much surgery.

1. to get the behavior testable, I had to first get the current `monitor` func testable (even if I all I did was move the `+`/`JoinPath` code out from `monitor` to `main`, of course). This was a sizeable refactoring.
2. reminded me that url parsers like Go's will generate strings that look right for `foobar.com:443`, but, in the intermediate struct value, will put `foobar.com` in the `Scheme` field of the `URL`. So, I've had to change the behavior some from the PR description.

jmhodges:
Made #1040 which this will now be blocked by!

tools/autograph-monitor/monitor.go
```diff
@@ -97,6 +109,21 @@ func main() {
 	}
 }
 
+func rawAutographURLToMonitorEndpoint(autographURLEnvVar string) (string, error) {
+	baseURL, err := url.Parse(autographURLEnvVar)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse AUTOGRAPH_URL as url: %s", err)
+	}
+	if baseURL.Scheme != "https" && baseURL.Scheme != "http" {
+		return "", fmt.Errorf("AUTOGRAPH_URL %#v must be an https:// (or http:// url in testing)", autographURLEnvVar)
```
bhearsum, line 118: Is there any way to definitively detect a testing environment? If so, we should only allow `http` for that situation.

bhearsum (approved):
Thanks for adding the tests! lgtm modulo the optional comment below.

jmhodges:
PTAL, I suspect I had something wrong (using Start instead of StartTLS) in an earlier version of the GoldenPath test that made me think the `http` scheme was required.

bhearsum (approved):

jmhodges:
Oh, wait, it's the integration tests that require it!

jmhodges:
Getting the integration tests right would take more work (either by making them run TLS or start down the dark path of "testing only" configuration) and be another change in behavior from the current which is to allow `http` implicitly. I think I'm going to ship with the explicit call out as-is.

PR #1023, opened 2024-10-08
AUT-251 - Use the same rand in SignHash method

say-yawn:
As part of initializing the contentsignature signer we [call conf.GetRand()](https://github.com/mozilla-services/autograph/blob/19c4068ab8bb2041c913b5212152c7f36478f3cb/signer/contentsignature/contentsignature.go#L77). This will be either [a rand.Reader or crypto11.PKCSRandReader](https://github.com/mozilla-services/autograph/blob/19c4068ab8bb2041c913b5212152c7f36478f3cb/signer/signer.go#L253-L256) depending on whether or not an HSM is available. However, the SignHash method ends up just [using rand.Reader directly](https://github.com/mozilla-services/autograph/blob/19c4068ab8bb2041c913b5212152c7f36478f3cb/signer/contentsignature/contentsignature.go#L157), and s.rand goes unused altogether.

Make sure `SignHash` uses the same rand as the initialized `ContentSigner` by using `s.rand`.

Fix AUT-251

### DIFF
--- a/signer/contentsignature/contentsignature.go
+++ b/signer/contentsignature/contentsignature.go
@@ -3,7 +3,6 @@ package contentsignature // import "github.com/mozilla-services/autograph/signer
 import (
 	"crypto"
 	"crypto/ecdsa"
-	"crypto/rand"
 	"crypto/sha256"
 	"crypto/sha512"
 	"crypto/x509"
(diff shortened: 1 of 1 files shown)


alexcottner (approved):

jmhodges (approved):

say-yawn:
Decided to merge this without tests because the `ContentSignature` is only used for integration testing.

PR #1026, opened 2024-10-08
AUT-287 Note discrepancy between local and CI integration tests

say-yawn:
Additional information on AUT-287

---
`make integration-test` - runs successfully locally

`make generate` and then `docker compose run --build unit-test` runs successfully locally


alexcottner:
I think our objective here is for `make integration-test` to do the same thing that [integration-tests.yaml](https://github.com/mozilla-services/autograph/blob/main/tools/autograph-client/integration-tests.yml) does (which is triggered [here](https://github.com/mozilla-services/autograph/blob/main/.github/workflows/tests.yaml#L96-L130). 

At the moment these are duplicated efforts (two ways to do the same thing, but one only works on github CI).

jmhodges (approved):

PR #1027, opened 2024-10-09
remove make-hsm-ee tool

bhearsum:
This is tangentially related to AUT-264, where I'm looking at updating any tools that work with HSMs to support GCP. However, it appears that this one is simply unused since everything migrated to `contentsignaturepki`. I can't actually find proof of this, but I can't find any documentation or information on when we would actually use this tool...so, let's remove it?


bhearsum:
I found a reference to this in a different repo which has a README and some tools that seem to confirm that this was for usage with the `contentsignature` signer, which we no longer use. Most notably, there's some tools that are designed to update the autograph app config in a way that clearly only works with `contentsignature` (it updates the `privatekey` attribute on a signer, which simply does not exist for `contentsignaturepki`).

jmhodges:
Just to clarify: I think this was actually used in conjunction with the development of the `contentsignaturepki` but is no longer used. It seems to have been a testing tool

Makefile
```diff
@@ -44,11 +44,11 @@ vet:
 	go vet ./...
 
 fmt-diff:
-	gofmt -d *.go database/ signer/ tools/autograph-client/ tools/autograph-monitor tools/softhsm/ tools/hawk-token-maker/ tools/make-hsm-ee/ tools/makecsr/ tools/genpki/
+	gofmt -d *.go database/ signer/ tools/autograph-client/ tools/autograph-monitor tools/softhsm/ tools/hawk-token-maker/ tools/makecsr/ tools/genpki/
 
 fmt-fix:
 	go fmt ./...
-	gofmt -w tools/autograph-client/ tools/softhsm/ tools/hawk-token-maker/ tools/make-hsm-ee/ tools/makecsr/ tools/genpki/
```
jmhodges, line 51: (Not for now, but I wonder why we aren't running `-w` on all of the packages here?)

Makefile
```diff
@@ -44,11 +44,11 @@ vet:
 	go vet ./...
```
jmhodges, line 44: Could you also remove `tools/make-hsm-ee/make-hsm-ee` from .gitignore?

jmhodges (changes requested):

jmhodges:
> Just to clarify: I think this was actually used in conjunction with the development of the `contentsignaturepki` but is no longer used. It seems to have been a testing tool

Oh, you found evidence it was for contentsignature in the sops repo? Well, either way, it's def not something we use today!

Makefile
bhearsum: yes!

bhearsum:
> > Just to clarify: I think this was actually used in conjunction with the development of the `contentsignaturepki` but is no longer used. It seems to have been a testing tool
> 
> Oh, you found evidence it was for contentsignature in the sops repo? Well, either way, it's def not something we use today!

Yeah, if you dig into `hsm/content-signature` there's a README and various scripts that seem to deal with generating new certs and updating autograph app configs afterwards.

jmhodges (approved):

PR #1029, opened 2024-10-09
bump Go version to 1.23

jmhodges:
I would like `slices.Collect` and `maps.Keys` for my code and those were
added in Go 1.23.


jmhodges:
Going to sideport us to golang-bookworm at the same version, first.

PR #1030, opened 2024-10-10
refresh makecsr and add it to the Dockerfile

jmhodges:
The makecsr tool is used to generate a CSR for our addon and
contentsignaturepki intermediates

This patch does a mild refactoring to bring it under test and adds it to
the docker image so it can be used from our private hsm tooling repo.

We also add setting the Subject Alternative Name DNS names to it. This
will make it easier for future tooling that's dealing with Name
Constraints verify it correctly.

Along the way, we also require passing in a signature algorithm.
Signature algorithm, like the name constraints, is something that can be
overridden at CA signing time, but this removes some of the per-cloud
implicit configuration that was in the tool.

Updates AUT-279
Updates AUT-280

### DIFF
--- a/tools/makecsr/README.md
+++ b/tools/makecsr/README.md
@@ -5,10 +5,10 @@ hosted in our HSMs. It's used particularly to issue the CS and AMO intermediates
 used by autograph. See our private hsm repo for how we've invoked it.
 
 Note: nearly all of the CSRs attributes can be overridden at signing time, so
-this is not a complete picture (esp. of the name constraints). From SAN to
-signature algorithm. But we include those out of a desire to be as explicit as
-we can be, at the cost of perhaps confusing ourselves about all the places those
-attributes must be specified.
+this is not a complete picture of what will be signed. But we include attributes
+like subject alternative names and signature algorithm out of a desire to be as
+explicit as we can be. This comes at the cost of perhaps confusing ourselves
+about all the places those attributes must be specified.
 
 If you're invoking in GCP, be sure to set the `KMS_PKCS11_CONFIG` env var to the
 YAML config file that the libkmsp11 library requires.
--- a/tools/makecsr/README.md
+++ b/tools/makecsr/README.md
@@ -1,8 +1,52 @@
-makecsr
-=======
+# makecsr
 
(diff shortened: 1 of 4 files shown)


tools/makecsr/README.md
```diff
@@ -1,8 +1,52 @@
-makecsr
-=======
+# makecsr
 
-This is a small helper used to generate a CSR from a private key hosted in HSM.
-It's used particularly to issue the CS and AMO intermediates used by autograph.
+This is a small helper used to generate a PEM-encoded CSR from a private key
+hosted in our HSMs. It's used particularly to issue the CS and AMO intermediates
+used by autograph. See our private hsm repo for how we've invoked it.
```
bhearsum, line 5: It doesn't necessarily need to be noted here, but I assume we use this for any CSRs we may need from non-exported HSM keys? (eg: authenticode)

tools/makecsr/makecsr.go
```diff
@@ -39,20 +18,59 @@ import (
 	"github.com/mozilla-services/autograph/crypto11"
 )
 
+var (
+	allowedSigAlgs = map[string]x509.SignatureAlgorithm{
+		"SHA256WithRSA":   x509.SHA256WithRSA,
+		"SHA384WithRSA":   x509.SHA384WithRSA,
```
bhearsum, line 24: SHA384WithRSA doesn't work with GCP KMS. Can we add a note here to remove this when AWS support is removed? I also wonder if it's worth checking for this explicitly further down so we can offer a very clear error message.

bhearsum (approved):

tools/makecsr/README.md
jmhodges: The only documented use was for creating intermediates from the addons/content signature root. 

I've not yet seen what templating Digicert and authenticode needs in its CSR, however. The mozilla-services/hsm scripts do a lot of overriding attributes in the CSR that I imagine Digicert wouldn't.

tools/makecsr/makecsr.go
jmhodges: Added. PTAL

bhearsum (approved):

PR #1032, opened 2024-10-10
AUT-313: break up too-large-for-GCP random reads before sending them to libkmsp11

bhearsum:
This allows for the reading of random bytes of any size, which is needed by various operations that Autograph performs.

### DIFF
--- a/signer/hsm.go
+++ b/signer/hsm.go
@@ -27,12 +27,12 @@ func (hsm *GenericHSM) GetPrivateKey(label []byte) (crypto.PrivateKey, error) {
 	return crypto11.FindKeyPair(nil, label)
 }
 
-func (hsm *GenericHSM) GetRand() io.Reader {
-	return new(crypto11.PKCS11RandReader)
-}
-
 type AWSHSM struct {
 	GenericHSM
+}
+
+func (hsm *AWSHSM) GetRand() io.Reader {
+	return new(crypto11.PKCS11RandReader)
 }
 
 func (hsm *AWSHSM) MakeKey(keyTpl interface{}, keyName string) (crypto.PrivateKey, crypto.PublicKey, error) {
@@ -105,6 +105,35 @@ const CKA_GOOGLE_DEFINED_KMS_ALGORITHM = CKA_GOOGLE_DEFINED | CKA_KMS_ALGORITHM
 
 type GCPHSM struct {
 	GenericHSM
+}
+
+type unlimitedBytesRandReader struct {
+	rr crypto11.PKCS11RandReader
+}
+
+func (ubrr *unlimitedBytesRandReader) Read(data []byte) (int, error) {
+	var n int
+	var err error
+	need := len(data)
+	for need > 0 {
+		var bytesRead int
+		next := min(need, 1024)
+		bytesRead, err = ubrr.rr.Read(data[n : n+next])
+		n += bytesRead
+		if err != nil {
+			return n, err
+		}
+		need -= bytesRead
+	}
+	return n, err
+}
+
+// GCPHSM.GetRand returns an io.Reader that can generate an unlimited
+// number of bytes from a crypto11.PKCS11RandReader. This is needed because
+// GCP KMS can generate a maximum of 1024 bytes per call.
+// https://cloud.google.com/kms/docs/generate-random#known_limitations
+func (hsm *GCPHSM) GetRand() io.Reader {
+	return new(unlimitedBytesRandReader)
 }
 
 func (hsm *GCPHSM) MakeKey(keyTpl interface{}, keyName string) (crypto.PrivateKey, crypto.PublicKey, error) {
--- a/signer/signer_test.go
+++ b/signer/signer_test.go
@@ -7,9 +7,11 @@
 package signer
 
 import (
+	"bytes"
 	"crypto/ecdsa"
 	"crypto/rsa"
 	"encoding/asn1"
+	"errors"
 	"fmt"
 	"testing"
 
@@ -218,7 +220,7 @@ func TestNoSuitableKeyFound(t *testing.T) {
 }
 
 func TestMakeKey(t *testing.T) {
-	for i, testcase := range PASSINGTESTCASES {
+	for i, testcase := range passingTestCases {
 		_, keyTpl, _, err := testcase.cfg.GetKeys()
 		if err != nil {
 			t.Fatalf("testcase %d failed to load signer configuration: %v", i, err)
@@ -235,7 +237,7 @@ func TestMakeKey(t *testing.T) {
 	}
 }
 
-var PASSINGTESTCASES = []struct {
+var passingTestCases = []struct {
 	cfg Configuration
 }{
 	{cfg: Configuration{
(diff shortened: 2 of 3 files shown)


signer/hsm.go
```diff
@@ -107,6 +107,34 @@ type GCPHSM struct {
...
 
+type UnlimitedBytesRandReader struct {
+	rr crypto11.PKCS11RandReader
+}
+
+func (ubrr *UnlimitedBytesRandReader) Read(data []byte) (n int, err error) {
+	var result []byte
+	need := len(data)
+	for need > 0 {
+		next := make([]byte, 1024)
```
jmhodges, line 118: `min(need, 1024)` or similar?

bhearsum: Yeah, this needs some more iteration (and tests)!

bhearsum:
@say-yawn prompted me to add some test cases for failures; I'm moving back to draft while I do that.

signer/hsm.go
```diff
@@ -107,6 +107,36 @@ type GCPHSM struct {
...
+func (ubrr *unlimitedBytesRandReader) Read(data []byte) (n int, err error) {
+	var result []byte
+	need := len(data)
+	for need > 0 {
+		var bytesRead int
+		next := make([]byte, min(need, 1024))
+		bytesRead, err = ubrr.rr.Read(next)
+		n += bytesRead
+		if err != nil {
+			return
```
alexcottner, line 123: Is there any way to not lose the error here and return/log it?

bhearsum: AIUI, we're returning that error (the `err` variable is bound to the return as a "named result parameter": https://go.dev/doc/effective_go#named-results)

I think this is verified by the new test where we mock in a specific error that `GenerateRandom` returns, and verify that it comes out of UnlimitedBytesRandReader.Read: https://github.com/mozilla-services/autograph/pull/1032/files#diff-a43f626c786b027528255923aee643b55513fb5d596a69e082c63dba5b9ccfb9R833-R857

alexcottner: Ah, overlooked that thank you.

alexcottner (approved):

jmhodges:
Good patch! I had some small things that I could get push back on.

signer/hsm.go
```diff
@@ -107,6 +107,36 @@ type GCPHSM struct {
...
+		}
+		result = append(result, next...)
+		need -= bytesRead
+	}
+	copy(data, result)
```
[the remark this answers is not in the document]
jmhodges: Enh, this isn't worth it, probably.

signer/signer_test.go
```diff
@@ -662,3 +663,208 @@ func TestMakeKeyGCPRSA(t *testing.T) {
 		t.Fatalf("MakeKey failed: %v", err)
 	}
 }
+
+var UNLIMITEDBYTESREADERTESTCASES = []struct {
```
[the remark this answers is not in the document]
bhearsum: I can adjust this; I'm going to change the style for the whole file though to maintain consistency.

signer/signer_test.go
```diff
@@ -662,3 +663,208 @@ func TestMakeKeyGCPRSA(t *testing.T) {
...
+		expectedSessions += len(testcase.expectedPkcs11RandReaderBytes)
+	}
+	mockCtx.EXPECT().OpenSession(slot, uint(6)).Return(session, nil).Times(expectedSessions)
+
+	mockFactory := mockedPKCS11ContextFactory(mockCtx)
+	crypto11.Configure(&crypto11.PKCS11Config{}, mockFactory)
+	defer crypto11.Close()
+
+	for _, testcase := range UNLIMITEDBYTESREADERTESTCASES {
+		for _, size := range testcase.expectedPkcs11RandReaderBytes {
```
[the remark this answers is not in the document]
bhearsum: I could certainly add this, but unless I'm missing something it would require me to have the mock (or an additional mock) behave differently than the real method, which doesn't seem ideal.

A different thing we could do is to have `unlimitedBytesRandReader.Read` mirror the behaviour of the underlying method, and have `n` always be 0 or `len(data)`. I'm not super excited to do that either, but the more I think about it, the more it actually seems like the most correct thing to do?

signer/hsm.go
bhearsum: Sure. I thought they were idiomatic, but I have no real preference.

signer/signer_test.go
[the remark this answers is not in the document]
jmhodges: If it returns 0, it'd return an error, right? You could have the mock return an error.

Just an FYI, there is an [InOrder](https://pkg.go.dev/go.uber.org/mock/gomock#InOrder) api that makes it a lil easier to interleave responses.

signer/hsm.go
```diff
@@ -107,6 +107,36 @@ type GCPHSM struct {
...
+	need := len(data)
+	for need > 0 {
+		var bytesRead int
+		next := make([]byte, min(need, 1024))
+		bytesRead, err = ubrr.rr.Read(next)
+		n += bytesRead
+		if err != nil {
+			return n, err
+		}
+		data = append(data, next...)
```
jmhodges, line 126: I'm not sure how this is working correctly. It only appends to data, but does not overwrite existing data in `data`. That doesn't seem correct?

jmhodges (changes requested):

signer/hsm.go
jmhodges: I think going back to the copying version is fine, but we should figure out why the tests were able to pass.

signer/hsm.go
jmhodges: Some folks in Go-land like them, but I think the community is mostly on the side of "too easy to have shadowing problems".

signer/signer_test.go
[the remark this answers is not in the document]
bhearsum: That's my understanding - yes.

Oh wait, I think I just realized you were asking to make sure that our code behaves correctly -- not that it returns > 0 on error. Yes of course, that's totally doable (and sensible).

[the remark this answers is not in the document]
bhearsum: Oh, this is already covered though? The mock is on `GenerateRandom`, which AFAIK might return a non-zero number of bytes, but also an error. https://github.com/mozilla-services/autograph/pull/1032/files#diff-a43f626c786b027528255923aee643b55513fb5d596a69e082c63dba5b9ccfb9R840-R845 covers this, I believe?

bhearsum:
@jmhodges - we've had a bunch of back and forth here. I _think_ this is done now, so I'll send it back. Feel free to catch me on Slack if we need to hash anything out still.

signer/signer_test.go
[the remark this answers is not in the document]
jmhodges: SGTM!

jmhodges:
> @jmhodges - we've had a bunch of back and forth here. I _think_ this is done now, so I'll send it back. Feel free to catch me on Slack if we need to hash anything out still.

I don't see any changes in the testing which makes me nervous that our testing is incomplete. What am I missing?

signer/signer_test.go
```diff
@@ -662,3 +664,262 @@ func TestMakeKeyGCPRSA(t *testing.T) {
...
+	},
+	{
+		requestedBytes: 5000,
+		expectedBytes: append(
+			append(
+				bytes.Repeat([]byte("A"), 1024),
+				bytes.Repeat([]byte("B"), 1024)...,
+			),
+			bytes.Repeat([]byte{0}, 2952)...,
+		),
```
bhearsum, line 843: There may be a better way to generate these inline, but I couldn't find it.

bhearsum:
@jmhodges Sorry about all the confusion around the tests. I updated them to test actual bytes returned rather than just the lengths. This also, of course, uncovered the fact that we weren't actually returning partially satisfied requests in UnlimitedBytesRandReader.Read, so I fixed that (and got rid of the unnecessary `copy` on the way).

jmhodges (approved):

signer/signer_test.go
jmhodges: yeah, the only think i could think of was a "pre-processing" for-loop that turned sizes into slices and did byte math with the iterator (`byte('a')+i`) between the test cases and the actual uses. But that has the downsize of not being explicit in what folks should see. I like this better!

PR #1033, opened 2024-10-10
AUT-251 Use the same rand on SignHash as the ContentSigner

say-yawn:
While testing PR 1023, we found that the KMS permission was not added to the dev environment resulting in the error:
```
panic: interface conversion: signer.Signature is nil, not *contentsignature.ContentSignature

goroutine 101 [running]:
github.com/mozilla-services/autograph/signer/contentsignature.(*ContentSigner).SignData(0xc0003cc1a0, {0x20054d0?, 0xc0006b4418?, 0xc0006b4120?}, {0x0, 0x0})
	/app/src/autograph/signer/contentsignature/contentsignature.go:110 +0xfa
main.(*monitor).checkSigners(0xc0003d3830)
	/app/src/autograph/monitor.go:77 +0xc62
main.(*monitor).start(0xc0003d3830, 0xc00014a7e0?)
	/app/src/autograph/monitor.go:54 +0xe5
created by main.newMonitor in goroutine 1
	/app/src/autograph/monitor.go:148 +0x214
```
The real error on not being able to read the random stream is not returned and instead swallowed by the error trying to assign the signature’s hash name: https://github.com/mozilla-services/autograph/blob/1ca235057db53c8be20664d8c1b7d7ea7d4fff63/signer/contentsignature/contentsignature.go#L111

The fix was to check for the `ContentSigner` being `nil` before updating values to it.
The rand update was also applied to `SignData` in `ContentSignaturePKI`: https://github.com/mozilla-services/autograph/blob/1ca235057db53c8be20664d8c1b7d7ea7d4fff63/signer/contentsignaturepki/contentsignature.go#L265

# Note
Related #1028 and #1023
Dev permission to KMS updated on https://github.com/mozilla-it/webservices-infra/pull/3020
Fix AUT-251

### DIFF
--- a/signer/contentsignature/contentsignature.go
+++ b/signer/contentsignature/contentsignature.go
@@ -107,7 +107,9 @@ func (s *ContentSigner) SignData(input []byte, options interface{}) (signer.Sign
 	}
 	alg, hash := makeTemplatedHash(input, s.Mode)
 	sig, err := s.SignHash(hash, options)
-	sig.(*verifier.ContentSignature).HashName = alg
+	if err == nil {
+		sig.(*verifier.ContentSignature).HashName = alg
+	}
 	return sig, err
 }
 
--- a/signer/contentsignature/contentsignature.go
+++ b/signer/contentsignature/contentsignature.go
@@ -3,7 +3,6 @@ package contentsignature // import "github.com/mozilla-services/autograph/signer
 import (
 	"crypto"
 	"crypto/ecdsa"
-	"crypto/rand"
 	"crypto/sha256"
 	"crypto/sha512"
 	"crypto/x509"
@@ -154,7 +153,8 @@ func (s *ContentSigner) SignHash(input []byte, options interface{}) (signer.Sign
 		ID:   s.ID,
 	}
 
-	asn1Sig, err := s.priv.(crypto.Signer).Sign(rand.Reader, input, nil)
+	// asn1Sig, err := s.priv.(crypto.Signer).Sign(rand.Reader, input, nil)
+	asn1Sig, err := s.priv.(crypto.Signer).Sign(s.rand, input, nil)
 	if err != nil {
 		return nil, fmt.Errorf("contentsignature: failed to sign hash: %w", err)
 	}
--- a/signer/contentsignature/contentsignature.go
+++ b/signer/contentsignature/contentsignature.go
@@ -105,13 +105,14 @@ func (s *ContentSigner) SignData(input []byte, options interface{}) (signer.Sign
 	if len(input) < 10 {
 		return nil, fmt.Errorf("contentsignature: refusing to sign input data shorter than 10 bytes")
 	}
+
 	alg, hash := makeTemplatedHash(input, s.Mode)
 	sig, err := s.SignHash(hash, options)
-
-	if sig != nil {
-		// sig could be nil when there is misconfigured permission or the kms is not available
-		sig.(*verifier.ContentSignature).HashName = alg
+	if err != nil {
+		return nil, err
 	}
+
+	sig.(*verifier.ContentSignature).HashName = alg
 	return sig, err
 }
 
--- a/signer/contentsignature/contentsignature.go
+++ b/signer/contentsignature/contentsignature.go
@@ -107,7 +107,9 @@ func (s *ContentSigner) SignData(input []byte, options interface{}) (signer.Sign
 	}
 	alg, hash := makeTemplatedHash(input, s.Mode)
 	sig, err := s.SignHash(hash, options)
-	if err == nil {
+
+	if sig != nil {
+		// sig could be nil when there is misconfigured permission or the kms is not available
 		sig.(*verifier.ContentSignature).HashName = alg
 	}
 	return sig, err
@@ -155,7 +157,6 @@ func (s *ContentSigner) SignHash(input []byte, options interface{}) (signer.Sign
 		ID:   s.ID,
 	}
 
-	// asn1Sig, err := s.priv.(crypto.Signer).Sign(rand.Reader, input, nil)
 	asn1Sig, err := s.priv.(crypto.Signer).Sign(s.rand, input, nil)
 	if err != nil {
 		return nil, fmt.Errorf("contentsignature: failed to sign hash: %w", err)
--- a/signer/contentsignature/contentsignature_test.go
+++ b/signer/contentsignature/contentsignature_test.go
@@ -218,3 +218,71 @@ func TestNoShortData(t *testing.T) {
 		t.Fatalf("expected to fail with input data too short but failed with: %v", err)
 	}
 }
+
+// Reproduce nil error
+var NILTESTCASE = []struct {
+	cfg signer.Configuration
+}{
+	// 	{cfg: signer.Configuration{
+	// 		Type: Type,
+	// 		ID:   P256ECDSA,
+	// 		X5U:  "https://foo.bar/chain.pem",
+	// 		PrivateKey: `
+	// -----BEGIN EC PARAMETERS-----
+	// BggqhkjOPQMBBw==
+	// -----END EC PARAMETERS-----
+	// -----BEGIN EC PRIVATE KEY-----
+	// MHcCAQEEII+Is30aP9wrB/H6AkKrJjMG8EVY2WseSFHTfWGCIk7voAoGCCqGSM49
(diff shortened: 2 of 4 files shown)


alexcottner (approved):

signer/contentsignature/contentsignature.go
```diff
@@ -108,7 +107,11 @@ func (s *ContentSigner) SignData(input []byte, options interface{}) (signer.Sign
 	}
 	alg, hash := makeTemplatedHash(input, s.Mode)
 	sig, err := s.SignHash(hash, options)
-	sig.(*verifier.ContentSignature).HashName = alg
+
+	if sig != nil {
+		// sig could be nil when there is misconfigured permission or the kms is not available
+		sig.(*verifier.ContentSignature).HashName = alg
+	}
```
jmhodges, line 114: So, I have a few different bits of feedback for this.

The first: Only use objects after you're sure the error returned with them is nil. So, instead of doing this sig check, return early if `err != nil`. That err check and early return is the change that needs to happen for this patch to be shippable.

Second, doing nil checks against interfaces generally don't go well because people misinterpret their meaning and assume things are safe about the interface after that check that are not true. See https://trstringer.com/go-nil-interface-and-interface-with-nil-concrete-value/ and https://go.dev/tour/methods/13. In case be fine in cases, but it's a pattern we don't want to support and I don't think this change is shippable with it. 

I had some other points about what you could do if backed into a corner with nil interface checks, but if you were, other things would be different and my advice would be, too.

signer/contentsignaturepki/contentsignature.go
```diff
@@ -219,7 +218,9 @@ func (s *ContentSigner) SignData(input []byte, options interface{}) (signer.Sign
 	}
 	alg, hash := MakeTemplatedHash(input, s.Mode)
 	sig, err := s.SignHash(hash, options)
-	sig.(*verifier.ContentSignature).HashName = alg
+	if sig != nil {
+		sig.(*verifier.ContentSignature).HashName = alg
+	}
```
jmhodges, line 223: Same feedback about err checking here.

signer/contentsignature/contentsignature_test.go
```diff
@@ -218,3 +219,21 @@ func TestNoShortData(t *testing.T) {
 		t.Fatalf("expected to fail with input data too short but failed with: %v", err)
 	}
 }
+
+// Reproduce nil error
```
jmhodges, line 223: This comment doesn't explain what's going on. I think we could leave it out.

signer/contentsignature/contentsignature_test.go
```diff
@@ -218,3 +219,21 @@ func TestNoShortData(t *testing.T) {
...
+
+func TestRandReadFailureOnSignHash(t *testing.T) {
+	input := []byte("foobarbaz1234abcd")
+	testcase := PASSINGTESTCASES[0]
+	s, _ := New(testcase.cfg)
+	s.rand = &ErrorReader{}
+	_, err_nil := s.SignData(input, nil)
+	if err_nil == nil {
+		t.Fatal("Should have failed to sign data with error in the SignHash.")
+	}
```
jmhodges, line 238: just `err`, please. There are no underscores in Go code, but also giving errors different names are things we Go hackers do only to prevent shadowing.

signer/contentsignature/contentsignature_test.go
```diff
@@ -218,3 +219,21 @@ func TestNoShortData(t *testing.T) {
...
 }
+
+// Reproduce nil error
+type ErrorReader struct{}
+
+func (e *ErrorReader) Read(p []byte) (n int, err error) {
+	return 0, errors.New("read error")
+}
+
+func TestRandReadFailureOnSignHash(t *testing.T) {
```
jmhodges, line 230: Thanks for this test!

signer/contentsignature/contentsignature_test.go
```diff
@@ -218,3 +219,21 @@ func TestNoShortData(t *testing.T) {
...
+	input := []byte("foobarbaz1234abcd")
+	testcase := PASSINGTESTCASES[0]
+	s, _ := New(testcase.cfg)
+	s.rand = &ErrorReader{}
```
jmhodges, line 234: I think we may want a TODO, at least, to inject `rand.Reader` into either the configuration objects (ugh) or pass them into the `New` constructors here and in contentsignaturepki to allow us to more easily test this code.

I assume y'all took a look at what it would take to pass the Reader into all the `signer/*.New` calls were left hoping for a smaller change.

signer/contentsignaturepki/contentsignature_test.go
```diff
@@ -278,3 +279,21 @@ func TestNoShortData(t *testing.T) {
 		t.Fatalf("expected to fail with input data too short but failed with: %v", err)
 	}
 }
+
+// Reproduce nil error
```
jmhodges, line 283: same feedback

signer/contentsignaturepki/contentsignature_test.go
```diff
@@ -278,3 +279,21 @@ func TestNoShortData(t *testing.T) {
...
+func (e *ErrorReader) Read(p []byte) (n int, err error) {
+	return 0, errors.New("read error")
+}
+
+func TestReadRandFailureOnSignHash(t *testing.T) {
+	input := []byte("foobarbaz1234abcd")
+	testcase := PASSINGTESTCASES[0]
+	s, _ := New(testcase.cfg)
+	s.rand = &ErrorReader{}
+	_, err_nil := s.SignData(input, nil)
```
jmhodges, line 295: same err name feedback

jmhodges (changes requested):

signer/contentsignature/contentsignature.go
jmhodges: Here's a related problem we had to fix up this week: https://github.com/mozilla-services/autograph/pull/1034

signer/contentsignaturepki/contentsignature_test.go
say-yawn: Resolved in [7b19fee](https://github.com/mozilla-services/autograph/pull/1033/commits/7b19fee2a2e3daac50d64af473f12aa4ea1f72c8)

signer/contentsignaturepki/contentsignature_test.go
say-yawn: Resolved [7b19fee](https://github.com/mozilla-services/autograph/pull/1033/commits/7b19fee2a2e3daac50d64af473f12aa4ea1f72c8)

signer/contentsignature/contentsignature_test.go
say-yawn: Decided to keep it small and add a TODO. @alexcottner and I wanted to think about it more since the rand not being available is an error case caused by other dependency. Resolved [7b19fee](https://github.com/mozilla-services/autograph/pull/1033/commits/7b19fee2a2e3daac50d64af473f12aa4ea1f72c8)

signer/contentsignature/contentsignature_test.go
say-yawn: Good to know. Thanks for sharing this bit on Go best practices.

Resolved [7b19fee](https://github.com/mozilla-services/autograph/pull/1033/commits/7b19fee2a2e3daac50d64af473f12aa4ea1f72c8)

signer/contentsignature/contentsignature_test.go
say-yawn: Resolved [7b19fee](https://github.com/mozilla-services/autograph/pull/1033/commits/7b19fee2a2e3daac50d64af473f12aa4ea1f72c8)

signer/contentsignaturepki/contentsignature.go
say-yawn: Resolved [7b19fee](https://github.com/mozilla-services/autograph/pull/1033/commits/7b19fee2a2e3daac50d64af473f12aa4ea1f72c8)

signer/contentsignature/contentsignature.go
say-yawn: Thanks for this info @jmhodges, I was unfamiliar with interface and happy to learn more about it. Resolved [7b19fee](https://github.com/mozilla-services/autograph/pull/1033/commits/7b19fee2a2e3daac50d64af473f12aa4ea1f72c8)

PR #1034, opened 2024-10-11
crypto11: handle nil returns from pcks11.New

jmhodges:
Previous to this patch, a busted Path configuration passed to crypto11
would cause nil pointer panics.

This is because attempt to check the nil pointer of the PKCS11Context
interface would run into the usual "nil interface" problems where the
pointer would not be technically "nil" but calling methods on it would
cause nil pointer panics. See <https://go.dev/tour/methods/12> and
<https://trstringer.com/go-nil-interface-and-interface-with-nil-concrete-value/>
for details.

So, to handle this correctly, we need to return an error along with our
interface context type.

It's pretty surprising that `pkcs11.New` can return nil, but we do what
we must.

Updates AUT-283

### DIFF
--- a/crypto11/crypto11.go
+++ b/crypto11/crypto11.go
@@ -238,10 +238,14 @@ type PKCS11Config struct {
 	PoolWaitTimeout time.Duration
 }
 
-type PKCS11ContextFactory func(*PKCS11Config) PKCS11Context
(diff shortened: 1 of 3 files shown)


bhearsum (approved):
Thanks for fixing this up.

PR #1040, opened 2024-10-16
monitor: make monitor testable

jmhodges:
This makes the Handler and monitor funcs accept some arguments so we can
write tests for them. We write one simple golden path test to
demonstrate.

This is an outcome of trying to make the AUTOGRAPH_URL env var handling
better in #1012. That PR will be rebased on to this one.

### DIFF
--- a/tools/autograph-monitor/monitor.go
+++ b/tools/autograph-monitor/monitor.go
@@ -44,11 +44,8 @@ type configuration struct {
 
 const inputdata string = "AUTOGRAPH MONITORING"
 
-var (
-	conf configuration
-)
-
 func main() {
+	conf := &configuration{}
 	conf.url = os.Getenv("AUTOGRAPH_URL")
 	if conf.url == "" {
 		log.Fatal("AUTOGRAPH_URL must be set to the base url of the autograph service")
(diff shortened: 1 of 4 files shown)


tools/autograph-monitor/monitor.go
```diff
@@ -90,9 +87,9 @@ func main() {
 
 	if os.Getenv("LAMBDA_TASK_ROOT") != "" {
 		// we are inside a lambda environment so run as lambda
-		lambda.Start(Handler)
+		lambda.Start(func() error { return Handler(conf, http.DefaultClient) })
```
alexcottner, line 90: A little nit-picky. I think `Handler`'s only purpose was to be a lambda-compatible wrapper around `monitor`. But now we're wrapping the wrapper with this function. Why not just get rid of `Handler` now?

jmhodges: It has that GC call inside of it and that might actually still matter, so I didn't want to adjust or remove it.

jmhodges: And moving all that code out here just to capture it all again feels pretty not great.

jmhodges: (Cuz the point of this wrapper func is to capture the local variable `conf` in `main`.)

jmhodges: Hm, are you maybe asking for `conf` to still be a global that Handler knows to look up? I was hoping to avoid needing mutable globals after this change.

alexcottner: I guess this is a moot point as the lambda is going away soon anyway.

alexcottner (approved):

PR #1045, opened 2024-10-22
Add Makefile commands to run docker container interactively for dev work

alexcottner:
Added a couple options to Makefile to run the docker container locally in interactive mode for easier dev work.


Dockerfile
```diff
@@ -29,7 +29,8 @@ RUN apt-get update && \
         libc6-dev \
         pkg-config \
         curl \
-        jq
+        jq \
```
alexcottner, line 32: From #1044

Makefile
```diff
@@ -96,3 +96,19 @@ dummy-statsd:
...
+	docker run --rm -it --user 0:0 \
+		-v "./:/app/src/autograph" \
+		mozilla/autograph:latest /bin/bash
+
+# pull the docker image, and run it locally with our source code mounted under /app/src/autograph
+# run as root to allow for changes, package installs, etc
+pull-and-run-interactive:
+	docker pull mozilla/autograph:latest
+	docker run --rm -it --user 0:0 \
+		-v "./:/app/src/autograph" \
```
jmhodges, line 113: Have you tested this code? I don't believe relative paths like `./` work on the left side here (which is why we have to put `$PWD` there). Did something change?

bhearsum: It seems like this works:
```
~/tmp/2024-10-30 ❯ touch foo
~/tmp/2024-10-30 ❯ docker run -v ./:/app --rm --entrypoint /bin/bash -it ubuntu:latest -i                  
Unable to find image 'ubuntu:latest' locally
latest: Pulling from library/ubuntu
ff65ddf9395b: Pull complete 
Digest: sha256:99c35190e22d294cdace2783ac55effc69d32896daaa265f0bbedbcde4fbe3e5
Status: Downloaded newer image for ubuntu:latest
root@eb64bee7ac63:/# cd /app
root@eb64bee7ac63:/app# ls
foo
```

But if we're concerned about `make -C` invocations it may be problematic. Usually in cases like this I would use [the dirname of the current file](https://searchfox.org/mozilla-central/rev/5fdbe0d78d61581aa4b37ce82bbf31c2bd7826f1/tools/update-verify/release/updates/verify.sh#28-30), but I'm not sure offhand how to do that with Make.

Dockerfile
jmhodges: Rebase?

jmhodges (approved):

Makefile
jmhodges: Sorry, I tested this and I was wrong! Forgot to leave a reply here to that effect. Approved!

PR #1046, opened 2024-10-25
add benchmarks for GKE random generators

jmhodges:
These are to be run in the GKE to confirm my hypothesis about the
performance of the GCP KMS GenerateRandom RPC vs the local random data
generator. I suspect that we're not getting much benefit from the KMS
GenerateRandom (which is called by libkmsp11). The security properties
of it are also not sufficiently different than the modern, very good
random number generators available on Linux in the Go standard library.

Updates AUT-335

### DIFF
--- a/tools/benchrand/benchrand_test.go
+++ b/tools/benchrand/benchrand_test.go
@@ -1,0 +1,50 @@
+package main
+
+import (
+	"crypto/rand"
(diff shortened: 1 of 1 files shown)


jmhodges:
Mooted by the rate limit problems! See #1047

PR #1047, opened 2024-10-25
use local random generator in GCP

jmhodges:
We tried making the GCP KMS GenerateRandom one work (that's called by
libkmsp11 through the crypto11.PKCS11RandReader interface) but we hit
rate limits quickly (esp because of Go's use of randutil.MaybeReadByte)
and it required a lot of code to fit all reads between 8 and 1024 bytes.
See https://cloud.google.com/kms/docs/generate-random#known_limitations

Modern Linux systems have a great /dev/urandom and Go uses that. It's
security properties are as good or better than HSM implementations.

We hadn't even finished the code to use at least 8 bytes, so this avoids
even more code churn.


signer/signer_test.go
```diff
@@ -664,262 +662,3 @@ func TestMakeKeyGCPRSA(t *testing.T) {
 		t.Fatalf("MakeKey failed: %v", err)
 	}
 }
-
```
bhearsum, line 667: My beautiful tests! :sob:

bhearsum (approved):

signer/signer_test.go
jmhodges: lmao the life of the engineer 💀

PR #1048, opened 2024-10-25
monitor: add the gcp dev environment

jmhodges:
This requires setting up AUTOGRAPH_ENV to `dev` in the GCP dev
environment to use. But this the root cert now used by autograph there.

Along the way, we do some renames for clarity.


jbuck (approved):

say-yawn (approved):
LGTM though we need to make sure to push the `AUTOGRAPH_ENV` changes from the config repo before.

PR #1051, opened 2024-10-30
fix: replace setup-crane step with copypasta

bhearsum:
This step is trivial enough that we may as well copy/paste rather than go through the trouble of getting it approved.


bhearsum:
I tested the copypaste in a separate (but now removed) workflow: https://github.com/mozilla-services/autograph/actions/runs/11603541515/job/32310695841?pr=1051

jmhodges (approved):

jmhodges:
This could also just be a `go install` call which would get us binary transparency checks, but copy pasta is fine

PR #1058, opened 2024-11-05
AUT-350 - Adding more permissions for unit test database container.

alexcottner:
[AUT-350](https://mozilla-hub.atlassian.net/browse/AUT-350)

**Problem:** When unit tests run, they fail to truncate the database table as the postgres user account doesn't have permission to do that. But it fails silently without an explicit `Fatalf` call, so we didn't realize.

**This solution:** We mount an additional initdb file for the database container in docker-compose. This means we don't include it in the Dockerfile or modify `schema.sql` so it's clear that this is for unit tests only.

**Other solutions:**

1. Modify `schema.sql` - This could grant our user account more permissions than intended if somebody runs it without checking.
2. Modify the database Dockerfile - This means it would be included whenever the container is built, which again could accidentally grant our user account more permissions than intended if we decide to use a container DB in the future.
3. Modify the postgres startup command (grant more permissions on the fly) - This _feels_ less intuitive.
4. Make an admin user account for the unit tests to clear data - More code just for unit tests makes me itchy.

[AUT-350]: https://mozilla-hub.atlassian.net/browse/AUT-350?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jmhodges (approved):

jmhodges:
Oh, I just approved it, but could we change our handlers_test.go to do this truncate before the test runs? I think we need to do that anyway, but would also demonstrate the problem.

Or we could panic in Cleanup or something

jmhodges:
I could be told nah, tho. We can do it as follow ups

jmhodges:
You wanna merge this or are you thinking about doing that work? (I would like to post my PR that uses this work to require a database in contentsignaturepki)

jmhodges:
Hrm, I'm not seeing this allow me to use `truncate` in the tests in my branch. Have you confirmed it works? Or is this me  PEBKAC'ing my docker compose work?

alexcottner:
> Hrm, I'm not seeing this allow me to use `truncate` in the tests in my branch. Have you confirmed it works? Or is this me PEBKAC'ing my docker compose work?

To double check myself, I made this change.
```
diff --git a/database/queries_test.go b/database/queries_test.go
index 80a84de7..39b02006 100644
--- a/database/queries_test.go
+++ b/database/queries_test.go
@@ -2,6 +2,7 @@ package database
 
 import (
        "fmt"
+       "log"
        "reflect"
        "sync"
        "testing"
@@ -17,6 +18,12 @@ func TestConcurrentEndEntityOperations(t *testing.T) {
                Host:                host + ":5432",
                MonitorPollInterval: 10 * time.Second,
        })
+
+       _, err = db.Exec("truncate table endentities;")
+       if err != nil {
+               log.Fatalf("Failed to truncate endentities;")
+       }
+
        if err != nil {
                t.Fatal(err)
        }
```

And then ran:
```
make build
docker compose run unit-test
```

And it worked fine for me. Wanna pair tomorrow?

jmhodges:
I missed this comment in reply to me! I apologize for the delay. Let me investigate my situation. Perhaps the same problem that caused my debian GPG problems also caused my local problem.

jmhodges:
It was PEBKAC! I had left in code using a different user

jmhodges (approved):

PR #1059, opened 2024-11-07
build our docker images for multiple platforms

jmhodges:
Following this documentation:
https://docs.docker.com/build/ci/github-actions/multi-platform/

This clears up a warning in our build, and makes it easier for folks on
arm64 (like us macOS people) to pull and use the images.

Along the way, I also fixed the warnings about using lowercase `as` when
we use uppercase commands in our Dockerfile.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -20,6 +20,9 @@ jobs:
       - name: Clone repository
         uses: actions/checkout@v4
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
       - name: Setup Docker Buildx
         uses: docker/setup-buildx-action@v3
 
@@ -73,6 +76,7 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
           context: .
+          platforms: linux/amd64,linux/arm64
 
       # copypasta from https://github.com/imjasonh/setup-crane/blob/main/action.yml
       - name: Set up crane
@@ -91,7 +95,6 @@ jobs:
           PATH=${PATH}:${tmp}
           echo "${tmp}" >> $GITHUB_PATH
           echo "${{ github.token }}" | crane auth login ghcr.io --username "dummy" --password-stdin
-
 
       - name: Tag and push
         # For releases, we specifically do _not_ want to rebuild, just tag the
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ ARG LIBKMSP11_VERSION=1.6
 #------------------------------------------------------------------------------
 # Base Debian Image
 #------------------------------------------------------------------------------
-FROM --platform=linux/amd64 debian:bookworm as base
+FROM debian:bookworm AS base
 ARG GO_VERSION
 
 ENV DEBIAN_FRONTEND='noninteractive' \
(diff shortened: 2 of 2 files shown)


.github/workflows/deploy.yaml
```diff
@@ -73,6 +76,7 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
           context: .
+          platforms: linux/amd64,linux/arm64
```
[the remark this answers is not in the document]
jmhodges: linux/arm64 is what you want! The OS half is the actual OS the container is running

.github/workflows/deploy.yaml
```diff
@@ -20,6 +20,9 @@ jobs:
       - name: Clone repository
         uses: actions/checkout@v4
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
```
say-yawn, line 25: I think this is [optional if we want to support more platforms](https://github.com/marketplace/actions/docker-setup-buildx#usage) and [it's recommended](https://docs.docker.com/build/building/multi-platform/#qemu) we use [multiple native nodes](https://docs.docker.com/build/building/multi-platform/#multiple-native-nodes) or [cross-compilation](https://docs.docker.com/build/building/multi-platform/#cross-compilation).

It seems the [cross-compilation](https://docs.docker.com/build/building/multi-platform/#cross-compilation) might be better here and it was what we were attempting but having problems. Wondering if:
> Set the GOOS and GOARCH environment variables to the values of TARGETOS and TARGETARCH. The Go compiler uses these variables to do cross-compilation.

would have resolved that ([example code](https://docs.docker.com/build/building/multi-platform/#cross-compiling-a-go-application))?

say-yawn: Oh just checked the errors and they are:
```
 5 warnings found (use docker --debug to expand):
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 59)
 - FromPlatformFlagConstDisallowed: FROM --platform flag should not use constant value "linux/amd64" (line 7)
 - JSONArgsRecommended: JSON arguments recommended for CMD to prevent unintended behavior related to OS signals (line 86)
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 7)
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 41)
```

` - FromPlatformFlagConstDisallowed: FROM --platform flag should not use constant value "linux/amd64" (line 7)`
It's complaining because we should be using pre-defined `BUILDPLATFORM` variable

jmhodges: Considered it, but it would require us to also be using arm64 libraries and os in the container itself. 

First, we'd have to build `libkmsp11` ourselves. The pre-built binary library [they release has only amd64 builds](https://github.com/GoogleCloudPlatform/kms-integrations/releases/tag/pkcs11-v1.6).  Building it ourself would mean having to the team learn bazel and good bit more code in the Dockerfile. 

Second, even though we don't have many other libraries that we build by hand, we don't actually run in arm64 in production right now. So, we'd have a different arch in our images on our local machines than what we use in prod.

Third, even with a pre-built arm library, we'd have to special case the dockerfile to work with it.

You didn't mention this, but the docs the qemu stuff is supposed to slow us down, but I don't actually see that when comparing this PR's [build image step](https://github.com/mozilla-services/autograph/actions/runs/11727864782/job/32672286685) vs [a previous pr's](https://github.com/mozilla-services/autograph/actions/runs/11733018974/job/32686436697). 

So, I think we should stick with the qemu stuff until we have a good reason to move to arm64 (and there are some energy considerations we could make in the future! But we'd need some arm64 nodepools in kubernetes)

say-yawn: Cool, thanks for the explanation. Yes, I did read the recommendation came with dealing with the slowness, but your other reasons and there being no build time difference makes sense to move forward :shipit:

say-yawn (approved):
Tested my local (MacOS) build works and warnings have been removed :shipit:

PR #1070, opened 2024-11-25
migrate to golang:bookworm image

jmhodges:
This allows us to keep up to date with recent Go releases and security
patches while reducing our configuration in the Dockerfile. We'll bump
the Go version in a separate change.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,11 @@
-ARG GO_VERSION=1.22
 ARG LIBKMSP11_VERSION=1.6
 
 #------------------------------------------------------------------------------
 # Base Debian Image
 #------------------------------------------------------------------------------
-FROM debian:bookworm AS base
-ARG GO_VERSION
+FROM golang:1.22.1-bookworm AS base
 
-ENV DEBIAN_FRONTEND='noninteractive' \
-    PATH="${PATH}:/usr/lib/go-${GO_VERSION}/bin:/go/bin" \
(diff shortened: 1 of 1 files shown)


oskirby:
FWIW, I have noticed from other Go projects I have worked on - this does increase the size of the docker image by several hundred MBs as the official golang images are way bigger than the base debian images.

oskirby:
I stand corrected.

Also, wow those are some big images.

oskirby (approved):

jmhodges:
@oskirby My data was bad because I had a dirty workspace and, as you pointed out in Slack, we have the thing that `ADD` all of the code to it.

With a clean working space, the `debian` comes out ahead by about 100MB or more.

```
$  ls -lh *.tar                                  
-rw-------@ 1 jhodges  staff   507M Nov 25 09:40 debimage.tar
-rw-------@ 1 jhodges  staff   609M Nov 25 09:42 golangimage.tar
```

```
$  docker images | egrep '(debimage|golangimage)'
golangimage   latest    c257b106e395   About a minute ago   2.55GB
debimage      latest    347d368c9a08   3 minutes ago        2.14GB
```

jmhodges:
The ability to use a newer Go is worth the extra 100MB to me.

oskirby:
> The ability to use a newer Go is worth the extra 100MB to me. (Security updates and features, both)

I think that's fine. Yeah

PR #1074, opened 2024-12-05
AUT-371 - Adjusting db locking strategy for concurrent end-entity operations

alexcottner:
[AUT-371](https://mozilla-hub.atlassian.net/browse/AUT-371)

The `TestConcurrentEndEntityOperations` test is flaky, but it's difficult to reproduce this. The cause appears to be a race condition with how we're locking the DB tables and checking for competing processes. Making it possible for different threads/processes/containers to hit the tables at nearly the same time.

Adjusting the tables we're locking and the lock types to try and adjust how this is handled.

```
queries_test.go:53: successfully returned a single label ["1733504987947077173"] for all 373 goroutines
```

This PR:

1. Significantly bumps up the number threads we hit the DB with to increase chances of having a conflict
2. Locks the `endentities` table, removes the need for `endentities_lock` table
3. Removes the `endentities` truncate post-test step as there is only one test hitting that table but other tests are finishing in parallel and truncating the table for us.

Created [AUT-376](https://mozilla-hub.atlassian.net/browse/AUT-376) as a follow-up ticket.

[AUT-371]: https://mozilla-hub.atlassian.net/browse/AUT-371?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AUT-376]: https://mozilla-hub.atlassian.net/browse/AUT-376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


database/queries.go
```diff
@@ -25,7 +25,9 @@ func (db *Handler) BeginEndEntityOperations() (*Transaction, error) {
 		return nil, err
 	}
 	// lock the table
-	_, err = tx.Exec("LOCK TABLE endentities_lock IN ACCESS EXCLUSIVE MODE")
+	_, err = tx.Exec(`LOCK TABLE endentities IN EXCLUSIVE MODE;
+	LOCK TABLE endentities_lock IN EXCLUSIVE MODE;
```
alexcottner, line 29: Locking both tables, because they are both important for this.
But still allowing reads to happen. See [pg docs](https://www.postgresql.org/docs/current/explicit-locking.html)

database/queries.go
```diff
@@ -25,7 +25,9 @@ func (db *Handler) BeginEndEntityOperations() (*Transaction, error) {
 		return nil, err
 	}
 	// lock the table
-	_, err = tx.Exec("LOCK TABLE endentities_lock IN ACCESS EXCLUSIVE MODE")
+	_, err = tx.Exec(`LOCK TABLE endentities IN EXCLUSIVE MODE;
```
jmhodges, line 28: This is dope. How hard would it be to feature flag this just in case we need to roll it back?

database/queries.go
jmhodges: Is there a reason this has to go into the same statement?

database/queries.go
alexcottner: It would just be a env var and an if statement to do the old way vs this way. Is that worth doing for this or would we rather just roll back to the previously deployed container?

database/queries.go
alexcottner: Honestly I thought about grouping all the SQL statements in this function together just because they all either need to happen together or fail anyway. 

This is mostly a habit from much more data driven projects. Less trips to the database is better, but this isn't exactly a bottleneck for us.

main_test.go
```diff
@@ -51,17 +50,6 @@ func newTestAutographer(t *testing.T) (*autographer, configuration) {
 	}
 
 	t.Cleanup(func() {
-		host := database.GetTestDBHost()
```
jmhodges, line 54: This is unrelated change and I worry about having to put this right back after we add more tests.

database/queries.go
jmhodges: Word, in offline convo, we talked about you wanting to wait on dropping the use of endentities_lock because you would want to remove it from the schema immediately but I'm being a jerk and asking that we just do this now and drop the database table in this jira ticket [AUT-374](https://mozilla-hub.atlassian.net/browse/AUT-374)

So, please remove the use of endentities_lock

jmhodges (changes requested):

database/queries.go
alexcottner: Adjusted to remove `endentiteis_lock`

database/queries.go
```diff
@@ -25,22 +25,13 @@ func (db *Handler) BeginEndEntityOperations() (*Transaction, error) {
 		return nil, err
 	}
 	// lock the table
-	_, err = tx.Exec("LOCK TABLE endentities_lock IN ACCESS EXCLUSIVE MODE")
+	_, err = tx.Exec(`LOCK TABLE endentities IN EXCLUSIVE MODE;`)
```
alexcottner, line 28: Allowing reads while blocking writes

database/queries_test.go
```diff
@@ -28,7 +28,7 @@ func TestConcurrentEndEntityOperations(t *testing.T) {
 	defer db.Close()
 
 	var wg sync.WaitGroup
-	concurrency := 73 // took a long time to pick that number!
+	concurrency := 373 // took a long time to pick that number!
```
alexcottner, line 31: Massively increase parallelism to increase race condition chance

database/schema.sql
```diff
@@ -15,13 +15,3 @@ ALTER TABLE endentities ADD CONSTRAINT endentities_unique_label UNIQUE (label);
 GRANT SELECT, INSERT ON endentities TO myautographdbuser;
 GRANT UPDATE (is_current) ON endentities TO myautographdbuser;
 GRANT USAGE ON endentities_id_seq TO myautographdbuser;
-
-CREATE TABLE endentities_lock(
-      id          SERIAL PRIMARY KEY,
```
alexcottner, line 20: No longer required

signer/contentsignaturepki/contentsignature.go
```diff
@@ -153,17 +153,26 @@ func (s *ContentSigner) initEE(conf signer.Configuration) error {
 			break
 		default:
 			// some other error popped up, exit
+			if tx != nil {
+				tx.Rollback()
```
alexcottner, line 157: don't leave transactions hanging

PR #1076, opened 2024-12-07
upgrade statsd: github.com/DataDog/datadog-go to v4.8.3+incompatible from v3.7.2+incompatible

jmhodges:
This doesn't quite get us to a version of datadog-go that fully
supported Go modules (see the "+incompatible" suffix), but this bump
gets us closer.

Inspecting <https://pkg.go.dev/github.com/DataDog/datadog-go> tells us
there's a v5 version we should be importing:
<https://pkg.go.dev/github.com/DataDog/datadog-go/v5>. We can do that
after performing this upgrade and seeing it work well.

This change was made by running:

```
$ go get github.com/DataDog/datadog-go@latest
$ go mod tidy
$ go mod vendor
```


say-yawn (approved):
Changes LGTM but quick questions:

- Why was this update separate from dependabot updates?
- Why was this package behind on being updated?

jmhodges:
> Changes LGTM but quick questions:
> 
>     * Why was this update separate from dependabot updates?
> 
>     * Why was this package behind on being updated?

@say-yawn A previous maintainer had explicitly closed and pause the updates for this dependency back in 2021 in https://github.com/mozilla-services/autograph/pull/803 . That pause was done because they were not software engineers and couldn't vet the dependency. I've been slow-rolling version bumps. 

This PR will reopen this line of deps to be automatically updated.

You can see all the things we're behind on by running:

```
go get -u
go mod tidy
```

PR #1078, opened 2024-12-09
count signer requests with some useful tags

jmhodges:
This patch records what signers are used, and by whom. This is useful
for a variety of reasons including seeing patterns of problems in
roughly real time and tracking usage of signers.

It also includes whether the user used their "default" signer id
(`keyid` in the API itself). We're trying to move folks away from using
the default signer id and requiring `keyid` to be set. We want that
because we want our clients to be more easily moved from one signer to
another as our keys and algorithm requirements change. (Folks had to
change their code during the add-ons and content-signature root
certificate succession instead of just changing some configs because
they relied on the default signer id.)

Updates AUT-206

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -148,6 +148,7 @@ func (a *autographer) handleSignature(w http.ResponseWriter, r *http.Request) {
 		httpError(w, r, http.StatusBadRequest, "failed to parse request body: %v", err)
 		return
 	}
+
 	for i, sigreq := range sigreqs {
 		if r.URL.RequestURI() == "/sign/files" {
 			if sigreq.Input != "" {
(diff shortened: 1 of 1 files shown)


handlers.go
```diff
@@ -215,6 +216,8 @@ func (a *autographer) handleSignature(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 		requestedSignerConfig := requestedSigner.Config()
+		a.stats.Incr("signer.requests", []string{"keyid:" + requestedSignerConfig.ID, "user:" + userid, "request_id:" + rid, usedDefaultSignerTag(sigreq)}, 1.0)
```
jbuck, line 219: What is the cardinality of each tag? I think `keyid` and `user` will be a limited list (because it's just the value from the config file, right) but I think `request_id` will blow up influx storage and prometheus costs.

jbuck (changes requested):

handlers.go
say-yawn: Combining the signer and user ID it's [38 on production](https://lookerstudio.google.com/u/0/reporting/69aab939-4946-4157-b33c-252c82780cb8/page/p_xpwq5dyxmd/edit). Individually it's 32 user ID and 37 signer ID. 

@jbuck that's a good point about `request_id`.

say-yawn (approved):

handlers.go
jmhodges: Yeah, I was thinking of tracing systems, not metric systems, when adding that request_id. Removed.

jbuck (approved):

PR #1079, opened 2024-12-10
bump Go min version to 1.23.4

jmhodges:
This bumps the required Go version to 1.23.4 to take advantage of new
APIs and language features. See https://golang.org/doc/go1.23 for more
information.

Along the way, we use some of the new APIs to simplify some code in
makecsr.


alexcottner (approved):
:call_me_hand:

PR #1084, opened 2024-12-12
monitor: add ability to ignore expiration of signer ids

jmhodges:
This patch adds an env var of comma-separated signer ids to the monitor.
The signer ids in the env var will have any expiration date violations
ignored. In the monitor, the expiration violations include the
expiration being close to the current date.

We do this to clear out the monitor errors returned for the
pre-cert-succession signer ids for the remote-settings service. The
remote-settings service won't flip over until mid-January, but we'd like
to have clear monitors before then.

I went back and forth on using an env var or a command-line flag. I
could go either way.

Updates AUT-345

### DIFF
--- a/tools/autograph-monitor/monitor.go
+++ b/tools/autograph-monitor/monitor.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"encoding/pem"
 	"errors"
+	"flag"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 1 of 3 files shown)


say-yawn (approved):
I prefer env var because it feels more explicit and can be tracked in the config files. TIL about `slices`.

tools/autograph-monitor/monitor.go
```diff
@@ -65,6 +71,16 @@ func main() {
 		log.Fatal("AUTOGRAPH_KEY must be set to the api monitoring key")
 	}
 
+	ignoredExpSignerIdsVar := os.Getenv("IGNORE_EXPIRATION_SIGNER_IDS")
```
say-yawn, line 74: IMO best practices for adding new env var in code is to always add the env in the corresponding config file, even if it's not used. Will you create corresponding ENV var with empty value in the autograph-config repo?

PR #1086, opened 2024-12-20
Adding more details to makecsr docs with a very practical example

alexcottner:


tools/makecsr/README.md
```diff
@@ -50,3 +50,47 @@ Note that the `label` must match between the two configuration files.
...
+
+2. A crypto11-config.json file like this
+```
+{
+  "Path": "/app/libkmsp11.so",
+  "TokenLabel": "gcp-autograph-token"
+}
+```
+
+3. Be authenticated with GCP. Ex: `gcloud auth login --update-adc`
```
say-yawn, line 73: (suggestion) I think we need to add a step here about giving the permission to the developer to access and use the key?

alexcottner: Yeah, pushed that info to the top. Also revised the points to read better.

bhearsum (approved):

PR #1087, opened 2024-12-20
upgrade sops to v3.9.3 from v3.5.0

jmhodges:
This upgrades sops from v3.5.0 to v3.9.3.

A variety of corrections and improvements have happened since the sops
version we've been shipping on. The most obvious change is the new Go
modules package path.

A number of deprecated, archived, and unused transitive dependencies
were also removed by this upgrade. Seeing autograph stuck on those
unmaintained dependencies (like the v1 aws libraries) through the sops
dependency was the inspiration for this. New transitive deps were added,
per usual.

This patch was created by replacing the `go.mozilla.org/sops/v3` imports
with the new `github.com/getsops/sops/v3` imports in main.go, and then
running:

```
$ go mod tidy
$ go mod vendor
```


alexcottner:
:dizzy: Woah.

jmhodges:
We could go with just doing the upgrade to 3.7.3 (which is the last of the `go.mozilla.org/sops/v3` line) first if this is too large.

alexcottner (approved):

PR #1090, opened 2025-01-08
AUT-291 - expanding makecsr options

alexcottner:
[AUT-291](https://mozilla-hub.atlassian.net/browse/AUT-291)
I ran into some missing options and hard-coded values in `makecsr` that we need to be able to adjust for firefox ([link](https://searchfox.org/mozilla-central/rev/7d1b5c88343879056168aa710a9ee743392604c0/taskcluster/kinds/repackage-msix/kind.yml#104)).

Expanding options for `makecsr` and updating read-me to match.

[AUT-291]: https://mozilla-hub.atlassian.net/browse/AUT-291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tools/makecsr/makecsr.go
```diff
@@ -44,9 +48,13 @@ func main() {
 	allowedSigNames := slices.Collect(maps.Keys(allowedSigAlgs))
 
 	flag.StringVar(&crypto11ConfigFilePath, "crypto11Config", "crypto11-config.json", "Path to the crypto11 configuration file")
-	flag.StringVar(&keyLabel, "l", "mykey", "Label of the key in the HSM")
+	flag.StringVar(&keyLabel, "lbl", "mykey", "Label of the key in the HSM")
+	flag.StringVar(&o, "o", "Mozilla Corporation", "Organization of the Subject")
 	flag.StringVar(&ou, "ou", "Mozilla AMO Production Signing Service", "OrganizationalUnit of the Subject")
 	flag.StringVar(&cn, "cn", "Content Signing Intermediate", "CommonName of the Subject")
+	flag.StringVar(&c, "c", "Country", "Country of the Subject")
```
jmhodges, line 55: It seems like these `c`, `st`, and `l` should be empty by default. I don't understand what was problematic about the prior values that are solved by this change.

This PR would also be clearer if its description talked about what actually was problematic in makecsr and how this fixes that.

I know that `mykey` is defaulted weird, but I think that was a mistake by the prior authors. In that case, I think it was a mistake because it made it impossible to set a requirement that the flag be set.

jmhodges (changes requested):

tools/makecsr/makecsr.go
alexcottner: I can add blank defaults. Our problem is the hard coded checks ([link](https://searchfox.org/mozilla-central/rev/7d1b5c88343879056168aa710a9ee743392604c0/taskcluster/kinds/repackage-msix/kind.yml#104)) that are made in the firefox source code.

jmhodges: Update the commit with that info when you merge (and preferably the PR summary), please

jmhodges: With that knowledge, do we need to write down somewhere how to correctly use makecsr to make new authenticode certs somewhere? (makecsr really was for a niche use previously)

jmhodges (approved):

PR #1095, opened 2025-01-14
add direct-to-prometheus metric for signer requests

jmhodges:
We could have misunderstood what the problem is by looking at a metric
that only changed before the first scrape.

Updates AUT-393
Updates AUT-199

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -17,9 +17,11 @@ import (
 	"net/url"
 	"os"
 	"path"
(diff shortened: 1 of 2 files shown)


say-yawn (approved):

jbuck (approved):

say-yawn (approved):
Was reading about `NewCounterVec`. LGTM

PR #1097, opened 2025-01-16
add prometheus metrics for our http handler

jmhodges:
We're migrating from statsd to prometheus and this is one step in it.
This also gives us the chance to correct how we export our API metrics
from having them rolled up by name to have them rolled up by label, as
promql (used by yardstick and Google Metrics Explorer) and prometheus
wants.

We also give up having the success rate's denominator be based on the
requests that are in flight and move to using only the ones returned and
then use the loadbalancer's view of the success rate plus the in-flight
requests to detect requests that are stuck.

(We also need to set up some more request timeouts, but that's not this
patch.)

Updates AUT-199

### DIFF
--- a/main.go
+++ b/main.go
@@ -43,6 +43,7 @@ import (
 
 	"github.com/DataDog/datadog-go/v5/statsd"
 	"github.com/mozilla-services/autograph/crypto11"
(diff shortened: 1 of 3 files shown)


jmhodges:
This is handed off to @alexcottner

PR #1098, opened 2025-01-16
move profilers to debug server

jmhodges:
This makes it more plausible for us to run the profilers in a deployment
setting but also puts the debug stuff where the debug stuff goes.


alexcottner (approved):
Seems reasonable to me

PR #1102, opened 2025-01-23
AUT-199 - Replacing statsd with prometheus. First pass.

alexcottner:
[AUT-199](https://mozilla-hub.atlassian.net/browse/AUT-199)
This PR replaces statsd with prometheus. 

This is a first draft and I ran with some ideas. We can definitely change things.

[AUT-199]: https://mozilla-hub.atlassian.net/browse/AUT-199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


authorize.go
```diff
@@ -47,19 +47,17 @@ func (a *autographer) authorizeHeader(r *http.Request) (auth *hawk.Auth, userid
 		return nil, "", fmt.Errorf("missing Authorization header")
 	}
 	auth, err = hawk.ParseRequestHeader(r.Header.Get("Authorization"))
-	sendStatsErr := a.stats.Timing("hawk.header_parsed", time.Since(getRequestStartTime(r)), nil, 1.0)
-	if sendStatsErr != nil {
-		log.Warnf("Error sending hawk.header_parsed: %s", sendStatsErr)
-	}
+	signerRequestsTiming.With(prometheus.Labels{
+		"step": "hawk_header_parsed",
+	}).Observe(float64(time.Since(getRequestStartTime(r)).Milliseconds()))
```
alexcottner, line 52: Sticking with milliseconds

main.go
```diff
@@ -493,14 +471,14 @@ func (a *autographer) addSigners(signerConfs []signer.Configuration) error {
 		}
 		sids[signerConf.ID] = true
 		var (
-			s           signer.Signer
-			statsClient *signer.StatsClient
-			err         error
+			s signer.Signer
+			// statsClient *signer.StatsClient
```
say-yawn, line 475:
```suggestion
```

main.go
```diff
@@ -493,14 +471,14 @@ func (a *autographer) addSigners(signerConfs []signer.Configuration) error {
...
+			err error
 		)
-		statsClient, err = signer.NewStatsClient(signerConf, a.stats)
-		if statsClient == nil || err != nil {
-			return fmt.Errorf("failed to add signer stats client %q or got back nil statsClient: %w", signerConf.ID, err)
-		}
+		// statsClient, err = signer.NewStatsClient(signerConf)
+		// if statsClient == nil || err != nil {
+		// 	return fmt.Errorf("failed to add signer stats client %q or got back nil statsClient: %w", signerConf.ID, err)
+		// }
```
say-yawn, line 481:
```suggestion
```

main.go
say-yawn: (suggestion) Add TODO here to add metrics on config. [AUT-107](https://mozilla-hub.atlassian.net/browse/AUT-170) is probably the appropriate ticket.

main.go
```diff
@@ -244,13 +230,6 @@ func run(conf configuration, listen string, debug bool) {
 		router.PathPrefix(prefix).Handler(http.StripPrefix(prefix, http.FileServer(http.Dir(parsedURL.Path))))
 	}
 
-	go func() {
-		time.Sleep(5 * time.Minute)
-		ag.stats.Incr(foobarTestCounterName, []string{"statsd:yes"}, 1)
-		foobarTestCounter.Inc()
-		promOnlyFoobarTestCounterName.Inc()
-	}()
-
```
say-yawn, line 253: (note) This might be a hack we might end up using when sending metrics about config.

stats.go
```diff
@@ -5,115 +5,101 @@ import (
...
-			// to have these on a dashboard as a double check, though.
-			w.stats.Incr(fmt.Sprintf("%s.response.success", w.metricPrefix), nil, 1)
-		case statusCode >= 500 && statusCode < 600:
-			w.stats.Incr(fmt.Sprintf("%s.response.status.5xx", w.metricPrefix), nil, 1)
+		responseStatusCounter.With(prometheus.Labels{
+			"handler":    w.handlerName,
+			"statusCode": fmt.Sprintf("%d", statusCode),
+		}).Inc()
+
+		if statusCode >= 200 && statusCode < 300 || statusCode >= 400 && statusCode < 500 {
```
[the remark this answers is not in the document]
alexcottner: That's what the old logic had and I carried that over here. I think the logic is/was: a 4xx response means the client did something wrong and not the service (bad request, bad auth, etc).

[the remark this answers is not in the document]
say-yawn: Thanks, I see what you mean by that from the comment here https://github.com/mozilla-services/autograph/blob/e1d83310dc30b57286ce8acecb80ff9529bd435b/stats.go#L94-L99
I think it would be better to separate this into another metric then. How about `"status": "authentication_failure"` and keep the comment?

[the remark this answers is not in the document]
say-yawn: Thanks, I see what you mean by that from the comment here https://github.com/mozilla-services/autograph/blob/e1d83310dc30b57286ce8acecb80ff9529bd435b/stats.go#L94-L99
I think it would be better to separate this into another metric then. How about `"status": "authentication_failure"` and keep the comment?

main.go
say-yawn: Missed another comment here.

stats.go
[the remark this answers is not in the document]
alexcottner: Yeah that seems fine, as long as we aren't counting it against ourselves.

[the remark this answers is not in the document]
alexcottner: Gonna change it to "client_failure" since it's likely authentication but could be a 404 or bad verb or something.

say-yawn (approved):

PR #1112, opened 2025-02-19
Fix unit tests command in README

leplatrem:


alexcottner (approved):
Pull in main and you should be good to go

PR #1113, opened 2025-02-19
Fixing our unit tests with the expired cert.

alexcottner:
We have some unit tests that use a certificate that is now expired, causing them to fail.
Updating the unit tests with the updated certificate chains.
Since we don't have a great way to fake time in golang, we may need to do this again.


say-yawn (approved):

tools/autograph-monitor/contentsignaturepki_test.go
```diff
@@ -599,19 +599,19 @@ func Test_certChainValidityNotifications(t *testing.T) {
 			},
 			wantNotifications: []*CertNotification{
 				&CertNotification{
-					CN:       "Content Signing Intermediate/emailAddress=foxsec@mozilla.com",
+					CN:       "Content Signing Intermediate",
 					Severity: "info",
-					Message:  `Certificate 0 "Content Signing Intermediate/emailAddress=foxsec@mozilla.com" is valid from 2020-12-31 00:00:00 +0000 UTC to 2025-03-14 22:53:57 +0000 UTC`,
+					Message:  `Certificate 0 "Content Signing Intermediate" is valid from 2015-03-17 00:00:00 +0000 UTC to 2101-05-19 00:00:00 +0000 UTC`,
```
say-yawn, line 604:
```suggestion
					Message:  `Certificate 0 "Content Signing Intermediate" is valid from 2015-03-17 00:00:00 +0000 UTC to 2200-05-19 00:00:00 +0000 UTC`,
```
Any reason why this one has shorter period than the others?

alexcottner: The intermediate cert does actually expire on 2101-05-19. Unsure why that was picked.

PR #1115, opened 2025-02-21
Use gauge metrics on requests and response

say-yawn:
According to the co-founder of Prometheus counters should not be used as an absolute value as it does not account for when the metric is reset ([video](https://youtu.be/fhx0ehppMGM?t=237)):

> Coming to PromQL again, you'll almost never want to look at the absolute value of a counter. If you think about it, counters are just cumulative counts over some arbitrary amount of time that totally depends on when the counter start from zero ... And the absolute value doesn't tell you anything useful at all.

So I'm testing the data with gauge to see it handles the metric reset better. I'm also [reading](https://iximiuz.com/en/posts/prometheus-functions-agg-over-time/) that `sum_over_time()` works better with gauges which is what we want to do instead of going sampling the sum of the increase (instant vector) which is what `sum by(namespace, keyid, user) (autograph_signer_requests{namespace=~"autograph-$env"}[$__rate_interval])` seem to be doing.

Fix AUT-393

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -221,6 +221,12 @@ func (a *autographer) handleSignature(w http.ResponseWriter, r *http.Request) {
 			// TODO(AUT-206): remove this when we've migrate everyone off of the default keyid
 			"used_default_signer": strconv.FormatBool(sigreq.KeyID == ""),
 		}).Inc()
+		signerRequestsGauge.With(prometheus.Labels{
+			"keyid": requestedSignerConfig.ID,
+			"user":  userid,
+			// TODO(AUT-206): remove this when we've migrate everyone off of the default keyid
+			"used_default_signer": strconv.FormatBool(sigreq.KeyID == ""),
+		}).Inc()
 
 		sigresps[i] = formats.SignatureResponse{
 			Ref:        id(),
--- a/stats.go
+++ b/stats.go
@@ -18,8 +18,20 @@ var (
 		Help:      "A counter for how many requests are made to a given handler",
 	}, []string{"handler"})
 
+	requestGauge = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name:      "request_gauge",
+		Namespace: statsNamespace,
+		Help:      "A counter for how many requests are made to a given handler",
+	}, []string{"handler"})
+
 	signerRequestsCounter = promauto.NewCounterVec(prometheus.CounterOpts{
 		Name:      "signer_requests",
+		Namespace: statsNamespace,
+		Help:      "A counter for how many authenticated and authorized requests are made to a given signer",
+	}, []string{"keyid", "user", "used_default_signer"})
+
+	signerRequestsGauge = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name:      "signer_requests_gauge",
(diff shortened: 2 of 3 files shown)


alexcottner:
I think I'm missing how this would be different from our current counter metrics. We don't reset the counters to 0 outside of unit tests, right?

say-yawn:
@alexcottner not anywhere in the app but as you mentioned in Slack
> Whenever an updated deployment goes out, or a rollout-restart happens, new pods start up with new vectors starting at zero.

these metrics reset at 0 and the `increase()` does not handle when the metric is first emitted after the reset nor does it count how many time the reset happened. So as an example, this [widevine signer request logs ](https://cloudlogging.app.goo.gl/cqsoYnG9xmExkucU6)show that the particular pod was hit 7 times but the counter shows there were only 4 events ([Metric Explorer](https://console.cloud.google.com/monitoring/metrics-explorer;endTime=2025-02-12T18:11:40.957Z;startTime=2025-02-12T16:11:40.957Z?pageState=%7B%22domainObjectDeprecationId%22:%2293870983-F56F-439F-853E-9A8E4503AE4C%22,%22title%22:%22Untitled%22,%22xyChart%22:%7B%22constantLines%22:%5B%5D,%22dataSets%22:%5B%7B%22plotType%22:%22LINE%22,%22prometheusQuery%22:%22autograph_signer_requests%7Bkeyid%3D%5C%22widevine_rel1%5C%22,%20pod%3D%5C%22autograph-7745d6f97c-zk294%5C%22%7D%22,%22targetAxis%22:%22Y1%22,%22unitOverride%22:%22%22%7D%5D,%22options%22:%7B%22mode%22:%22COLOR%22%7D,%22y1Axis%22:%7B%22label%22:%22%22,%22scale%22:%22LINEAR%22%7D%7D%7D&project=moz-fx-webservices-high-prod)/[Yardstick](https://yardstick.mozilla.org/explore?schemaVersion=1&panes=%7B%22ka1%22%3A%7B%22datasource%22%3A%22adpvtjmrxoc1sb%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22autograph_signer_requests%7Bkeyid%3D%5C%22widevine_rel1%5C%22%2C+pod%3D%5C%22autograph-7745d6f97c-zk294%5C%22%7D%22%2C%22range%22%3Atrue%2C%22instant%22%3Atrue%2C%22datasource%22%3A%7B%22type%22%3A%22prometheus%22%2C%22uid%22%3A%22adpvtjmrxoc1sb%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22legendFormat%22%3A%22__auto%22%2C%22useBackend%22%3Afalse%2C%22disableTextWrap%22%3Afalse%2C%22fullMetaSearch%22%3Afalse%2C%22includeNullMetadata%22%3Atrue%7D%5D%2C%22range%22%3A%7B%22from%22%3A%221739376902067%22%2C%22to%22%3A%221739384810659%22%7D%7D%7D&orgId=1)).

Also, the general recommendation on counter type metrics is to use rate/increase instead of checking the total number hence the decimal point requests instead of whole number of requests.

PR #1130, opened 2025-03-25
AUT-433 - Adjusting signer config load so a bad signer doesn't cause startup failure

alexcottner:
[AUT-433](https://mozilla-hub.atlassian.net/browse/AUT-433)
Adjusting the signer config load so that a bad signer (ex: malformed config, missing required properties, expired cert, etc) does not cause the application load to fail. Instead we log an error.

Added unit test to confirm a valid signer still loads as expected.

[AUT-433]: https://mozilla-hub.atlassian.net/browse/AUT-433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/main.go
+++ b/main.go
@@ -482,22 +482,22 @@ func (a *autographer) addSigners(signerConfs []signer.Configuration) error {
 		case contentsignature.Type:
 			s, err = contentsignature.New(signerConf)
 			if err != nil {
-				return fmt.Errorf("failed to add signer %q: %w", signerConf.ID, err)
+				log.Error(fmt.Errorf("failed to add signer %q: %w", signerConf.ID, err))
 			}
 		case contentsignaturepki.Type:
 			s, err = contentsignaturepki.New(signerConf)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


say-yawn (changes requested):
I am concerned that only logging on the error could mean we might silently miss some valid errors that needs to be handled for. Can we also add metrics so we can set some alerts on this OR restrict the error handling to only log for expired or soon to expire signers?

alexcottner:
I'm hesitant to use metrics to alert because of the current prometheus issues we're seeing. Also, we'll lose some information because we won't include the full error in the metric. I am going to create an alert policy though that will forward the logs to us though.

say-yawn (approved):
Approving this in favor for bubbling up the error logs into Slack or Yardstick. Related work is: https://mozilla-hub.atlassian.net/browse/AUT-410

PR #1143, opened 2025-05-05
AUT-449 - Removing AWS lambda dependency

alexcottner:
This PR removes the AWS lambda dependency from autograph monitor, since we no longer need it.
This leaves our only AWS depdencies for the `contentsignaturepki` signer, which it uses to handle S3 files. We could _probably_ slim those down a bit if we wanted in the future.


say-yawn (approved):
Removal looks good. Thanks for creating this cleanup PR.

PR #1146, opened 2025-05-19
AUT-267: add request duration metric

leplatrem:
https://mozilla-hub.atlassian.net/browse/AUT-267

### DIFF
--- a/stats.go
+++ b/stats.go
@@ -27,7 +27,7 @@ var (
 
 	responseDurationHistogram = promauto.NewHistogramVec(prometheus.HistogramOpts{
 		Name:      "request_duration",
-		Help:      "A historygram for how long requests take to process",
+		Help:      "A histogram for how long requests take to process",
 		Namespace: statsNamespace,
 		Buckets:   prometheus.ExponentialBucketsRange(0.01, 6.0, 7), // 8 buckets with +Inf
 	}, []string{"handler"})
--- a/stats.go
+++ b/stats.go
@@ -25,11 +25,15 @@ var (
 		Help:      "A counter for how many requests are made to a given handler",
 	}, []string{"handler"})
 
-	responseDurationHistogram = promauto.NewHistogramVec(prometheus.HistogramOpts{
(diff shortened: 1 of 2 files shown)


stats.go
```diff
@@ -24,6 +25,13 @@ var (
 		Help:      "A counter for how many requests are made to a given handler",
 	}, []string{"handler"})
 
+	responseDurationHistogram = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Name:      "request_duration",
+		Help:      "A histogram for how long requests take to process",
+		Namespace: statsNamespace,
+		Buckets:   prometheus.ExponentialBucketsRange(0.01, 6.0, 7), // 8 buckets with +Inf
+	}, []string{"handler"})
```
alexcottner, line 33: I don't think a histogram will work here. Some endpoints will be consistent for < 1 second, but others will vary between 20-60. I think we should probably do a summary vector on .5, .95, .99 quantiles instead, similar to `signerRequetsTiming`.

leplatrem: In RS we use histogram and then have 50th/95th/99th quantiles agreggations done in Yardstick.

Also, I'm suprised to that we report milliseconds, because I was running into unit issues on yardsticks.
Shall I align this new one and report msec too?

alexcottner (approved):

stats.go
```diff
@@ -153,7 +165,11 @@ func statsMiddleware(h http.HandlerFunc, handlerName string) http.HandlerFunc {
 			"handler": handlerName,
 		}).Inc()
 		w = newStatsWriter(w, handlerName)
+		// Process the request
 		h(w, r)
+		// Report the request duration
+		var durationSeconds = time.Since(getRequestStartTime(r)).Seconds()
```
leplatrem, line 171: Shall we have seconds or milleseconds like in `signer_request_timing`?

alexcottner: I don't think we need millisecond precision for this one, so seconds should be good.

PR #1227, opened 2026-05-01
AUT-305 - Making adjustments for migration out of S3 to a mounted volume for cert-chain storage

alexcottner:
[AUT-305](https://mozilla-hub.atlassian.net/browse/AUT-305)

We are migrating away from S3 as the cert-chain storage layer.
And while I'm here, let's just make this work for any path. 
That way we can mount a volume and this becomes an implementation detail rather than a code detail.
Working with @say-yawn who is handling infra

[AUT-305]: https://mozilla-hub.atlassian.net/browse/AUT-305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


say-yawn (dismissed):
Paired while reviewing code and the fixes were applied synchronously. Will follow-up with autograph config changes before before I merge this.

⚠️ DO NOT MERGE UNTIL AUTOGRAPH-CONFIG CHANGES

say-yawn (approved):

PR #1248, opened 2026-07-28
AUT-452 - Add macOS signing support

hneiva:
Add an "apple" signer that signs macOS artifacts by shelling out to rcodesign (github.com/indygreg/apple-platform-rs). It handles all four artifact types from the /sign/file endpoint, auto-detecting each:

  - standalone Mach-O binaries
  - application bundles, received and returned as a .tar.gz containing a .app (extracted, signed recursively, and re-archived)
  - DMG disk images
  - XAR archives (.pkg installers)

Only signing is implemented; notarization and stapling are handled by a separate system. Signatures are notarization-ready (hardened runtime + secure timestamp) when for_notarization is enabled.

Key handling differs by mode, deliberately:

  - application mode (Mach-O/.app/.dmg) signs with an HSM-resident key via PKCS#11, reusing the provider autograph is already configured with (conf.HSM). The private key never leaves the HSM, and the PIN is passed through a 0400 config file rather than the command line.
  - installer mode (.pkg) signs with an in-memory PEM key. rcodesign's XAR signer does not support PKCS#11 keys (it rejects any HSM key regardless of token/label/id), so .pkg installers must use a software Developer ID Installer key until upstream adds XAR+PKCS#11 support. See the note at the top of signer/apple/apple.go.

Per-artifact signing settings (entitlements, hardened runtime) follow the hardened-sign-config model from Mozilla iscript: signer-level defaults that a request may override, with globs mapped to rcodesign scoped settings.

Bundle extraction is hardened against malicious .tar.gz input: entry and hardlink paths are resolved with filepath-securejoin (clamping symlink and ".." traversal so nothing is written or read outside the work dir), hardlinks are deduplicated by inode when re-archiving, and total size and entry count are capped to bound decompression.

PKCS#11 support exists only on the upstream rcodesign main branch (no released version includes it), so the Dockerfile builds rcodesign from a pinned commit with the pkcs11 cargo feature in a dedicated builder stage and copies the binary into the runtime image. softhsm2 is added to the base image, and the SoftHSM dev config gains apple signers used for integration and by the package tests.


Dockerfile
```diff
@@ -1,5 +1,21 @@
...
+#------------------------------------------------------------------------------
+# rcodesign build stage (Apple code signing tool)
+#------------------------------------------------------------------------------
+# The apple signer shells out to rcodesign (github.com/indygreg/apple-platform-rs).
+# PKCS#11/HSM signing exists only on the upstream main branch — no released
+# version (0.29.0 is newest) or prebuilt binary includes it — so we build from a
+# pinned commit with the pkcs11 feature. It's built here and copied into the
+# final image so the runtime image carries no Rust toolchain. Revisit (switch to
+# a pinned release/prebuilt) once upstream ships a release with PKCS#11.
+FROM rust:1-bookworm AS rcodesign-builder
```
alexcottner, line 20: Can we use trixie and a more specific rust version to reduce breaks? 
Like `1.98-trixie`?

Dockerfile
```diff
@@ -26,7 +42,8 @@ RUN apt update && \
         pkg-config \
         curl \
         jq \
-        libengine-pkcs11-openssl
+        libengine-pkcs11-openssl \
+        softhsm2
```
alexcottner, line 59: Still reviewing, but this breaks our separation a little bit with the `app-hsm` container from `tools/softhsm`.

signer/apple/apple.go
```diff
@@ -0,0 +1,976 @@
...
+	// and total content re-archived by createTarGz.
+	maxUncompressedBundleBytes int64 = 2 << 30 // 2 GiB
+	// maxBundleEntries caps the number of archive entries, bounding inode and
+	// CPU use from an archive of many tiny/empty entries.
+	maxBundleEntries = 200000
+)
+
+// extractTarGz extracts a gzip-compressed tar archive into dest, preserving
+// file modes and symlinks. Every entry location (and every hardlink source) is
+// resolved with filepath-securejoin, which follows existing symlink components
```
alexcottner, line 729: Let's get rid of `securejoin` as a depednecy. Autograph is an internal tool and payloads can only come from authenticated and trusted sources. If we want to be sure we could use `filepath.EvalSymlinks` and `filepath.IsLocal` in a few lines rather than adding a new vendored dependency that we need to keep up with.

signer/apple/apple.go
```diff
@@ -0,0 +1,976 @@
...
+
+	switch conf.Mode {
+	case ModeApplication, ModeInstaller:
+	case "":
+		conf.Mode = ModeApplication
+	default:
+		return nil, fmt.Errorf("apple: invalid mode %q, must be %q or %q", conf.Mode, ModeApplication, ModeInstaller)
+	}
+	s.mode = conf.Mode
+	s.Mode = conf.Mode
```
alexcottner, line 129: Duplicated line?

PR #1249, opened 2026-08-06
Vendoring libkmsp11 and building it in dockerfile so we can support multiplatform builds

alexcottner:
The goal of this PR is to support multi-platform builds by vendoring libkmsp11 and building it in dockerfile.

This PR:

 - Vendors the libkmsp11 v1.9 code in /kmsp11 from the [v1.9 release](https://github.com/GoogleCloudPlatform/kms-integrations/releases/tag/pkcs11-v1.9)
 - Builds libkmsp11 in the prebuild step on the autograph image
 -  Updates our base images to debian trixie
     - Handles some small changes required due to using updated packages
 -  Removes the amd64 platform targeting 

[AUT-638](https://mozilla-hub.atlassian.net/browse/AUT-638)

[AUT-638]: https://mozilla-hub.atlassian.net/browse/AUT-638?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Dockerfile
```diff
@@ -28,29 +28,27 @@ RUN apt-get update && \
...
-
-# fetch the RDS CA bundles
-# https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html#UsingWithRDS.SSL.CertificatesAllRegions
-RUN curl -o /usr/local/share/old-rds-ca-bundle.pem https://s3.amazonaws.com/rds-downloads/rds-combined-ca-bundle.pem && \
-      curl -o /usr/local/share/new-rds-ca-bundle.pem https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem && \
-      cat /usr/local/share/old-rds-ca-bundle.pem /usr/local/share/new-rds-ca-bundle.pem > /usr/local/share/rds-combined-ca-bundle.pem
+# Build the Google KMS library
+RUN echo "Building Google KMS library for ${TARGETARCH}."
+ADD ./kmsp11 /kmsp11
+WORKDIR /kmsp11
```
say-yawn: Why not continue to build using `/tmp` directory?

Dockerfile
```diff
@@ -28,29 +28,27 @@ RUN apt-get update && \
         jq \
         libengine-pkcs11-openssl
 
+RUN wget https://github.com/bazelbuild/bazelisk/releases/download/v1.29.0/bazelisk-${TARGETARCH}.deb
```
say-yawn: (nit) add comment that this is for libkmsp.

tools/configurator/configurator.py
say-yawn, line 1: I'm guessing there was an update to `ruamel` while upgrading to trixie?

say-yawn (changes requested):
First pass. Will look at vendored code now.

tools/configurator/configurator.py
alexcottner: Yeah from 17.21 to 18.10, not a huge difference for us but had to adjust a little.

Dockerfile
alexcottner: I suppose there isn't a real difference here so I can move it back to `/tmp`.

Dockerfile
alexcottner: added

Dockerfile
```diff
@@ -28,25 +28,26 @@ RUN apt update && \
...
 
+## install bazelisk for building libkmsp11
+RUN wget https://github.com/bazelbuild/bazelisk/releases/download/v1.29.0/bazelisk-${TARGETARCH}.deb
+RUN dpkg -i bazelisk-${TARGETARCH}.deb
+RUN rm bazelisk-${TARGETARCH}.deb
+
 COPY google-pkcs12-release-signing-key.pem /app/src/autograph/
 
 # Build the Google KMS library
 RUN echo "Building Google KMS library for ${TARGETARCH}."
```
say-yawn, line 48: While reading some documentation around Bazel, I thought this documentation on [building libraries using the Bezel Docker Container](https://bazel.build/install/docker-container) to be interesting. Not sure if it's something we want to use because it won't be vendored but potential future usage when we stop vendoring code.

say-yawn (approved):
LGTM. I only reviewed the docker related changes. For the vendored kmsp11 and bazel code, I'm using local environment building as a check to see that it works as expected. Other alternatives to consider in the future:
- We could vendor the kmsp11 code for linux/arm64 since we only have engineers using the arm64 or amd64
- Build kmsp11 using the Bezel Docker Container

Dockerfile
alexcottner: That is good to note. For now, I'll get this merged and create a follow-up ticket with these references.

PR #1250, opened 2026-08-11
Build docker images parallel in a matrix

alexcottner:
Looking to improve performance on docker build times.
This PR's goal is to:
1. Build containers in parallel in a matrix
2. Still publish a unified multi-arch manifest for pushes to main and releases

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -7,11 +7,20 @@ on:
   release:
     types:
       - released
+  pull_request:
+    branches: [ "main" ]
 
 jobs:
-  docker:
-    name: Docker Images
-    runs-on: ubuntu-22.04
+  docker-build:
+    name: Docker Build
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-22.04
+            architecture: amd64
+          - os: ubuntu-22.04-arm
+            architecture: arm64
+    runs-on: ${{ matrix.os }}
     environment: build
     permissions:
       contents: read
@@ -19,9 +28,6 @@ jobs:
     steps:
       - name: Clone repository
         uses: actions/checkout@v4
-
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v3
 
       - name: Setup Docker Buildx
         uses: docker/setup-buildx-action@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yaml
```diff
@@ -7,11 +7,20 @@ on:
   release:
     types:
       - released
+  pull_request:
+    branches: [ "main" ]
```
say-yawn: I'm guessing this is temporary code to see that the changes work during the PR stage.

.github/workflows/deploy.yaml
```diff
@@ -7,11 +7,20 @@ on:
...
 
 jobs:
-  docker:
-    name: Docker Images
-    runs-on: ubuntu-22.04
+  docker-build:
+    name: Docker Build
+    strategy:
+      matrix:
```
say-yawn, line 15: 👀 TIL about using matrix for running variations of jobs in GitHub Actions ([doc](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#expanding-or-adding-matrix-configurations))

.github/workflows/deploy.yaml
```diff
@@ -36,9 +42,9 @@ jobs:
           images: |
             ${{ vars.GAR_LOCATION }}-docker.pkg.dev/${{ vars.GCP_PROJECT_ID }}/${{ vars.GAR_REPOSITORY}}/autograph
           tags: |
-            type=sha,format=long,enable=${{ github.event_name == 'push' }}
-            type=raw,value=latest,enable=${{ github.event_name == 'push' }}
-            type=semver,pattern={{raw}},enable=${{ github.event_name == 'release' }}
+            type=sha,format=long,enable=true,suffix=-${{ matrix.architecture }}
```
say-yawn: Will the `enable=true` will stay even after testing? Not sure what having the `enable` to `true` in the tags do and why this was set to true only during push event... since the deploy event should only be triggered during merge to main and new tag releases.

.github/workflows/deploy.yaml
```diff
@@ -143,10 +152,47 @@ jobs:
...
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+
+      - name: Publish latest
+        if: ${{ github.event_name == 'push' }}
+        run: |
+            LATEST_TAG=$(git describe --tags --abbrev=4)
+            git describe --tags --abbrev=4
+            docker manifest create mozilla/autograph:latest \
+              mozilla/autograph:sha-${{ github.sha }}-arm64 \
+              mozilla/autograph:sha-${{ github.sha }}-amd64 \
```
say-yawn, line 156:
```suggestion
              mozilla/autograph:sha-${{ github.sha }}-{{ matrix.architecture }}
```
Wouldn't `docker-manifest` job also run twice because of the matrix?

say-yawn (changes requested):

.github/workflows/deploy.yaml
alexcottner: Correct, I will remove before the merge to main.

.github/workflows/deploy.yaml
alexcottner: Ah this was also just for testing in my PR branch.

.github/workflows/deploy.yaml
alexcottner: Nope, this is in a separate job. `docker-build` has the matrix, but `docker-manifest` does not.

say-yawn (approved):

PR #1257, opened 2026-09-04
AUT-602 - Moving signer config to database and secret manager

alexcottner:
[AUT-602](https://mozilla-hub.atlassian.net/browse/AUT-602)

Related:

- https://github.com/mozilla/webservices-infra/pull/12375

TODO:

- [x] Setup new database schema
- [x] Allow signer config to be read from both DB schema and files, DB should win conflicts
- [x] Adjust test setup to only use DB schema
- [x] Adjust integration tests to only use DB schema
- [x] Add logic to pull optional sensitive signer config from GCP secrets manager
- [x] Add tests for any new business logic
- [x] Update documentation

Follow-ups:
I would like to refactor most of autograph's code. I think we can cut out a good 20% of the code and make the signer config live rather than loading on initial startup.

[AUT-602]: https://mozilla-hub.atlassian.net/browse/AUT-602?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eviljeff:
Can you provide some details of how this patch can be reviewed?  Obviously reading line by line through +9,255-883 lines is essentially impossible.

database/queries.go
```diff
@@ -106,3 +106,32 @@ func (tx *Transaction) End() error {
...
+		) auth
+		from auth a
+		inner join auth_signers a_s on a.id = a_s.auth
+		group by a.id, a.key
+	), signerQuery as (
+		select s.public ||  jsonb_build_object(
+			'id', s.id,
+			'type', s.type,
+			'mode', s.mode,
+			'secret', secret
```
say-yawn:
```suggestion
			'secret', s.secret
```

database/schema.sql
```diff
@@ -25,3 +25,1672 @@ CREATE TABLE endentities_lock(
 );
 GRANT SELECT, INSERT, UPDATE ON endentities_lock TO myautographdbuser;
 GRANT USAGE ON endentities_lock_id_seq TO myautographdbuser;
+
+CREATE TABLE auth(
+      id          VARCHAR(128) PRIMARY KEY,
+      key         VARCHAR(64) NOT NULL,
```
say-yawn: I see that setting the `id` as primary key makes it unique ([doc](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-PRIMARY-KEYS)) but wondering if the key should at least be set to unique. Also, why is the id longer than the key?

say-yawn: Should we also add a column here to deprecate the old auth or prefer deleting?

database/queries_test.go
```diff
@@ -97,3 +98,101 @@ func waitAndMakeEE(j int, db *Handler, wg *sync.WaitGroup, t *testing.T, signerI
...
+		t.Fatal("Should have 7 auth_signers")
+	}
+
+	if len(conf.Signers) != 6 {
+		t.Fatal("Should have 6 signers")
+	}
+
+	for _, s := range conf.Signers {
+		if s.ID == "bad-id" {
+			t.Fatal("The signer id public property should have been overwritten by the record's id")
```
say-yawn: I'm a bit confused by this error message.

main.go
```diff
@@ -140,27 +162,79 @@ func parseArgsAndLoadConfig(args []string) (serviceConf serviceConfig, signerCon
 		log.SetLevel(level)
 		log.Infof("Set logging level to %s", level)
 	}
+	return
```
say-yawn: Why not also early return during positive err checks?

say-yawn: 🤦 I see that during other signer config parsing we won't early return when we get err.

main.go
```diff
@@ -140,27 +162,126 @@ func parseArgsAndLoadConfig(args []string) (serviceConf serviceConfig, signerCon
...
+	// return just dbConf if there is no file to load
+	if signerFile == "" {
+		return &dbConf, nil
+	}
+
+	// otherwise load and merge the two config sources
+	var fileConf signerConfig
+	err = fileConf.loadFromFile(signerFile)
+	if err != nil {
+		return &dbConf, err
```
say-yawn:
```suggestion
		return &fileConf, err
```
If you intended to return `&dbConf` instead can you share why?

main.go
```diff
@@ -140,27 +162,126 @@ func parseArgsAndLoadConfig(args []string) (serviceConf serviceConfig, signerCon
...
 	err = serviceConf.loadFromFile(serviceFile)
 	if err != nil {
 		log.Fatal(err)
 	}
-	err = signerConf.loadFromFile(signerFile)
+
+	return
+}
+
+func loadSignerConfig(db *database.Handler, signerFile string) (*signerConfig, error) {
```
say-yawn: It would be helpful when we early return with `err` we prefix the err so we know where in the checks the err happened.

main.go
```diff
@@ -140,27 +162,126 @@ func parseArgsAndLoadConfig(args []string) (serviceConf serviceConfig, signerCon
...
+	}
+
+	// otherwise load and merge the two config sources
+	var fileConf signerConfig
+	err = fileConf.loadFromFile(signerFile)
+	if err != nil {
+		return &dbConf, err
+	}
+
+	// add any missing authorizations from fileConf to dbConf
```
say-yawn: Can you share if this is actually feasible or this is just one-time check during migration. If it's possible (not ideal) outside the initial migration I would like logs on the missing auths/signers so we are reminded to migrate the missing auths and signers to the DB.

database/schema.sql
```diff
@@ -25,3 +25,1672 @@ CREATE TABLE endentities_lock(
...
+GRANT SELECT, INSERT ON auth TO myautographdbuser;
+
+CREATE TABLE signer (
+      id          VARCHAR(128) PRIMARY KEY,
+      type        VARCHAR(64) NOT NULL,
+      mode        VARCHAR(64) NULL,
+      comments    VARCHAR,
+      created     TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
+      updated     TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
+      secret      VARCHAR(256), -- points to a secret that can be read by application
```
say-yawn: Shouldn't this also be not null?

main.go
```diff
@@ -140,27 +162,79 @@ func parseArgsAndLoadConfig(args []string) (serviceConf serviceConfig, signerCon
...
 	err = serviceConf.loadFromFile(serviceFile)
 	if err != nil {
 		log.Fatal(err)
 	}
-	err = signerConf.loadFromFile(signerFile)
+
+	return
+}
+
+func loadSignerConfig(db *database.Handler, signerFile string) (*signerConfig, error) {
```
say-yawn: Is there a unittest for this function?

database/queries_test.go
say-yawn: Is there a linter you are using that's changing these white spaces?

main.go
```diff
@@ -234,6 +238,36 @@ func loadSignerConfig(db *database.Handler, signerFile string) (*signerConfig, e
 	return &dbConf, nil
 }
 
+func loadSignerSecrets(gcpProjectId string, signerConf signerConfig) error {
+	var errs []error
+	for i := range signerConf.Signers {
+		if signerConf.Signers[i].Secret == "" || signerConf.Signers[i].SecretLoaded {
+			continue
+		}
```
say-yawn: Are there signers where there is no secret value?

main.go
```diff
@@ -234,6 +238,36 @@ func loadSignerConfig(db *database.Handler, signerFile string) (*signerConfig, e
...
+		if val, ok := secret["privatekey"]; ok {
+			signerConf.Signers[i].PrivateKey = val
+		}
+		if val, ok := secret["passphrase"]; ok {
+			signerConf.Signers[i].Passphrase = val
+		}
+		if val, ok := secret["issuerprivkey"]; ok {
+			signerConf.Signers[i].IssuerPrivKey = val
+		}
+		signerConf.Signers[i].SecretLoaded = true
```
say-yawn: Just in case we add a new secret that's not `privatekey`, `passphrase`, or `issuerprivkey` can we add a log and a way to handle it?

main.go
```diff
@@ -140,27 +162,126 @@ func parseArgsAndLoadConfig(args []string) (serviceConf serviceConfig, signerCon
...
+	dbConf.Signers = append(dbConf.Signers, signersToAdd...)
+
+	return &dbConf, nil
+}
+
+func loadSignerSecrets(gcpProjectId string, signerConf signerConfig) error {
+	if gcpProjectId == "" {
+		log.Warn("No gcpProjectId configured. Skipping loadSignerSecrets.")
+		return nil
+	}
```
say-yawn: This check/error logging should be called at `main.go` where `loadSignerSecrets` are called since it already checks that the `gcpProjectId` is not empty.

say-yawn (changes requested):
Was only able to review up to 83cfacf4764fd5599ea6f35228735a148a18f0c8

main.go
alexcottner: We have tests covering `db.GetSignerConfig` and `signerConfig.loadFromFile`. And this is covered by integration tests. But unit a test for `loadSignerConfig` would require mocking all of that IO so I didn't think it was super useful.

main.go
alexcottner: This was intentional. In this case we have failed to load the file config, but we still have the config we loaded from the database. So `dbConf` that is the only valid config object we have (although it may be empty).

main.go
alexcottner: This is just a temporary measure so that we can run signer config in the db and in files side-by-side for 1-2 releases as we migrate them over. We can absolutely add a log statement though to help identify signers/auths that need to be migrated.

main.go
alexcottner: Absolutely, and as time goes on fewer signers will have secret values thanks to better kms/hsm support.

main.go
alexcottner: Oh I double handled this scenario. Good catch.

alexcottner:
> Can you provide some details of how this patch can be reviewed? Obviously reading line by line through +9,255-883 lines is essentially impossible.

Yeah this is part 1 of N for a big tech debt reduction push. From an add-ons perspective, you'll only care about a few lines: https://github.com/mozilla-services/autograph/pull/1257/changes#diff-7da932eaccc702303470a66346796e68c5cb5194b61720038306989dabcd8355

And those are just me wrapping the `time.Duration` type into something that I can deserialize from JSON. @say-yawn will be reviewing all of the changes if you'd like to ignore this one.

When we get into quantum safe stuff in the near future you'll have some real review work to do though. These big refactors will hopefully make that work a lot easier for us.

database/queries_test.go
alexcottner: Just `make fmt-fix`

main.go
say-yawn: These errors are being logged using `Fatalf` and will early return.

database/queries.go
alexcottner: Added in latest commit.

main.go
say-yawn: Resolved in 1028bdc

database/queries_test.go
alexcottner: This is to check that an `id` property in the `public` jsonb column doesn't overwrite the actual ID of the record in response. The base columns are considered sacred to prevent collisions. Otherwise it would be possible for `signer-a` to be overwritten by `signer-b` due to a bad duplicate property.

database/schema.sql
alexcottner: Secret? Secret can definitely be null, like if we have a fully KMS/HSM backed signer.

main.go
alexcottner: Added in latest commit.

main.go
```diff
@@ -171,32 +299,44 @@ func run(serviceConf serviceConfig, signerConf signerConfig, listen string, debu
...
+
+	// retrieve secrets for signer config as needed
+	if serviceConf.GCP.ProjectId != "" {
+		err = loadSignerSecrets(serviceConf.GCP.ProjectId, *signerConf)
+		if err != nil {
+			log.Errorf("Failed to retrieve some signer secrets! %s", err)
+		}
+	} else {
+		log.Warn("No GCP Project ID configured. Unable to load secrets.")
```
say-yawn:
```suggestion
		log.Fatalf("No GCP Project ID configured. Unable to load secrets.")
```
IMO this should exit.

database/schema.sql
alexcottner: So in this case, the `id` column is the authorization id. And the `key` column is the HAWK token. The HAWK token will always be 64 characters. I suppose we could add more restrictions on it. But the auth ID might be anywhere from like 5 to 100+ as it's a user id basically.

main.go
alexcottner: In general, we should make our error messages better.

In this specific case, we have a good situation as the calling function will log this:
`log.Fatalf("Failed to load service config. %s", err)`

